PR #123, opened 2022-06-24
Drafts in Realm

KevinBoulongne:
Depends on #122


app/src/main/java/com/infomaniak/mail/data/api/MailApi.kt
```diff
@@ -199,6 +196,20 @@ object MailApi {
...
+        apiDraft?.let { draft ->
+            draft.apply {
+                initLocalValues(parentUid)
+                // TODO: Remove this `forEachIndexed` when we have EmbeddedObjects
+                attachments.forEachIndexed { index, attachment -> attachment.initLocalValues(index, parentUid) }
+            }
+            MailboxContentController.upsertDraft(draft)
+        }
+
+        return apiDraft
```
sirambd, line 210: use return `ApiRepository.getDraft(draftResource).data?.also`

KevinBoulongne: Done in https://github.com/Infomaniak/android-mail/pull/154

PR #129, opened 2022-06-24
UI fixes

FabianDevel:
Depends on #128

PR #130, opened 2022-06-24
Rework Realm

KevinBoulongne:
Depends on #129

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/MailData.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/MailData.kt
@@ -60,13 +60,41 @@ object MailData {
     var currentMailbox: Mailbox? = null
     var currentFolder: Folder? = null
     var currentThread: Thread? = null
+    var currentMessage: Message? = null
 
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/infomaniak/mail/data/MailData.kt
```diff
@@ -0,0 +1,397 @@
...
+        currentFolder = null
+        currentMailbox = null
+    }
+
+    private fun closeFlows() {
+        mutableMessagesFlow.value = null
+        mutableThreadsFlow.value = null
+        mutableFoldersFlow.value = null
+        mutableMailboxesFlow.value = null
+    }
```
sirambd, line 88: Is it really necessary?

sirambd (changes requested):
Please make a refactor for better maintainability,
by moving all the logic that needs to be in `ViewModels` or `MainViewModel`, so we won't need to use `CoroutineScope` and it's the viewModel that will be responsible for making the api calls if possible, and the controller will just give us the realm data.

PR #132, opened 2022-06-24
Account menu

KevinBoulongne:
Depends on #143

PR #133, opened 2022-06-24
Bug fixes

KevinBoulongne:
Depends on #132

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
@@ -19,12 +19,12 @@
 
 package com.infomaniak.mail.data.models.thread
 
+import com.infomaniak.mail.data.MailData
 import com.infomaniak.mail.data.api.ApiRepository
 import com.infomaniak.mail.data.api.RealmInstantSerializer
 import com.infomaniak.mail.data.api.RealmListSerializer
 import com.infomaniak.mail.data.cache.MailRealm
 import com.infomaniak.mail.data.cache.MailboxContentController
-import com.infomaniak.mail.data.cache.MailboxInfoController
 import com.infomaniak.mail.data.models.Recipient
 import com.infomaniak.mail.data.models.message.Message
 import io.realm.kotlin.ext.realmListOf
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -45,7 +46,7 @@ class ThreadViewModel : ViewModel() {
         if (listenToMessagesJob != null) listenToMessagesJob?.cancel()
 
         listenToMessagesJob = CoroutineScope(Dispatchers.IO).launch {
```
sirambd, line 48: Why a `CoroutineScope`?  a `viewModelScope` it could be better, right?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadListViewModel.kt
```diff
@@ -42,7 +43,7 @@ class ThreadListViewModel : ViewModel() {
         if (listenToThreadsJob != null) listenToThreadsJob?.cancel()
 
         listenToThreadsJob = CoroutineScope(Dispatchers.IO).launch {
```
sirambd, line 45: Why a `CoroutineScope`? a `viewModelScope` it could be better, right?

app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -81,17 +81,15 @@ class Thread : RealmObject {
...
+                MailRealm.mailboxContent.writeBlocking {
+                    findLatest(coldThread)?.let { hotThread ->
+                        hotThread.apply {
+                            messages.forEach { it.seen = true }
+                            unseenMessagesCount = 0
                         }
+                        ApiRepository.markMessagesAsSeen(mailbox.uuid, ArrayList(messages))
                     }
                 }
             }
```
sirambd, line 95: This is the viewModel that is supposed to handle this

app/src/main/java/com/infomaniak/mail/data/models/user/UserInfos.kt
```diff
@@ -29,7 +29,7 @@ class UserInfos : RealmObject {
     var login: String = ""
     var email: String = ""
     @SerialName("avatar_url")
-    var avatarUrl: String = ""
+    var avatarUrl: String = "" // TODO: We already have an Avatar from the logged in User. Isn't this a duplicate?
```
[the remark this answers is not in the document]
KevinBoulongne: It's exactly the same content as the one from the logged in User, so we can remove this one.

PR #137, opened 2022-06-24
Mark as seen

KevinBoulongne:
Depends on #136

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
@@ -19,8 +19,12 @@ package com.infomaniak.mail.ui.main.thread
 
 import androidx.lifecycle.ViewModel
 import com.infomaniak.mail.data.MailData
+import com.infomaniak.mail.data.api.ApiRepository
+import com.infomaniak.mail.data.cache.MailRealm
 import com.infomaniak.mail.data.cache.MailboxContentController
+import com.infomaniak.mail.data.cache.MailboxContentController.getLatestThread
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -51,10 +55,32 @@ class ThreadViewModel : ViewModel() {
     }
 
     fun loadMessages(threadUid: String) {
-        val thread = MailboxContentController.getThread(threadUid) ?: return
-        MailData.selectThread(thread)
-        thread.markAsSeen()
-        MailData.loadMessages(thread)
+        MailboxContentController.getThread(threadUid)?.let { thread ->
+            MailData.selectThread(thread)
+            markAsSeen(thread)
```
sirambd, line 60: This method is blocking for the ui

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -51,10 +55,32 @@ class ThreadViewModel : ViewModel() {
...
+                    val apiResponse = ApiRepository.markMessagesAsSeen(mailboxUuid, latestThread.messages.map { it.uid })
+
+                    if (apiResponse.isSuccess()) {
+                        latestThread.apply {
+                            messages.forEach { it.seen = true }
+                            unseenMessagesCount = 0
+                        }
+                    }
+                }
+            }
```
sirambd, line 82: Move realm operation in its controller.
In addition we can make the call api well before because here we are sure that the thread exists in realm

PR #147, opened 2022-07-01
Add Contact & AddressBook

KevinBoulongne:
Depends on #139

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/MailData.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/MailData.kt
@@ -20,12 +20,14 @@ package com.infomaniak.mail.data
 import android.util.Log
 import com.infomaniak.mail.data.api.ApiRepository
 import com.infomaniak.mail.data.api.MailApi
+import com.infomaniak.mail.data.cache.ContactsController
 import com.infomaniak.mail.data.cache.MailRealm
 import com.infomaniak.mail.data.cache.MailboxContentController
 import com.infomaniak.mail.data.cache.MailboxContentController.getLatestFolder
 import com.infomaniak.mail.data.cache.MailboxContentController.getLatestMessage
 import com.infomaniak.mail.data.cache.MailboxInfoController
 import com.infomaniak.mail.data.models.AppSettings
+import com.infomaniak.mail.data.models.Contact
 import com.infomaniak.mail.data.models.Folder
 import com.infomaniak.mail.data.models.Folder.FolderRole
 import com.infomaniak.mail.data.models.Mailbox
@@ -35,10 +37,13 @@ import com.infomaniak.mail.utils.AccountUtils
 import io.realm.kotlin.Realm
 import io.realm.kotlin.UpdatePolicy
 import io.realm.kotlin.ext.toRealmList
+import io.realm.kotlin.notifications.ResultsChange
+import io.realm.kotlin.types.BaseRealmObject
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.SharedFlow
 import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.launch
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/infomaniak/mail/data/MailData.kt
```diff
@@ -104,6 +124,10 @@ object MailData {
         readMessages(thread)
     }
 
+    fun deleteDraft(message: Message) {
+        if (ApiRepository.deleteDraft(message.draftResource).isSuccess()) MailboxContentController.deleteMessage(message.uid)
+    }
```
sirambd, line 129: This method should rather be in a viewModel, as it will be directly from a UI.

app/src/main/java/com/infomaniak/mail/data/MailData.kt
```diff
@@ -292,6 +312,30 @@ object MailData {
...
+        // Save new data
+        Log.d("API", "Contacts: Save new data")
+        ContactsController.upsertContacts(apiContacts)
+
+        // Delete outdated data
+        Log.d("API", "Contacts: Delete outdated data")
+        ContactsController.deleteContacts(deletableContacts)
+
+        return apiContacts
+    }
```
sirambd, line 337: This method should rather be in `ContactsController `, since it's the same context

app/src/main/java/com/infomaniak/mail/data/MailData.kt
```diff
@@ -292,6 +312,30 @@ object MailData {
...
+     */
+
+    private fun mergeContacts(realmContacts: List<Contact>, apiContacts: List<Contact>): List<Contact> {
+
+        // Get outdated data
+        Log.d("API", "Contacts: Get outdated data")
+        // val deletableContacts = ContactsController.getDeletableContact(apiContacts)
+        val deletableContacts = realmContacts.filter { realmContact ->
+            !apiContacts.any { it.id == realmContact.id }
+        }
```
sirambd, line 326: I think it'll be better with a realm query

app/src/main/java/com/infomaniak/mail/data/MailData.kt
```diff
@@ -429,4 +473,18 @@ object MailData {
...
+    private fun <T : BaseRealmObject> SharedFlow<ResultsChange<T>>.collectOnce(completion: (List<T>) -> Unit) {
+        var job: Job? = null
+        job = CoroutineScope(Dispatchers.IO).launch {
+            this@collectOnce.collect {
+                completion(it.list.toList())
+                job?.cancel()
+            }
+        }
+    }
 }
```
sirambd, line 490: No need to rewrite this method as similar methods already exist (`first()`, `firstOrNull()`, etc...).

app/src/main/java/com/infomaniak/mail/data/api/MailApi.kt
```diff
@@ -38,6 +35,8 @@ import java.io.File
 
 object MailApi {
 
+    fun fetchContacts(): List<Contact> = ApiRepository.getContacts().data ?: emptyList()
```
sirambd, line 38: 🤔 I don't see the point unless it has many calls

app/src/main/java/com/infomaniak/mail/data/cache/ContactsController.kt
```diff
@@ -0,0 +1,47 @@
...
+    private fun MutableRealm.getLatestContact(id: String): Contact? = getContact(id)?.let(::findLatest)
+
+    fun getContacts(): Flow<ResultsChange<Contact>> = MailRealm.contacts.query<Contact>().asFlow()
+
+    fun upsertContacts(contacts: List<Contact>) {
+        MailRealm.contacts.writeBlocking { contacts.forEach { copyToRealm(it, UpdatePolicy.ALL) } }
+    }
+
+    fun deleteContacts(contacts: List<Contact>) {
+        MailRealm.contacts.writeBlocking { contacts.forEach { getLatestContact(it.id)?.let(::delete) } }
```
sirambd, line 45: you can simply this with a `realm.delete(...)` if `contacts` is a `RealmResults`

app/src/main/java/com/infomaniak/mail/data/cache/MailRealm.kt
```diff
@@ -113,25 +132,36 @@ object MailRealm {
...
-            .deleteRealmIfMigrationNeeded()
-            .build()
-
-        fun mailboxContent(currentMailboxId: Int) = RealmConfiguration
-            .Builder(RealmSets.mailboxContent)
-            .name(MAILBOX_CONTENT_DB_NAME(currentMailboxId))
-            .deleteRealmIfMigrationNeeded()
-            .build()
+        private fun MAILBOX_CONTENT_DB_NAME(mailboxId: Int) = "${AccountUtils.currentUserId}-${mailboxId}.realm"
+        private val CONTACTS_DB_NAME get() = "${AccountUtils.currentUserId}.realm"
```
sirambd, line 136: I think it'll better to also add the db name, for example `contacts-XXX.realm`, `mailbox-XXX-XXX.realm`

KevinBoulongne:
Always communicate Controller -> ViewModel -> Fragment

PR #228, opened 2022-08-30
Remove duplicated code in `loadCurrentMailbox`

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -116,6 +116,14 @@ class MainViewModel : ViewModel() {
         loadContacts()
     }
 
+    fun loadCurrentMailbox() {
+        Log.i(TAG, "loadCurrentMailbox")
+        val mailboxes = loadMailboxes()
```
sirambd, line 121: It would be nice to rename this method, its purpose is just to update realm to remote

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -116,6 +116,14 @@ class MainViewModel : ViewModel() {
...
+        computeMailboxToSelect(mailboxes)?.let { mailbox ->
+            openMailbox(mailbox)
+        }
```
sirambd, line 124: use realm query to do this

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: `updateMailboxes`, `updateFolders`, `updateThreads`, etc...
`getCurrentMailbox`, `getCurrentFolder`, etc...

sirambd (changes requested):
Waiting for request changes

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: Will be fixed in #240

sirambd (approved):

PR #250, opened 2022-09-13
Realm DiffUtil

sirambd:
Create a Realm DiffUtil to notify an adapter for all Realm changes automatically


JorisBodin (approved):

KevinBoulongne (changes requested):
Cannot be used as is.

We lost animations when the list is updated.

sirambd:
> 

Fixed

sirambd:
> 

Fixed, support animation for any changes with stable ids.

KevinBoulongne (approved):

PR #253, opened 2022-09-20
Remove Controller call in Fragment

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -182,9 +181,10 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
 
         threadsList.swipeListener = object : OnItemSwipeListener<Any> {
             override fun onItemSwiped(position: Int, direction: SwipeDirection, item: Any): Boolean {
+                item as Thread
```
sirambd, line 184: Why here ?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
KevinBoulongne: Because we'll also need it there:
```
SwipeDirection.RIGHT_TO_LEFT -> notYetImplemented() // TODO: Delete thread
```

We could wait for the day that we implement the `RIGHT_TO_LEFT` swipe, but @JorisBodin asked to do it now.

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListViewModel.kt
```diff
@@ -57,4 +59,8 @@ class ThreadListViewModel : ViewModel() {
             }
         }
     }
+
+    fun toggleSeenStatus(thread: Thread, folderId: String) = viewModelScope.launch(Dispatchers.IO) {
+        ThreadController.toggleSeenStatus(thread, folderId)
```
sirambd, line 64: Need a refactor, move api call here

sirambd (changes requested):

LunarX (approved):

PR #259, opened 2022-09-21
Pop backstack & delete Thread when opening an empty Thread

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -170,10 +170,18 @@ class ThreadFragment : Fragment() {
...
         Log.i("UI", "Received messages (${messages.size})")
+        leaveIfThreadIsEmpty(messages)
         threadAdapter.notifyAdapter(messages.toMutableList())
         binding.messagesList.scrollToPosition(threadAdapter.lastIndex())
     }
 
+    private fun leaveIfThreadIsEmpty(messages: List<Message>) {
+        if (messages.isEmpty()) {
+            threadViewModel.deleteThread(navigationArgs.threadUid)
+            findNavController().popBackStack()
```
LunarX, line 181: If a bottom sheet is open when realm triggers, it might simply close the bottom sheet instead of closing the thread

LunarX (changes requested):

JorisBodin (approved):

PR #284, opened 2022-10-11
Fix Login `accentColor`

KevinBoulongne:
Depends on #279

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
@@ -54,6 +54,8 @@ class IntroFragment : Fragment() {
     private val navigationArgs: IntroFragmentArgs by navArgs()
     private val loginViewModel: LoginViewModel by activityViewModels()
 
+    private val uiSettings by lazy { UiSettings.getInstance(requireContext()) }
+
     override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View {
         return FragmentIntroBinding.inflate(inflater, container, false).also { binding = it }.root
     }
@@ -86,13 +88,15 @@ class IntroFragment : Fragment() {
         }
 
         updateUiWhenThemeChanges(navigationArgs.position)
+
+        setUi(uiSettings.accentColor, navigationArgs.position, animate = false)
     }
 
     private fun setTabSelectedListener() = with(binding) {
         pinkBlueTabLayout.addOnTabSelectedListener(object : TabLayout.OnTabSelectedListener {
             override fun onTabSelected(tab: TabLayout.Tab?) {
                 val newSelectedAccentColor = if (pinkBlueTabLayout.selectedTabPosition == PINK.introTabIndex) PINK else BLUE
-                UiSettings.getInstance(requireContext()).accentColor = newSelectedAccentColor
+                uiSettings.accentColor = newSelectedAccentColor
                 triggerUiUpdateWhenAnimationEnd(newSelectedAccentColor)
             }
 
@@ -111,15 +115,17 @@ class IntroFragment : Fragment() {
 
     private fun updateUiWhenThemeChanges(position: Int?) {
         loginViewModel.currentAccentColor.observeNotNull(viewLifecycleOwner) { accentColor ->
-            setUi(accentColor, position)
+            setUi(accentColor, position, animate = true)
         }
     }
 
-    private fun setUi(accentColor: AccentColor, position: Int?) = with(binding) {
-        val animate = position == 0
+    /**
+     * `animate` is necessary because when the activity is started, we want to avoid animating the color change the first time.
+     */
+    private fun setUi(accentColor: AccentColor, position: Int?, animate: Boolean) = with(binding) {
         updateEachPageUi(accentColor, animate)
-        if (position == 0) updateFirstPageUi(accentColor, animate)
-        if (position == 3) updateActivityUi(accentColor, animate)
+        if (position == ACCENT_COLOR_PICKER_PAGE) updateAccentColorPickerPageUi(accentColor, animate)
+        if (position == A_SINGLE_ALWAYS_PRESENT_PAGE) updateActivityUi(accentColor, animate)
     }
 
     private fun updateEachPageUi(accentColor: AccentColor, animate: Boolean) = with(binding) {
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
```diff
@@ -109,29 +106,21 @@ class IntroFragment : Fragment() {
...
+            setUi(accentColor, position)
         }
     }
 
-    /**
-     * `animate` is necessary because when the activity is started we want to avoid animating the color change the first time.
-     */
-    private fun setUi(accentColor: AccentColor, position: Int?, animate: Boolean = true) = with(binding) {
+    private fun setUi(accentColor: AccentColor, position: Int?) = with(binding) {
+        val animate = position == 0
```
LunarX, line 120: This won't prevent animation when adding your first account to the app. Also add back old comment

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
```diff
@@ -109,29 +106,21 @@ class IntroFragment : Fragment() {
...
         updateEachPageUi(accentColor, animate)
-        if (position == 0) {
-            updateFirstPageUi(accentColor, animate)
-            if (navigationArgs.isFirstAccount) updateActivityUi(accentColor, animate)
-        }
+        if (position == 0) updateFirstPageUi(accentColor, animate)
+        if (position == 3) updateActivityUi(accentColor, animate)
```
LunarX, line 123: Add comment to explain why doing it only at position == 3 is important

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
```diff
@@ -66,8 +64,7 @@ class IntroFragment : Fragment() {
         when (navigationArgs.position) {
             0 -> {
                 pinkBlueSwitch.isVisible = true
-                val tabIndex = introViewModel.currentAccentColor.value?.introTabIndex
-                    ?: UiSettings.DEFAULT_ACCENT_COLOR.introTabIndex
+                val tabIndex = loginViewModel.currentAccentColor.value?.introTabIndex!!
```
LunarX, line 67: Is it dangerous to !! a LiveData ?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
sirambd: At position 0, animate is always true
At position 3, animate is always false
Please review this code

KevinBoulongne: Magic numbers have been removed.

There was also a bug that was fixed, so `animate` computation has been reworked.

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
sirambd: If there is a possibility that the `livedata` becomes null yes

sirambd: But we can make it safer, we should use a `let` here, 
and move the default value from mainActivity into the `ViewModel` directly.

app/src/main/java/com/infomaniak/mail/ui/login/LoginActivity.kt
```diff
@@ -83,6 +87,8 @@ class LoginActivity : AppCompatActivity() {
             clientID = BuildConfig.CLIENT_ID,
         )
 
+        loginViewModel.currentAccentColor.value = UiSettings.getInstance(context).accentColor
```
sirambd, line 90: Init it directly in the `ViewModel`

app/src/main/java/com/infomaniak/mail/ui/login/LoginViewModel.kt
```diff
@@ -0,0 +1,26 @@
...
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.mail.ui.login
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import com.infomaniak.mail.data.UiSettings.AccentColor
+
+class LoginViewModel : ViewModel() {
+    var currentAccentColor: MutableLiveData<AccentColor?> = MutableLiveData(null)
```
sirambd, line 25: Init with this preference value instead of `null`

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
```diff
@@ -185,7 +181,8 @@ class IntroFragment : Fragment() {
         (requireActivity() as LoginActivity).updateUi(accentColor, animate)
     }
 
-    class IntroViewModel : ViewModel() {
-        var currentAccentColor: MutableLiveData<AccentColor?> = MutableLiveData(null)
+    private companion object {
+        const val ACCENT_COLOR_PICKER_PAGE = 0
+        const val A_SINGLE_ALWAYS_PRESENT_PAGE = 3
```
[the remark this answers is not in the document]
KevinBoulongne: `LOGIN_BUTTON_PAGE`, because it's not the Login webview page, but the page just before, the one with the "Login" button.

app/src/main/java/com/infomaniak/mail/ui/login/LoginViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+import androidx.lifecycle.MutableLiveData
+import com.infomaniak.mail.ApplicationMain
+import com.infomaniak.mail.data.UiSettings
+import com.infomaniak.mail.data.UiSettings.AccentColor
+
+class LoginViewModel(application: ApplicationMain) : AndroidViewModel(application) {
+
+    private inline val context: Context get() = getApplication<Application>().applicationContext
+
+    var currentAccentColor: MutableLiveData<AccentColor?> = MutableLiveData(UiSettings.getInstance(context).accentColor)
```
sirambd, line 32: You can directly use `application` here instead of `context`

app/src/main/java/com/infomaniak/mail/ui/login/LoginViewModel.kt
```diff
@@ -0,0 +1,33 @@
...
+import android.content.Context
+import androidx.lifecycle.AndroidViewModel
+import androidx.lifecycle.MutableLiveData
+import com.infomaniak.mail.ApplicationMain
+import com.infomaniak.mail.data.UiSettings
+import com.infomaniak.mail.data.UiSettings.AccentColor
+
+class LoginViewModel(application: ApplicationMain) : AndroidViewModel(application) {
+
+    private inline val context: Context get() = getApplication<Application>().applicationContext
```
sirambd, line 30: Useless

app/src/main/java/com/infomaniak/mail/ui/login/LoginActivity.kt
```diff
@@ -54,6 +55,7 @@ class LoginActivity : AppCompatActivity() {
 
     private val binding by lazy { ActivityLoginBinding.inflate(layoutInflater) }
     private val navigationArgs by lazy { LoginActivityArgs.fromBundle(intent.extras ?: bundleOf()) }
+    private val loginViewModel: LoginViewModel by viewModels()
```
sirambd, line 58: is never used

sirambd (approved):

PR #288, opened 2022-10-14
Directly go to NewMessage when opening a Thread that is just 1 Draft

KevinBoulongne:
Depends on #287

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -372,6 +372,18 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
     }
     //endregion
 
+    //region Open Draft
+    val openDraftUuid = SingleLiveEvent<String?>()
+
+    fun fetchDraft(message: Message) = viewModelScope.launch(Dispatchers.IO) {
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -313,6 +310,14 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
...
+    //region Open Draft
+    fun fetchDraft(draftResource: String, messageUid: String): String? {
+        return ApiRepository.getDraft(draftResource).data?.also { draft ->
+            DraftController.upsertDraft(draft.initLocalValues(messageUid))
+        }?.uuid
+    }
+    //endregion
+
     //region New Message
     fun executeDraftsActions() = viewModelScope.launch(Dispatchers.IO) {
```
sirambd, line 322: Clean actions when executed

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: Will be done in another PR.

sirambd (approved):

PR #299, opened 2022-10-18
Add `Signature` to Realm

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/models/signature/Signature.kt
```diff
@@ -43,8 +44,17 @@ class Signature : RealmObject {
     var isDefault: Boolean = false
     @SerialName("service_mail_model_id")
     var serviceMailModelId: Int? = null
-    var position: String = ""
+    @SerialName("position")
+    private var _position: String = ""
```
sirambd, line 48: use enum type instead of `String`

app/src/main/java/com/infomaniak/mail/data/models/signature/Signature.kt
```diff
@@ -43,8 +44,17 @@ class Signature : RealmObject {
...
     @SerialName("is_editable")
     var isEditable: Boolean = false
     var content: String = ""
+
+    val position: SignaturePosition?
+        get() = enumValueOfOrNull<SignaturePosition>(_position)
```
sirambd, line 54: use enum type in the class and set a default value

sirambd (changes requested):

sirambd (approved):

PR #313, opened 2022-10-27
Add Bugs tracker

LunarX:
Needs https://github.com/Infomaniak/android-core/pull/14 to be reviewed


JorisBodin (changes requested):
It is necessary that the PR in core is accepted before validating it
https://github.com/Infomaniak/android-core/pull/14

JorisBodin (approved):

PR #315, opened 2022-10-31
Implement `Reply` & `Reply all`

KevinBoulongne:
Depends on #314

PR #316, opened 2022-10-31
Get last Message to execute action

KevinBoulongne:
Depends on #315

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
@@ -34,6 +34,7 @@ import com.infomaniak.lib.core.utils.*
 import com.infomaniak.lib.core.utils.SnackbarUtils.showSnackbar
 import com.infomaniak.mail.R
 import com.infomaniak.mail.data.models.draft.Draft.DraftMode
+import com.infomaniak.mail.data.models.message.Message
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -121,6 +122,8 @@ fun Fragment.safeNavigateToNewMessageActivity(draftMode: DraftMode, messageUid:
     )
 }
 
+fun List<Message>.getLastMessageToExecuteAction(): Message = lastOrNull { !it.isDraft } ?: last()
```
JorisBodin, line 125: If there are only drafts in the list, will it return a draft? Is this a desired behavior?
If not, we might as well put `last { !it.isDraft }` directly and we'll have a crash if it happens.

JorisBodin (changes requested):

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
KevinBoulongne: If there's only Drafts, we want to get the last one. This is the desired behavior.

Example: Any action on a Thread in the Draft folder. They will all be actions on Thread of a single Message, with `isDraft = true`.

JorisBodin (approved):

PR #349, opened 2022-11-16
Improve Draft actions worker

sirambd:
**Fixes**
- [x] [Fix expedited work request for api before 31](https://github.com/Infomaniak/android-mail/commit/d6407b796c096882d7576c8b838625d5f550ee5a) 
> The expedited work before android 31 it creates services in the foreground and this is not what we want here

- [x] [Launch getRunningWorkInfosLiveData only once](https://github.com/Infomaniak/android-mail/commit/bbacea5773faff33187ce2b590a2d644a68c685d)
> Otherwise the service may be called back several times

- [x] [Worker with its own realms instances](https://github.com/Infomaniak/android-mail/commit/ceaaaca22bba1337d2f066f2c02d82af7a7d8dd0)
> Since the service can be in background, we don't want it to be dependent on the realm of the app, because it can create problems when the app is closed

- [x] [Manage api exceptions](https://github.com/Infomaniak/android-mail/commit/bc7162aca3bd2dc0cc54c75a0512312e5153c2d0)
> Need some modifications in `Core` to allow exceptions to be thrown from the `ApiController` whenever desired.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
@@ -47,6 +47,7 @@ object ApiRepository : ApiRepositoryCore() {
         method: ApiController.ApiMethod,
         body: Any? = null,
         okHttpClient: OkHttpClient = HttpClient.okHttpClient,
+        throwExceptions: Boolean = false
     ): T {
         return ApiController.callApi(url, method, body, okHttpClient, useKotlinxSerialization = true)
     }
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -47,6 +47,7 @@ object ApiRepository : ApiRepositoryCore() {
         method: ApiController.ApiMethod,
         body: Any? = null,
         okHttpClient: OkHttpClient = HttpClient.okHttpClient,
+        throwExceptions: Boolean = false
```
KevinBoulongne: This is unused, is it on purpose?

KevinBoulongne: I guess it was supposed to be passed down to the `Core.ApiController.callApi()` function.
I did it.

KevinBoulongne (approved):

JorisBodin (approved):

PR #352, opened 2022-11-17
Add `getCurrentMailbox` & `getCurrentMailboxUuid` helpers

KevinBoulongne:
Depends on #343

PR #354, opened 2022-11-18
Optimize `toggleSeenStatus` realm usage

KevinBoulongne:
Depends on #352

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -106,47 +106,43 @@ object ThreadController {
     private fun markAsUnseen(thread: Thread) {
 
         val mailboxUuid = MailboxController.getCurrentMailboxUuid() ?: return
+        val uid = getThreadLastMessageUid(thread)
 
-        RealmDatabase.mailboxContent().writeBlocking {
-            val latestThread = findLatest(thread) ?: return@writeBlocking
-            val uid = getThreadLastMessageUid(latestThread)
-            with(ApiRepository.markMessagesAsUnseen(mailboxUuid, uid)) {
-                if (isSuccess()) markThreadAsUnseen(latestThread)
-            }
-        }
+        if (ApiRepository.markMessagesAsUnseen(mailboxUuid, uid).isSuccess()) markThreadAsUnseen(thread.uid)
     }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -118,32 +100,53 @@ object ThreadController {
...
+        val mailboxUuid = MailboxController.getCurrentMailboxUuid() ?: return
+
+        RealmDatabase.mailboxContent().writeBlocking {
+            val latestThread = findLatest(thread) ?: return@writeBlocking
+            val uids = getThreadUnseenMessagesUids(latestThread)
+            with(ApiRepository.markMessagesAsSeen(mailboxUuid, uids)) {
+                if (isSuccess()) markThreadAsSeen(latestThread)
+            }
         }
     }
```
JorisBodin, line 131: I think we could factor better, there is some duplication

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -118,32 +100,53 @@ object ThreadController {
...
-        val uids = getThreadUnseenMessagesUids(latestThread)
-        with(ApiRepository.markMessagesAsSeen(latestThread.mailboxUuid, uids)) {
-            if (isSuccess()) realm.markThreadAsSeen(latestThread, folderId)
+    private fun MutableRealm.markThreadAsSeen(thread: Thread) {
+        thread.messages.forEach {
+            if (!it.seen) {
+                incrementFolderUnreadCount(it.folderId, -1)
+                it.seen = true
+            }
         }
```
JorisBodin, line 147: You could call `incrementFolderUnreadCount` after the loop

JorisBodin (approved):

JorisBodin (changes requested):
.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
KevinBoulongne: The `folderId` is different for each `Message`, so we need to do it in the loop.

sirambd (approved):

JorisBodin (approved):

PR #369, opened 2022-11-24
Notify only previous Folder & selected Folder in `FolderAdapter`

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
```diff
@@ -83,16 +84,24 @@ class FolderAdapter(
...
 
+    @SuppressLint("NotifyDataSetChanged")
     fun setFolders(newFolders: List<Folder>, newCurrentFolderId: String?) {
         currentFolderId = newCurrentFolderId
         folders = newFolders
         notifyDataSetChanged()
     }
 
     fun updateSelectedState(newCurrentFolderId: String) {
+        notifyCurrentItem()
```
sirambd, line 95: Why are we notifying twice in a row?
And when do we notify `newCurrentFolderId`?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
KevinBoulongne: This code was wrong, it's been redone.

JorisBodin (approved):

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
sirambd: Where was it redone?

KevinBoulongne: In this PR.

PR #379, opened 2022-11-28
Base network coroutine worker

sirambd:
A base of network CoroutineWorker

Depends on #368


app/src/main/java/com/infomaniak/mail/workers/BaseCoroutineWorker.kt
```diff
@@ -0,0 +1,56 @@
...
+        }.also {
+            onFinish()
+        }
+    }
+
+    companion object {
+        private const val MAX_RETRIES = 3
+    }
+
+}
```
KevinBoulongne, line 56:
```suggestion
    private companion object {
        const val MAX_RETRIES = 3
    }
}

```

JorisBodin (approved):

PR #400, opened 2022-12-02
Set Message' subject at `null` when empty

KevinBoulongne:
When an email has no subject, the string is supposed to be `null`.
But sometimes the string isn't null, instead it's empty _(for example, when sending the email via Gmail)_.

Depends on #403


JorisBodin (changes requested):
Modify directly `getFormattedThreadSubject` instead of doing this

    fun String?.getFormattedThreadSubject(context: Context): Spanned {
        return if (this?.isNotBlank() == true) {
            this.replace("\n+".toRegex(), " ").toSpanned()
        } else {
            HtmlCompat.fromHtml("<i>${context.getString(R.string.noSubjectTitle)}</i>", HtmlCompat.FROM_HTML_MODE_COMPACT)
        }
    }

JorisBodin (approved):

PR #406, opened 2022-12-06
Remove italic when subject is missing

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -170,8 +170,9 @@ class ThreadFragment : Fragment() {
 
     private fun onThreadUpdate(thread: Thread) = with(binding) {
 
-        thread.setFormattedSubject(threadSubject, R.style.H2)
-        thread.setFormattedSubject(toolbarSubject, R.style.H2)
+        val subject = thread.getFormattedSubject(context)
```
sirambd, line 173: You can use `with`

sirambd (approved):

PR #412, opened 2022-12-08
Made whole swipe setting illustration clickable

LunarX:


KevinBoulongne (approved):

app/src/main/res/values/attrs.xml
```diff
@@ -60,6 +60,7 @@
             <enum name="chevron" value="1" />
             <enum name="toggle" value="2" />
         </attr>
+        <attr name="noRipple" format="boolean" />
```
JorisBodin, line 63: I would have rather put simple `ripple` with default value `true`.

JorisBodin (changes requested):

JorisBodin (approved):

PR #416, opened 2022-12-12
Threads by Folder

KevinBoulongne:
Depends on #408


JorisBodin (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
```diff
@@ -69,7 +69,7 @@ class Folder : RealmObject {
     var unreadCount: Int = 0
     //endregion
 
-    val parentFolder by backlinks(Folder::children)
+    val parentFolders by backlinks(Folder::children)
```
sirambd, line 72: A folder can have several parents

app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -52,25 +53,25 @@ class Thread : RealmObject {
...
     }
 
-    fun removeMessage(message: Message) {
-        messages.removeIf { it.uid == message.uid }
-        recomputeThread()
-    }
+    fun recomputeThread(realm: MutableRealm) {
 
-    fun recomputeThread() {
+        //region Delete Thread if empty
```
sirambd, line 65: Instead of using regions, use functions instead

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
KevinBoulongne: It's been discussed & fixed.

sirambd (approved):

JorisBodin (approved):

PR #427, opened 2022-12-16
Clean network listener code

KevinBoulongne:


JorisBodin (approved):

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -118,16 +118,17 @@ class MainActivity : ThemedActivity() {
     }
 
     private fun observeNetworkStatus() {
+        var previousIsAvailable: Boolean? = null
```
sirambd, line 121: His trigger several times?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
KevinBoulongne: Exactly, it triggers twice when the network is available.

sirambd: So this is not the place to manage it

sirambd: It is necessary to do it from the `livedata` itself, so it notifies only when it is necessary

KevinBoulongne: At first, I thought the `LiveDataNetworkStatus` was bugged, but in fact it's not.
It broadcasts the global network status when any network's connectivity status changes.

So I just added a `distinctUntilChanged()` to only listen to the global network status changes.

JorisBodin (approved):

PR #428, opened 2022-12-19
Factorize code to get displayed Message in Threads list

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
```diff
@@ -123,16 +123,19 @@ class ThreadListAdapter(
         }
     }
 
-    private fun getDisplayedRecipient(thread: Thread): Recipient? = thread.messages.lastOrNull()?.from?.firstOrNull()
+    private fun Thread.getDisplayedMessage() = messages.lastOrNull()
```
sirambd, line 126: Is it possible to have threads that have no messages?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
KevinBoulongne: No. I've updated it.

JorisBodin (approved):

PR #431, opened 2022-12-19
Handle SendTo

sirambd:
**Features**
 - Handle `ACTION_SEND` (text or attachment) and `ACTION_SEND_MULTIPLE` with attachments
 - Handle `mailto` from `ACTION_VIEW` or `ACTION_SENDTO`

**Fix**
- intent has null extras with navArgs when navigate with deeplink from `ACTION_SENDTO` intent

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
@@ -72,7 +72,7 @@ class NewMessageFragment : Fragment() {
     private val newMessageActivityArgs by lazy {
         // When opening this fragment via deeplinks, it can happen that the navigation
         // extras aren't yet initialized, so we don't use the `navArgs` here.
-        requireActivity().intent?.extras?.let { NewMessageActivityArgs.fromBundle(it) } ?: NewMessageActivityArgs()
+        requireActivity().intent?.extras?.let(NewMessageActivityArgs::fromBundle) ?: NewMessageActivityArgs()
     }
     private val newMessageViewModel: NewMessageViewModel by activityViewModels()
 
--- a/app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
@@ -70,7 +70,7 @@ class NewMessageFragment : Fragment() {
 
     private lateinit var binding: FragmentNewMessageBinding
     private val newMessageActivityArgs by lazy {
-        // When opening this fragment via deeplinks, it can happen that the navigation
+        // When opening this fragment via deeplink, it can happen that the navigation
         // extras aren't yet initialized, so we don't use the `navArgs` here.
         requireActivity().intent?.extras?.let(NewMessageActivityArgs::fromBundle) ?: NewMessageActivityArgs()
     }
--- a/app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
@@ -19,7 +19,10 @@ package com.infomaniak.mail.ui.main.newMessage
 
 import android.annotation.SuppressLint
 import android.content.ClipDescription
+import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
+import android.os.Parcelable
 import android.text.InputFilter
 import android.text.Spanned
 import android.transition.TransitionManager
@@ -43,6 +46,8 @@ import com.google.android.material.chip.ChipGroup
 import com.google.android.material.textfield.MaterialAutoCompleteTextView
 import com.infomaniak.lib.core.utils.FilePicker
 import com.infomaniak.lib.core.utils.SnackbarUtils.showSnackbar
+import com.infomaniak.lib.core.utils.parcelableArrayListExtra
+import com.infomaniak.lib.core.utils.parcelableExtra
 import com.infomaniak.lib.core.utils.setMarginsRelative
 import com.infomaniak.mail.R
 import com.infomaniak.mail.data.models.Mailbox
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -163,15 +171,75 @@ class NewMessageFragment : Fragment() {
...
+                observeContacts()
+                populateUiWithExistingDraftData()
+                handleActionSend()
+            } else {
+                requireActivity().finish()
+            }
+        }
+    }
+
+    private fun handleActionSend() {
```
JorisBodin, line 185: Why don't you do this in the activity directly?

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -163,15 +171,73 @@ class NewMessageFragment : Fragment() {
...
+     */
+    private fun handleMailTo() = with(binding) {
+        val intent = requireActivity().intent
+        intent?.data?.let { uri ->
+            if (!MailTo.isMailTo(uri)) return@with
+
+            val mailTo = MailTo.parse(uri)
+            val to = mailTo.to?.split(",")?.map { it.trim() } ?: emptyList()
+            val cc = mailTo.cc?.split(",") ?: intent.getStringArrayExtra(Intent.EXTRA_CC)?.toList() ?: emptyList()
+            val bcc = mailTo.bcc?.split(",") ?: intent.getStringArrayExtra(Intent.EXTRA_BCC)?.toList() ?: emptyList()
```
JorisBodin, line 204: Why is only the "to" trim?

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -163,15 +171,73 @@ class NewMessageFragment : Fragment() {
...
+
+            to.forEach {
+                toAutocompleteInput.setText(it)
+                contactAdapter.addFirstAvailableItem()
```
JorisBodin, line 208: Why call `addFirstAvailableItem` at every loop?

JorisBodin (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
sirambd: Its clean to have all the shares in one place, a reason for the activity?
Otherwise the code will look like magic 🤔

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
sirambd: I agree, This is an oversight during my conflicts 😅

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
sirambd: To create the chips one by one for each recipient

JorisBodin: Might as well call "createChip" directly?

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -64,7 +69,10 @@ import com.infomaniak.lib.core.R as RCore
 class NewMessageFragment : Fragment() {
 
     private lateinit var binding: FragmentNewMessageBinding
-    private val newMessageActivityArgs by lazy { requireActivity().navArgs<NewMessageActivityArgs>().value }
+    private val newMessageActivityArgs by lazy {
+        // When deeplink it happens that the navigation extras are not yet initialized, so we don't use the `navArgs` here
```
KevinBoulongne, line 73:
```suggestion
        // When opening this fragment via deeplinks, it can happen that the navigation
        // extras aren't yet initialized, so we don't use the `navArgs` here.
```

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -64,7 +69,10 @@ import com.infomaniak.lib.core.R as RCore
...
+        requireActivity().intent?.extras?.let { NewMessageActivityArgs.fromBundle(it) } ?: NewMessageActivityArgs()
```
KevinBoulongne, line 74:
```suggestion
        requireActivity().intent?.extras?.let(NewMessageActivityArgs::fromBundle) ?: NewMessageActivityArgs()
```

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
LunarX: Most of the content of the `addUnrecognizedMail` should indeed be called when adding these new recipients but I think that the email should be passed directly to a function as an argument rather than being obtained through the content of the `TextView` (first line of `addUnrecognizedMail`)

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
sirambd: So the fact of using directly the method sa I already thought of it during my analysis, but I found interesting to be able to recover the contact mail in the case that it existed, so use what existed. 
Although I didn't like this quick solution, but moreover when I look closely there is a bug on it, so it's like using the method directly 🥲. 
In this case go use the method directly.

KevinBoulongne (approved):

JorisBodin (approved):

PR #433, opened 2022-12-20
Fix invalid MergedContact

KevinBoulongne:
Sentry: https://sentry.infomaniak.com/organizations/infomaniak/issues/63999/?project=45&query=&statsPeriod=14d


JorisBodin (approved):

sirambd (changes requested):
Do we know why this happens?
That an object is realm is invalid just after a return after a standby?

sirambd (approved):

PR #446, opened 2023-01-03
RealmChangesBinding can now check if the content has been changed in the ui

sirambd:
**Description**
Currently it is realm that takes care of updating our ui.
But there are some cases where we have realm changes but we don't want the ui to be updated.

**Fix**
This solution adds a feature that will prevent `RealmChangesBinding` to check if the content has really changed before notifying by overriding the `areContentsTheSame` method from the adapter

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/RealmChangesBinding.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/RealmChangesBinding.kt
@@ -21,6 +21,7 @@ import android.annotation.SuppressLint
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.LiveData
 import androidx.recyclerview.widget.RecyclerView
+import com.infomaniak.mail.utils.RealmChangesBinding.OnRealmChanged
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/utils/RealmChangesBinding.kt
```diff
@@ -109,25 +116,61 @@ class RealmChangesBinding<T : BaseRealmObject, VH : RecyclerView.ViewHolder> pri
...
+                    recyclerViewAdapter.notifyItemRangeChanged(start, count)
+                    count = 0
+                }
+            } else {
+                // For the first change we get its index in start and then we count the number of changes
+                if (count == 0) start = index
+                count++
+            }
+        }
+        // if we finish the iteration and we still have some modifications to notify, we treat them here
```
KevinBoulongne, line 172:
```suggestion
        // If we finish the iteration and we still have some modifications to notify, we treat them here
```

KevinBoulongne (approved):

JorisBodin (approved):

PR #447, opened 2023-01-03
Don't delete a Message if we can't download its `body`

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -64,24 +62,19 @@ class ThreadViewModel(application: Application) : AndroidViewModel(application)
 
     private fun updateMessages(thread: Thread) {
```
sirambd, line 63: The name is not very appropriate, you should use a name like, `fetchIncompleteMessages` 🤔

sirambd (changes requested):

sirambd (approved):

PR #450, opened 2023-01-04
Only scroll to expanded Message when opening the Thread

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -49,14 +49,28 @@ class ThreadViewModel(application: Application) : AndroidViewModel(application)
...
+            emit(null)
+            return@liveData
+        }
+
+        val thread = ThreadController.getThread(threadUid) ?: run {
+            emit(null)
+            return@liveData
+        }
+
+        val isExpanded = thread.messages.mapIndexed { index, message ->
```
sirambd, line 64: Rename this with `expandedList` because it is a list not a boolean

sirambd (changes requested):

sirambd (approved):

JorisBodin (approved):

PR #451, opened 2023-01-05
Only open a Thread if it exists in Realm

sirambd:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -41,6 +41,8 @@ class ThreadViewModel(application: Application) : AndroidViewModel(application)
 
     private val localSettings by lazy { LocalSettings.getInstance(application) }
 
+    fun getThread(threadUid: String) = ThreadController.getThread(threadUid)
```
KevinBoulongne, line 44: This is a Realm query, and it will be executed on the main thread. We don't want to do it on this thread.

KevinBoulongne:
Why do we need this feature ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
sirambd: No the case of a get it can pass, it is the case of a transaction or a long query.

KevinBoulongne: Ok.

PR #458, opened 2023-01-09
Remove API calls from ThreadController

KevinBoulongne:
Depends on #459

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -25,7 +25,10 @@ import com.infomaniak.mail.R
 import com.infomaniak.mail.data.LocalSettings
 import com.infomaniak.mail.data.api.ApiRepository
 import com.infomaniak.mail.data.cache.RealmDatabase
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -302,14 +300,76 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
         refreshThreads()
     }
 
+    //region Open Thread
+    fun openThread(threadUid: String) = liveData(Dispatchers.IO) {
```
sirambd, line 304: Why did you move it to here?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: Because I moved `markAsSeen()` from ThreadController to MainViewModel.

sirambd (approved):

JorisBodin (approved):

PR #455, opened 2023-01-09
Archive  action

FabianDevel:
Add archive action on thread and message + swipeAction

PR #465, opened 2023-01-10
Refresh all concerned Folders when executing an action

KevinBoulongne:
Depends on #455

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -121,7 +121,20 @@ object ThreadController {
         realm?.let(block) ?: RealmDatabase.mailboxContent().writeBlocking(block)
     }
 
-    // TODO: Replace this with a RealmList sub query (blocked by https://github.com/realm/realm-kotlin/issues/1037)
+    // TODO: Replace this with a RealmList sub query.
+    fun getThreadLastMessage(thread: Thread): List<Message> {
+        return mutableListOf<Message>().apply {
+
+            fun List<Message>.getLastMessageToExecuteAction(): Message = lastOrNull { !it.isDraft } ?: last()
+
+            val lastMessage = thread.messages.getLastMessageToExecuteAction()
+            add(lastMessage)
+
+            addAll(thread.getMessageDuplicates(lastMessage.messageId))
+        }
+    }
+
+    // TODO: Replace this with a RealmList sub query.
     fun getThreadLastMessageUids(thread: Thread): List<String> {
         return mutableListOf<String>().apply {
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -134,28 +147,45 @@ object ThreadController {
...
-                    addAll(thread.getMessageDuplicatesUids(message.messageId))
-                }
-            }
-        }
+        return getThreadMessagesAndDuplicatesUids(thread) { message -> !message.seen }
+    }
+
+    // TODO: Replace this with a RealmList sub query.
+    fun getThreadUnseenMessages(thread: Thread): List<Message> {
+        return getThreadMessagesAndDuplicates(thread) { message -> !message.seen }
```
sirambd, line 157: So the only difference between these two methods is the type of return that changes.
But except that it makes a dupplication of logic, and also especially when we will pass to the `RealmList subquery` the other method will be useless because this one will just return objects.

And also we make also dupplications of logic and also especially the met🤔hod which returns the uids is not even used because it is already treated with a map in `MainViewModel`

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -134,28 +147,45 @@ object ThreadController {
...
         return getThreadMessagesAndDuplicatesUids(thread) { message -> message.folderId == thread.folderId }
     }
 
+    fun getThreadFavoritesMessages(thread: Thread): List<Message> {
+        return getThreadMessagesAndDuplicates(thread) { message -> message.isFavorite && !message.isDraft }
+    }
+
     fun getThreadFavoritesMessagesUids(thread: Thread): List<String> {
         return getThreadMessagesAndDuplicatesUids(thread) { message -> message.isFavorite && !message.isDraft }
     }
```
sirambd, line 170: Is it really useful to have both, when the other is not even used ?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
KevinBoulongne: I removed the `...Uids()` versions.

This refactor is in another PR: https://github.com/Infomaniak/android-mail/pull/470

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
KevinBoulongne: I removed the `...Uids()` versions.

This refactor is in another PR: https://github.com/Infomaniak/android-mail/pull/470

sirambd (approved):

JorisBodin (approved):

PR #475, opened 2023-01-13
Display new Message in Thread after sending it

KevinBoulongne:
Depends on #477

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
+++ b/app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
@@ -1,6 +1,6 @@
 /*
  * Infomaniak kMail - Android
- * Copyright (C) 2022 Infomaniak Network SA
+ * Copyright (C) 2022-2023 Infomaniak Network SA
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -102,7 +113,33 @@ class DraftsActionsWorker(appContext: Context, params: WorkerParameters) : BaseC
...
             }
 
-            if (hasRemoteException) Result.failure() else Result.success()
+            return@writeBlocking hasRemoteException
+        }
+
+        return if (isFailure) {
+            Result.failure()
+        } else {
+            updateFolderAfterScheduledDate(scheduledDates)
```
sirambd, line 122: This function is called only if the list is not empty, because `isFailure` can be `true` when the list is not empty.

app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -174,6 +220,7 @@ class DraftsActionsWorker(appContext: Context, params: WorkerParameters) : BaseC
         private const val TAG = "DraftsActionsWorker"
         private const val USER_ID_KEY = "userId"
         private const val MAILBOX_ID_KEY = "mailboxIdKey"
+        private const val REFRESH_DELAY = 1_500L
```
sirambd, line 223: Add a comment to explain it

sirambd (changes requested):

JorisBodin (approved):

PR #476, opened 2023-01-13
feat: close current mailbox notifications

sirambd:


app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -95,6 +95,7 @@ class MainActivity : ThemedActivity() {
     override fun onStart() {
         super.onStart()
         SyncMessagesWorker.cancelWork(this)
+        mainViewModel.cancelCurrentMailboxNotifications()
```
KevinBoulongne, line 98: This probably won't trigger at boot, because the Mailbox flow will not be set yet.

KevinBoulongne: Fixed.

JorisBodin (approved):

PR #490, opened 2023-01-18
feat: download notifications messages body

sirambd:
When we display a message, we also download its body, but for the display of the body we parse it in html, while filtering the images

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -17,6 +17,7 @@
  */
 package com.infomaniak.mail.data.cache.mailboxContent
 
+import com.infomaniak.mail.data.api.ApiRepository
 import com.infomaniak.mail.data.cache.RealmDatabase
 import com.infomaniak.mail.data.models.thread.Thread
 import com.infomaniak.mail.data.models.thread.Thread.ThreadFilter
@@ -28,6 +29,7 @@ import io.realm.kotlin.notifications.ResultsChange
 import io.realm.kotlin.notifications.SingleQueryChange
 import io.realm.kotlin.query.*
 import kotlinx.coroutines.flow.Flow
+import okhttp3.OkHttpClient
 
 object ThreadController {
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -118,5 +119,32 @@ object ThreadController {
     fun deleteThreads(folderId: String, realm: MutableRealm) {
         realm.delete(getThreadsQuery(folderId, realm = realm))
     }
+
+    fun fetchIncompleteMessages(thread: Thread) {
+        RealmDatabase.mailboxContent().writeBlocking {
+            thread.messages.forEach { localMessage ->
+                if (!localMessage.fullyDownloaded) {
+                    ApiRepository.getMessage(localMessage.resource).data?.also { remoteMessage ->
```
KevinBoulongne, line 127: No API calls in Controllers.

sirambd: In this case, we can leave it as it is realm that drives the api call

app/src/main/java/com/infomaniak/mail/workers/SyncMessagesWorker.kt
```diff
@@ -83,7 +85,13 @@ class SyncMessagesWorker(appContext: Context, params: WorkerParameters) : BaseCo
         mailboxInfoRealm.close()
     }
 
-    private fun Thread.showNotification(userId: Int, mailbox: Mailbox, unReadThreadsCount: Int, realm: Realm) {
+    private fun Thread.showNotification(
+        userId: Int,
+        mailbox: Mailbox,
+        unReadThreadsCount: Int,
+        realm: Realm,
+        okHttpClient: OkHttpClient
```
KevinBoulongne, line 93: Remove unused `okHttpClient`.

app/src/main/java/com/infomaniak/mail/workers/SyncMessagesWorker.kt
```diff
@@ -83,7 +85,13 @@ class SyncMessagesWorker(appContext: Context, params: WorkerParameters) : BaseCo
...
+    ) {
```
KevinBoulongne, line 94: Revert this, `okHttpClient` is unused.

sirambd: no it's always used

PR #498, opened 2023-01-19
Fix topbar not updating

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -96,6 +96,11 @@ fun <T> LiveData<T?>.observeNotNull(owner: LifecycleOwner, observer: (t: T) -> U
     observe(owner) { it?.let(observer) }
 }
 
+inline fun <reified T> LiveData<T>.continuouslyObserve(viewLifecycleOwner: LifecycleOwner, noinline observer: (T) -> Unit) {
```
sirambd, line 99: What bothers me about this name is that it sounds like you're listening continuously, but you know that's not true because it's lifecycle sensitive, I'd prefer something like `refreshObserve` or something more meaningful.

sirambd (changes requested):

JorisBodin (approved):

PR #500, opened 2023-01-19
Use Realm `queries` to get Messages for Actions functions

KevinBoulongne:
Depends on #495

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
@@ -120,7 +120,10 @@ object MessageController {
     }
 
     fun getSpamMessages(thread: Thread): List<Message> {
-        return getMessagesAndDuplicates(thread) { message -> !message.scheduled && !message.from.first().isMe() }
+        val isNotScheduled = "${Message::scheduled.name} == false"
+        val isNotFromMe = "SUBQUERY(${Message::from.name}, \$recipient, " +
+                "\$recipient.${Recipient::email.name} != '${AccountUtils.currentMailboxEmail}').@count > 0"
+        return getMessagesAndDuplicates(thread, "$isNotScheduled AND $isNotFromMe")
     }
 
     fun getDeletableMessages(thread: Thread): List<Message> {
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
@@ -114,7 +114,9 @@ object MessageController {
     }
 
     fun getArchivableMessages(thread: Thread, folderId: String): List<Message> {
-        return getMessagesAndDuplicates(thread) { message -> message.folderId == folderId && !message.scheduled }
+        val byFolderId = "${Message::folderId.name} == '$folderId'"
+        val isNotScheduled = "${Message::scheduled.name} == false"
+        return getMessagesAndDuplicates(thread, "$byFolderId AND $isNotScheduled")
     }
 
     fun getSpamMessages(thread: Thread): List<Message> {
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
@@ -124,7 +124,8 @@ object MessageController {
     }
 
     fun getDeletableMessages(thread: Thread): List<Message> {
-        return getMessagesAndDuplicates(thread) { message -> !message.scheduled }
+        val isNotScheduled = "${Message::scheduled.name} == false"
+        return getMessagesAndDuplicates(thread, isNotScheduled)
     }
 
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -269,7 +269,7 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
             SwipeAction.QUICKACTIONS_MENU -> {
                 safeNavigate(
                     ThreadListFragmentDirections.actionThreadListFragmentToThreadActionsBottomSheetDialog(
-                        messageUidToReplyTo = MessageController.getMessageUidToReplyTo(thread.messages),
+                        messageUidToReplyTo = MessageController.getMessageToReplyTo(thread.messages).uid,
```
sirambd, line 272: it should not be done here, but directly where it is needed, i.e. in the destination fragment.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MessageActionsBottomSheetDialog.kt
```diff
@@ -41,36 +41,38 @@ class MessageActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         val threadUid = navigationArgs.threadUid
         val messageUid = navigationArgs.messageUid
 
-        messageActionsViewModel.threadAndMessage(threadUid, messageUid).observe(viewLifecycleOwner) { (thread, message) ->
+        messageActionsViewModel.threadAndMessage(threadUid, messageUid).observe(viewLifecycleOwner) {
+
+            val (thread, message) = it
```
sirambd, line 46: Why did you change it?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,18 +88,55 @@ object MessageController {
...
 
-    fun getMessageUidToReplyTo(messages: RealmList<Message>): String = getMessageToReplyTo(messages).uid
+    private fun getMessagesAndDuplicates(thread: Thread, condition: String): List<Message> {
+
+        val messages = thread.messages.query(condition).find()
+
+        val byMessagesIds = "${Message::messageId.name} IN {${messages.map { it.messageId }.joinToString { "'$it'" }}}"
+        val duplicates = thread.duplicates.query(byMessagesIds)
+
+        return messages + duplicates.find()
```
sirambd, line 134:
```suggestion
        val byMessagesIds = "${Message::messageId.name} IN {${messages.joinToString { "'${it.messageId}'" }}}"
        val duplicates = thread.duplicates.query(byMessagesIds).find()

        return messages + duplicates
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,18 +88,55 @@ object MessageController {
...
+        return getMessagesAndDuplicates(thread, "$isNotScheduled AND $isNotFromMe")
+    }
+
+    fun getDeletableMessages(thread: Thread): List<Message> {
+        return getMessagesAndDuplicates(thread, isNotScheduled)
+    }
+
+    fun getPermanentlyDeletableMessages(thread: Thread): List<Message> {
+        return thread.messages + thread.duplicates
+    }
```
sirambd, line 120: It's not a query, you can move it to `Thread` class

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,18 +88,55 @@ object MessageController {
...
+        return messages.query("$isNotDraft AND $isNotFromMe").find().lastOrNull()
+            ?: messages.query(isNotDraft).find().lastOrNull()
+            ?: messages.last()
+    }
 
-        val isNotFromMe = "SUBQUERY(${Message::from.name}, \$recipient, " +
-                "\$recipient.${Recipient::email.name} != '${AccountUtils.currentMailboxEmail}').@count > 0"
+    fun getUnseenMessages(thread: Thread): List<Message> {
+        val isUnseen = "${Message::seen.name} == false"
+        return getMessagesAndDuplicates(thread, isUnseen)
```
sirambd, line 98:
```suggestion
        return getMessagesAndDuplicates(thread, "${Message::seen.name} == false")
```

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
KevinBoulongne: I'd rather let it stay here, with the other similar functions. All actions functions that return a list of Messages to execute the action on, are here.

sirambd: 🤔 i understand but i still find this case special, limit since it is used only once, put it directly where it is called

sirambd (approved):

PR #527, opened 2023-01-27
Correctly refresh UI Draft when coming back to app

KevinBoulongne:
Depends on #526


app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -104,6 +104,11 @@ class NewMessageFragment : Fragment() {
         observeNewAttachments()
     }
 
+    override fun onResume() {
+        super.onResume()
+        newMessageViewModel.updateDraftInLocalIfRemoteHasChanged()
```
sirambd, line 109: Putting it in the start is not enough?

sirambd (changes requested):

sirambd (approved):

PR #542, opened 2023-02-02
fix: ForegroundServiceStartNotAllowedException

sirambd:
fix sentry https://sentry.infomaniak.com/share/issue/fd357ea10c0f4ce891480b989d44688c/

**Description**
> Apps that target Android 12 (API level 31) or higher can't start foreground services while running in the background, except for [a few special cases](https://developer.android.com/guide/components/foreground-services#background-start-restriction-exemptions). If an app tries to start a foreground service while the app runs in the background, and the foreground service doesn't satisfy one of the exceptional cases, the system throws a [ForegroundServiceStartNotAllowedException](https://developer.android.com/reference/android/app/ForegroundServiceStartNotAllowedException).

https://developer.android.com/guide/components/foreground-services#background-start-restrictions

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
+++ b/app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
@@ -18,6 +18,7 @@
 package com.infomaniak.mail.workers
 
 import android.content.Context
+import android.util.Log
 import androidx.lifecycle.LiveData
 import androidx.work.*
 import com.infomaniak.lib.core.api.ApiController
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -100,6 +105,8 @@ class DraftsActionsWorker(appContext: Context, params: WorkerParameters) : BaseC
 
             val drafts = DraftController.getDraftsWithActions(realm = this).ifEmpty { return@writeBlocking false }
 
+            Log.d(TAG, "handleDraftsActions: ${drafts.count()} draft to handle")
```
KevinBoulongne, line 108:
```suggestion
            Log.d(TAG, "handleDraftsActions: ${drafts.count()} draft(s) to handle")
```

KevinBoulongne (approved):

JorisBodin (approved):

PR #580, opened 2023-02-13
Fix `markAsReadUnread` back navigation when closed

KevinBoulongne:
`markAsReadUnread` for a Message now correctly closes the BottomSheetDialog
`markAsReadUnread` for a Thread now pop to the entry before ThreadFragment (could be ThreadListFragment or SearchFragment)


LunarX (approved):

FabianDevel (changes requested):
We also need to change the popBackStack of ThreadActionList to avoid force back to ThreadListFragment, because it will conflict with the search

LunarX (approved):

FabianDevel (approved):

PR #584, opened 2023-02-13
Make open with attachment

sirambd:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+    private fun Response.saveAttachmentTo(outputFile: File): Boolean {
+        BufferedInputStream(body?.byteStream() ?: return false).use { input ->
+            outputFile.parentFile?.mkdirs()
+            outputFile.outputStream().use { output -> input.copyTo(output) }
+        }
+        return true
+    }
+
+    override fun onCleared() {
+        // If we end up with an incomplete cached viewModel we delete it
```
KevinBoulongne, line 77: Isn'it an "incomplete cached Attachment" instead of "viewModel" ?

sirambd: Yes I made a mistake, it's attachement instead of viemodel.

KevinBoulongne:
```suggestion
        // If we end up with an incomplete cached Attachment, we delete it
```

KevinBoulongne (approved):

JorisBodin (approved):

PR #591, opened 2023-02-15
Add sentry log and fix app crash when messages are duplicated in realm

FabianDevel:


KevinBoulongne (approved):

FabianDevel:
Issue will be addressed in #597

PR #623, opened 2023-02-22
Search Threads utils

sirambd:
**Features**
Search logic
- Search from remote + pagination
- Search from local in (subject, preview, body)
- Handle and notify ui visibility mode

PR #624, opened 2023-02-22
Fix folders backlink not being single because of search folder

LunarX:
Depends on #623


KevinBoulongne (approved):

sirambd (approved):

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -126,7 +127,7 @@ class Message : RealmObject {
     val threadsDuplicatedIn by backlinks(Thread::duplicates)
 
     private val _folders by backlinks(Folder::messages)
-    val folder get() = _folders.single()
+    val folder get() = _folders.single { if (_folders.count() == 1) true else it.id != SEARCH_FOLDER_ID }
```
FabianDevel, line 130:
```suggestion
    val folder get() = _folders.single { _folders.count() == 1 || it.id != SEARCH_FOLDER_ID }
```

FabianDevel (approved):

PR #639, opened 2023-02-24
Fetch Folders after saving a Draft, so it automagically displays itself in ThreadFragment

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -203,6 +204,7 @@ class DraftsActionsWorker(appContext: Context, params: WorkerParameters) : BaseC
                         it.messageUid = data?.messageUid
                         it.action = null
                     }
+                    scheduledDate = dateFormatWithTimezone.format(Date())
```
LunarX, line 207: Why did this fix the issue ? Didn't sending null amount to the same ?

KevinBoulongne: Yes, sending `null` amounts to the same result, but the code will be more complicated to understand.

sirambd (changes requested):

LunarX (approved):

JorisBodin (approved):

PR #657, opened 2023-03-01
API call when flushing a Folder

KevinBoulongne:
Depends on #659

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -240,6 +240,16 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
         refreshThreads(folderId = folderId)
     }
 
+    fun flushFolder() = viewModelScope.launch(Dispatchers.IO) {
+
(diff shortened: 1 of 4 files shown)


LunarX (approved):

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -240,6 +240,16 @@ class MainViewModel(application: Application) : AndroidViewModel(application) {
...
     }
 
+    fun flushFolder() = viewModelScope.launch(Dispatchers.IO) {
+
+        val isSuccess = ApiRepository.flushFolder(
+            mailboxUuid = currentMailbox.value?.uuid ?: return@launch,
+            folderId = currentFolderId ?: return@launch,
+        ).isSuccess()
+
+        if (isSuccess) forceRefreshThreads()
```
FabianDevel, line 250: Don't we need an error case to inform user ?

sirambd, line 250: show error snackbar if an error has been occurred

JorisBodin (approved):

PR #678, opened 2023-03-06
fix: Missing message when we split a message

sirambd:


app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -101,7 +101,6 @@ object MessageBodyUtils {
                 htmlDocumentWithQuote.toString() to firstQuotedContent.toString()
             }
```
KevinBoulongne, line 103:
```suggestion
```

KevinBoulongne (approved):

PR #717, opened 2023-03-14
Add previous Message history when composing a new mail

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageViewModel.kt
```diff
@@ -140,18 +140,30 @@ class NewMessageViewModel(application: Application) : AndroidViewModel(applicati
     private fun splitSignatureAndQuoteFromBody() {
 
         fun Document.split(divClassName: String, defaultValue: String): Pair<String, String?> {
-            return getElementsByClass(divClassName).lastOrNull()?.let {
+            return getElementsByClass(divClassName).firstOrNull()?.let {
```
KevinBoulongne, line 143: Why was it `lastOrNull()` ? @sirambd

Do you think it's ok if we use `firstOrNull()` here ? Because it doesn't seem to work the right way with `lastOrNull()`.

JorisBodin (approved):

sirambd (approved):

PR #735, opened 2023-03-16
Correctly execute Draft's action if the Draft has become obsolete while editing it

KevinBoulongne:


sirambd (changes requested):
Use a realm query for this

sirambd (approved):

LunarX (approved):

PR #742, opened 2023-03-16
Add AppLock feature

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/45/files

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -78,6 +78,7 @@ android {
 dependencies {
     implementation project(path: ':Core')
     implementation project(path: ':Core:BugTracker')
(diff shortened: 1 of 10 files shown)


app/build.gradle
KevinBoulongne, line 1: This feature doesn't work on our Xiaomi test phone.

KevinBoulongne (changes requested):

app/build.gradle
FabianDevel: It isn't link to the android version, it works well on a Honor with same AI, it will need further investigations on specific behaviour of xiaomi

sirambd (approved):

PR #744, opened 2023-03-16
Add additional Matomo trackers

FabianDevel:


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
```diff
@@ -141,6 +142,8 @@ class IntroFragment : Fragment() {
                 if (newSelectedAccentColor == localSettings.accentColor) return
                 localSettings.accentColor = newSelectedAccentColor
                 triggerUiUpdateWhenAnimationEnd(newSelectedAccentColor)
+
+                trackOnBoardingEvent("switchColor", newSelectedAccentColor == BLUE)
```
LunarX, line 146: Later we will have more than just pink and blue with the system color. Shouldn't we send either the amount of time people change the theme with no regard for the choice of "BLUE" or maybe the color to which they send so that it becomes future proof for the system theme?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/login/IntroFragment.kt
FabianDevel: Well the number of time people change will always be right, the value is just an extra data that could not be use if we don't want to. We can suppress this value when we will add the system themen but for the moment we are iso with iOS and it's a good data to know for the first users, even if we delete it later, because adding the 3rd theme would make this stat useless.

FabianDevel: To avoid making this merged average, we will have to make 3 diffferent track calls

LunarX (approved):

PR #748, opened 2023-03-17
Split Outlook quote + Improve split blockquote logic

sirambd:


KevinBoulongne (changes requested):

FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -55,61 +56,78 @@ object MessageBodyUtils {
...
 
-        handleBlockQuote(htmlDocumentWithoutQuote, htmlQuotes)
-        val currentQuoteDescriptor = handleQuoteDescriptors(htmlDocumentWithoutQuote, htmlQuotes).ifEmpty { blockquote }
+        val blockquoteElement = findAndRemoveLastParentBlockquote(htmlDocumentWithoutQuote)
+        val currentQuoteDescriptor = findFirstKnownParentQuoteDescriptor(htmlDocumentWithoutQuote).ifEmpty {
+            if (blockquoteElement == null) "" else blockquote
+        }
 
-        val (body, quote) = splitBodyAndQuote(htmlQuotes, htmlDocumentWithQuote, currentQuoteDescriptor)
+        val (body, quote) = splitBodyAndQuote(blockquoteElement, htmlDocumentWithQuote, currentQuoteDescriptor)
```
LunarX, line 65: Shouldn't both of `splitBodyAndQuote` quote related arguments be next to each other rather than one being at the first position and one being at the third position ?

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -55,61 +56,78 @@ object MessageBodyUtils {
 
         val htmlDocumentWithQuote = Jsoup.parse(initialBody.value)
         val htmlDocumentWithoutQuote = Jsoup.parse(initialBody.value)
-        val htmlQuotes = mutableListOf<Element>()
 
-        handleBlockQuote(htmlDocumentWithoutQuote, htmlQuotes)
-        val currentQuoteDescriptor = handleQuoteDescriptors(htmlDocumentWithoutQuote, htmlQuotes).ifEmpty { blockquote }
+        val blockquoteElement = findAndRemoveLastParentBlockquote(htmlDocumentWithoutQuote)
```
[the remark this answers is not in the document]
sirambd: No worries, because it will be this function that will remove the quotes

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
sirambd: The order does not matter, because there is not necessarily a link on these parameters of how they are managed. 
But no worries about changing it.

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -39,13 +40,13 @@ object MessageBodyUtils {
...
-        ".zmail_extra", // Zoho
+        anyCssClassContaining("gmail_extra"),
+        anyCssClassContaining("gmail_quote"),
+        anyCssClassContaining(INFOMANIAK_REPLY_QUOTE_HTML_CLASS_NAME),
+        anyCssClassContaining("moz-cite-prefix"),
+        anyCssClassContaining("protonmail_quote"),
+        anyCssClassContaining("yahoo_quoted"),
+        anyCssClassContaining("zmail_extra"), // Zoho
         "[name=\"quote\"]", // GMX
         "blockquote[type=\"cite\"]",
```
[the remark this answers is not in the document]
sirambd: No idea who it can be this client, but if the standard is well respected then theoretically the algo on blockquotes should be enough, otherwise there will be a bug 🤔.
However, it can be deleted for the time being

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -39,77 +40,97 @@ object MessageBodyUtils {
...
             }
-        } ?: (htmlDocumentWithQuote.toString() to null)
+        }
+    }
+
+    //region Utils
+    /**
+     * Some email clients rename css classes to prefix them for example.
+     * We match all the css classes that contain the quote, in case this one has been renamed.
+     * @return a new css query
```
KevinBoulongne, line 121:
```suggestion
     * Some Email clients rename CSS classes to prefix them.
     * We match all the CSS classes that contain the quote, in case this one has been renamed.
     * @return a new CSS query
```

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -39,77 +40,97 @@ object MessageBodyUtils {
...
+     */
+    private fun anyCssClassContaining(cssClass: String) = "[class*=$cssClass]"
+
+    /**
+     * Some mail clients add the history in a new block, at the same level as the old one.
+     * And so we match the current block, as well as all those that follow and that are at the same level
+     * @return [Elements] which contains all the blocks that have been matched
```
KevinBoulongne, line 128:
```suggestion
     * Some Email clients add the Thread's History in a new block, at the same level as the previous one.
     * So we match the current block, as well as all those that follow and that are at the same level.
     * @return [Elements] which contains all the blocks that have been matched
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #776, opened 2023-03-27
Block adding more recipient with a snackbar if limit has been reached

LunarX:


app/src/main/res/values/strings.xml
```diff
@@ -345,6 +345,7 @@
     <string name="titleMyAccount">My account</string>
     <string name="titleMyAccounts">My accounts</string>
     <string name="toTitle">To:</string>
+    <string name="tooManyRecipients">You can’t add this address because you have reached the limit of 99 recipients</string>
```
KevinBoulongne, line 348: Magic number.

Maybe use a parameter instead, with `MAX_ALLOWED_RECIPIENT` ?

KevinBoulongne (changes requested):

JorisBodin (approved):

PR #782, opened 2023-03-28
Use `?attr/isLightTheme` for `windowLightStatusBar` item

KevinBoulongne:


FabianDevel (commented):
.

app/src/main/res/values/themes.xml
FabianDevel, line 1: Could it be use for android:windowLightNavigationBar also ?

LunarX (approved):

PR #829, opened 2023-04-04
Greys out already used emails when auto completing recipients

LunarX:
Depends on #826


app/src/main/res/drawable/ic_check_rounded.xml
```diff
@@ -0,0 +1,27 @@
+<!--
```
KevinBoulongne, line 1:
```suggestion
<?xml version="1.0" encoding="utf-8"?><!--
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #834, opened 2023-04-05
Make it clear that `background_popup` color should not be changed

LunarX:


KevinBoulongne (approved):

FabianDevel (changes requested):
Need further explanation

PR #850, opened 2023-04-11
Fix `inline` Attachment loaded several times

sirambd:


app/src/main/java/com/infomaniak/mail/utils/LocalStorageUtils.kt
```diff
@@ -55,24 +54,31 @@ object LocalStorageUtils {
...
         }
 
-        return ApiRepository.downloadAttachment(resource).saveAttachmentTo(resource, cacheFile)
+        return ApiRepository.downloadAttachment(resource).saveAttachmentTo(cacheFile)
     }
 
-    fun saveCacheAttachment(resource: String, inputStream: InputStream, outputFile: File) = with(outputFile) {
+    /**
+     * Saves the attachment in the disk memory.
+     * The file remains unreadable as long as it is in processing
```
KevinBoulongne, line 70:
```suggestion
     * Save the Attachment in disk memory.
     * The file remains unreadable as long as it's being processed.
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #875, opened 2023-04-18
Search: Only save previous state when opening Thread, and not in `onStop()`

KevinBoulongne:


JorisBodin (approved):

app/src/main/java/com/infomaniak/mail/ui/main/search/SearchFragment.kt
```diff
@@ -239,21 +229,30 @@ class SearchFragment : Fragment() {
         }
     }
 
-    private fun setMessagesUi() {
-        binding.mailRecyclerView.apply {
+    private fun setMessagesUi() = with(binding) {
+
+        mailRecyclerView.adapter = threadListAdapter.apply {
```
FabianDevel, line 234: Why removing the adapter from ` mailRecyclerView.apply {}` ?

KevinBoulongne: Because it's an `apply` in an `apply` in a `with`, so it's unreadable.

FabianDevel (approved):

PR #881, opened 2023-04-19
Fix "alpha must be between 0 and 255"

JorisBodin:


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -352,7 +357,7 @@ class ThreadFragment : Fragment() {
     }
 
     private companion object {
-        const val COLLAPSE_TITLE_THRESHOLD = 0.5
+        const val COLLAPSE_TITLE_THRESHOLD = 0.666
```
LunarX, line 360: Why is this threshold change better or needed ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -117,12 +118,16 @@ class ThreadFragment : Fragment() {
 
         val defaultTextColor = context.getColor(R.color.primaryTextColor)
         appBar.addOnOffsetChangedListener { appBarLayout, verticalOffset ->
-            val total = appBarLayout.height * COLLAPSE_TITLE_THRESHOLD
-            val removed = appBarLayout.height - total
-            val progress = (((-verticalOffset.toFloat()) - removed).coerceAtLeast(0.0) / total).toFloat() // Between 0 and 1
-            val opacity = (progress * 255).roundToInt()
 
+            val subjectHeight = appBarLayout.height.toDouble()
```
LunarX, line 122: This `toDouble()` is not needed as both other lines will still be double due to computation always involving a double. Do we really want it?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -117,12 +118,16 @@ class ThreadFragment : Fragment() {
...
+            val impactingHeight = subjectHeight * COLLAPSE_TITLE_THRESHOLD
+            val nonImpactingHeight = subjectHeight - impactingHeight
+
+            val absoluteProgress = verticalOffset.absoluteValue.toDouble() - nonImpactingHeight
```
LunarX, line 126: This `toDouble()` is also not needed, same question

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: With `0.5`, there is some frames where the subject is totally hidden.
It feels more smooth this way.
image

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: For the sake of clarity, I prefer to write it this way.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: https://github.com/Infomaniak/android-kMail/pull/881#discussion_r1172454057

LunarX (approved):

PR #909, opened 2023-05-02
Stop the Threads algo when a API call fails

KevinBoulongne:


FabianDevel (changes requested):
It looks like it's missing some runCatching

PR #931, opened 2023-05-09
fix: Application with hilt doesn't work in fdroid flavor

sirambd:
**Changes**
ApplicationMain has been separated, so that each flavor can do its processing in its own applicationMain instance

- fix: fdroid flavor with hilt
- fix: draft injection with hilt

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/workers/BaseProcessMessageNotificationsWorker.kt
+++ b/app/src/main/java/com/infomaniak/mail/workers/BaseProcessMessageNotificationsWorker.kt
@@ -15,11 +15,22 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package com.infomaniak.mail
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/infomaniak/mail/workers/BaseProcessMessageNotificationsWorker.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+package com.infomaniak.mail.workers
+
+import android.content.Context
+import androidx.work.WorkManager
+import androidx.work.WorkerParameters
+
+abstract class BaseProcessMessageNotificationsWorker(
+    appContext: Context,
+    params: WorkerParameters
```
KevinBoulongne, line 26:
```suggestion
    params: WorkerParameters,
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #937, opened 2023-05-10
Custom processing of emails to better fit on a smartphone screen

LunarX:
Depends on #936


sirambd (changes requested):
I approve this but check code smell

sirambd (approved):

PR #946, opened 2023-05-12
Auto hide/show "Load more" button

KevinBoulongne:
Depends on #963

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
@@ -422,13 +422,13 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
     private fun observeDownloadState() {
         mainViewModel.isDownloadingChanges
             .distinctUntilChanged()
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -148,6 +148,11 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
         updateSwipeActionsAccordingToSettings()
     }
 
+    override fun onPause() {
+        mainViewModel.isDownloadingChanges.value = false to null
+        super.onPause()
+    }
```
JorisBodin, line 154: Seems to be unnecessary

JorisBodin (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
KevinBoulongne: Not unnecessary, but not the right fix too.

JorisBodin (approved):

PR #954, opened 2023-05-16
Update Signatures Sentry to make it trigger more often

KevinBoulongne:


LunarX (approved):

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -346,7 +348,39 @@ class MainViewModel @Inject constructor(
...
+        val data = apiResponse.data
+        val signatures = data?.signatures
+
+        val defaultSignaturesCount = signatures?.count { it.isDefault } ?: -1
+        when {
+            data == null -> Sentry.withScope { scope ->
+                scope.level = SentryLevel.ERROR
+                scope.setExtra("email", AccountUtils.currentMailboxEmail.toString())
+                Sentry.captureMessage("Signature: The call to get Signatures returned a `null` data")
+            }
```
FabianDevel, line 362: We should wrap this in `isSuccess()` otherwise we will have this sentry every time the call fails. 
Or add more information on why it failed

PR #957, opened 2023-05-22
Block images in Spam folder with no regard for the user setting + Default to "Always show images"

LunarX:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -58,7 +57,9 @@ import org.jsoup.Jsoup
 import java.util.*
 import com.google.android.material.R as RMaterial
 
-class ThreadAdapter : RecyclerView.Adapter<ThreadViewHolder>(), RealmChangesBinding.OnRealmChanged<Message> {
+class ThreadAdapter(
+    private val shouldLoadDistantResources: Boolean
```
KevinBoulongne, line 61:
```suggestion
    private val shouldLoadDistantResources: Boolean,
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #971, opened 2023-05-26
mailto: open kMail directly

JorisBodin:


LunarX (changes requested):
Using the SENDTO intent to open NewMessageActivity won't work with the future "saving draft" snackbar PR. The activity should be opened with a requestForResult as required by the snackbar PR. I'll do the necessary changes next week

LunarX (approved):
Ok the bug is not that big, I'll fix it later anyway but this can still be merged

PR #974, opened 2023-05-26
Users can change signatures

FabianDevel:


LunarX:
Making an api call before updating the select signature UI in the settings makes it feel very unresponsive. Could we rather make the api call only once when leaving the setting ?

app/src/main/res/layout/fragment_signature_setting.xml
KevinBoulongne, line 1: We should display the preview of a signature when it's selected, to know what we are doing.

app/src/main/java/com/infomaniak/mail/ui/main/settings/mailbox/SignatureSettingAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+        return SettingsSignatureViewHolder(
+            ItemSettingsSignatureBinding.inflate(LayoutInflater.from(parent.context), parent, false)
+        )
+    }
+
+    override fun onBindViewHolder(holder: SettingsSignatureViewHolder, position: Int): Unit = with(holder.binding.root) {
+        val signature = signatures[position]
+        setText(signature.name)
+        if (signature.isDefault) check() else uncheck()
+        setOnClickListener { onSignatureSelected(signature) }
```
KevinBoulongne, line 45: Don't allow to click on the already selected signature.

app/src/main/java/com/infomaniak/mail/ui/main/settings/mailbox/SignatureSettingFragment.kt
LunarX, line 1: Do we really want to sync the selected signature setting with the api ? I don't think this is a good idea because you might want to have a different signature for each case. If you REALLY want to use the same default signature on both platforms you could always set it once and then you'd be good.

If we still go through with it, making an api call before updating the select signature UI in the settings makes it feel very unresponsive. Could we rather make the api call only once when leaving the setting ?

app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
KevinBoulongne, line 1: I'm not sure we want to update the default Signature for the Webmail too.

KevinBoulongne (changes requested):

KevinBoulongne (changes requested):

app/src/main/res/layout/item_settings_signature.xml
```diff
@@ -0,0 +1,23 @@
...
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<com.infomaniak.mail.ui.main.settings.SettingRadioButtonView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    tools:text="@tools:sample/lorem" />
+
```
KevinBoulongne, line 23:
```suggestion
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/settings/mailbox/SignatureSettingFragment.kt
```diff
@@ -22,11 +22,59 @@ import android.view.LayoutInflater
...
+            setupAdapter(mailbox)
+            mainViewModel.updateSignatures(mailbox, signatureSettingViewModel.customRealm)
+            observeSignatures()
+        }
+    }
+
+    private fun setupAdapter(mailbox: Mailbox) {
+        binding.signatureList.adapter = SignatureSettingAdapter(
+            canManageSignature = mailbox.permissions?.canManageSignatures ?: false,
+            onSignatureSelected = ::onSignatureClicked
```
KevinBoulongne, line 65:
```suggestion
            onSignatureSelected = ::onSignatureClicked,
```

app/src/main/java/com/infomaniak/mail/ui/main/settings/SettingRadioButtonView.kt
```diff
@@ -83,4 +89,13 @@ class SettingRadioButtonView @JvmOverloads constructor(
...
     }
+
+    fun setOnClickListener(listener: (() -> Unit)?) {
+        onClickListener = listener
+    }
+
+    fun enable(mustEnable: Boolean) = binding.root.apply {
+        isClickable = mustEnable
+        isEnabled = mustEnable
+    }
```
KevinBoulongne, line 100: Doesn't seem to work.
Possibly because Realm is flushed and all Signatures are inserted again at each click.
If we remove the API calls _(as discussed in another comment)_, it will possibly resolve this issue.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/settings/SettingRadioButtonView.kt
LunarX: Try to remove `isClickable` I think `isEnabled` should be enough

app/src/main/java/com/infomaniak/mail/ui/main/settings/mailbox/SignatureSettingAdapter.kt
LunarX: I think the issue has been fixed from setting radio groups but here it doesn't work because it uses a recyclerview instead

app/src/main/res/layout/fragment_signature_setting.xml
FabianDevel: This need to be discussed with UX. For now we need to get iso with iOS

app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
FabianDevel: Need further discussions, for now need to get iso with iOS

app/src/main/java/com/infomaniak/mail/ui/main/settings/mailbox/SignatureSettingFragment.kt
FabianDevel: We will still keep the API call on click like it's generally the case at this moment in the app. It's also ISO with iOS.
We need further discussions to know if the settings must be local or no, but for now we need to have a minimal version of this feature and be ISO with iOS.

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #975, opened 2023-05-30
Fix `NoSuchElementException` : List is empty

JorisBodin:


KevinBoulongne (approved):

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -127,7 +127,7 @@ class FetchMessagesManager @Inject constructor(
 
         ThreadController.fetchIncompleteMessages(messages, mailbox, okHttpClient, realm)
 
-        val message = ThreadController.getThread(uid, realm)?.messages?.last() ?: return
+        val message = ThreadController.getThread(uid, realm)?.messages?.lastOrNull() ?: return
```
KevinBoulongne, line 130: This won't fix the issue.
It will just make the app not crash, but the user won't have the notification.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
KevinBoulongne: Can't find the reason, so failed silently for now, and added Sentry.

KevinBoulongne (approved):

PR #986, opened 2023-06-05
Snackbar when sending an email

LunarX:
Depends on #985

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
@@ -140,8 +140,13 @@ class MainActivity : ThemedActivity() {
 
         draftsActionsWorkerScheduler.getRunningWorkInfoLiveData().observe(this) {
             it.forEach { workInfo ->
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -154,9 +159,16 @@ class MainActivity : ThemedActivity() {
...
+                draftAction?.let {
+                    if (draftAction == DraftAction.SAVE.name) {
+                        showSavedDraftSnackBar(remoteDraftUuid!!, associatedMailboxUuid!!)
+                    } else {
+                        showSentDraftSnackBar()
+                    }
+                }
             }
         }
     }
```
KevinBoulongne: I've reworked this part.
You can see it here: https://github.com/Infomaniak/android-kMail/pull/986/commits/3aeb1a89e4a541d3281f30d67e99334e0179b84b
But it'll be probably easier to read it in Android Studio.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
sirambd: You should update your commit in case it hasn't been pushed yet. And we merge this

PR #1006, opened 2023-06-08
Add Sentry if we can't decode the API result when sending an email

JorisBodin:


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -342,11 +342,21 @@ class DraftsActionsWorker @AssistedInject constructor(
...
+                when {
+                    isSuccess() -> {
+                        scheduledDate = data?.scheduledDate
+                        realm.delete(draft)
+                    }
+                    error?.exception is SerializationException -> {
+                        realm.delete(draft)
+                        Sentry.withScope { scope ->
+                            scope.level = SentryLevel.ERROR
+                            Sentry.captureMessage("Json api return is modified")
```
KevinBoulongne, line 354:
```suggestion
                            Sentry.captureMessage("Return JSON for SendDraft API call was modified")
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1029, opened 2023-06-14
Synchronize ViewModel & Realm drafts

FabianDevel:

PR #1030, opened 2023-06-14
Keep Draft data in UI when recreating the Activity

LunarX:
Depends on #1029


app/src/main/java/com/infomaniak/mail/ui/main/newMessage/NewMessageFragment.kt
```diff
@@ -128,6 +128,8 @@ class NewMessageFragment : Fragment() {
         observeNewAttachments()
         observeCcAndBccVisibility()
         observeDraftWorkerResults()
+
+        newMessageViewModel.activityRecreated = true
```
KevinBoulongne, line 132: The value of this data is wrong at this time.

Maybe use a count instead ?

KevinBoulongne (changes requested):

PR #1039, opened 2023-06-19
Add the settings dividers back

FabianDevel:


app/src/main/res/values/dimens.xml
```diff
@@ -41,6 +41,9 @@
 
     <dimen name="menuDrawerItemViewHeight">48dp</dimen>
 
+    <!-- Settings -->
+    <dimen name="settingsSeparatorHorizontalPadding">8dp</dimen>
```
LunarX, line 45: I don't think specs mention 8dp. It should be 16dp instead
```suggestion
    <dimen name="settingsSeparatorHorizontalPadding">16dp</dimen>
```

LunarX (changes requested):

PR #1042, opened 2023-06-20
Fix `null` Mailbox at boot

KevinBoulongne:
Depends on #1041

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -227,19 +227,21 @@ class MainViewModel @Inject constructor(
                 }
             }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -249,7 +249,7 @@ class MainViewModel @Inject constructor(
                 }
             }
 
-            val mailbox = currentMailbox.value!!
+            val mailbox = currentMailbox.value ?: return@launch
```
FabianDevel, line 252: We should talk about that in meeting to know what to do for know, because it's probably not good to let the user on a "bugged" screen without mailbox

FabianDevel (changes requested):

PR #1044, opened 2023-06-20
In Search, we don't loose Messages bodies anymore

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -127,22 +128,24 @@ object ThreadController {
...
+                        it.isAnswered = remoteMessage.isAnswered
+                        it.isFavorite = remoteMessage.isFavorite
+                        it.isForwarded = remoteMessage.isForwarded
+                        it.isScheduled = remoteMessage.isScheduled
+                        it.isSeen = remoteMessage.isSeen
+                        remoteThread.messages[index] = it
+                    }
                 } ?: run {
-                    remoteMessage.isFromSearch = true
+                    remoteMessage.also {
```
FabianDevel, line 143: use `apply` instead of these `also`

FabianDevel (approved):

PR #1045, opened 2023-06-20
Dagger / Hilt / Realm

sirambd:


app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
```diff
@@ -91,26 +93,16 @@ object RealmDatabase {
...
-        closeMailboxInfo()
         closeUserInfo()
-        closeAppSettings()
-    }
-
-    private fun closeAppSettings() {
-        _appSettings?.close()
-        _appSettings = null
+        _appSettings = null // TODO: To be removed when the injection is done
+        _mailboxInfo = null // TODO: To be removed when the injection is done
```
KevinBoulongne, line 98: Don't we need to close them too, like before?

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1076, opened 2023-06-28
Threads are now refreshed once instead of thrice when coming back to app

KevinBoulongne:
We are now ISO with iOS.

Should fix https://sentry-mobile.infomaniak.com/share/issue/023cac333c1447599562130c31f2dd34/ 🤞


FabianDevel (commented):
We should probably wait for @JorisBodin's return to see if we really want this. 
For the record, iOS doesn't refresh in this case

FabianDevel (changes requested):
We should probably wait for @JorisBodin's return to see if we really want this. 
For the record, iOS doesn't refresh in this case

sirambd (approved):

PR #1081, opened 2023-06-29
Fix deeplink Draft's saving

FabianDevel:


KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -330,18 +330,19 @@ class DraftsActionsWorker @AssistedInject constructor(
         var scheduledDate: String? = null
         var savedDraftUuid: String? = null
 
+        val updatedDraft = DraftController.getDraft(draft.localUuid, mailboxContentRealm)!!
```
KevinBoulongne, line 333: Rename `updatedDraft` into `latestDraft`.

KevinBoulongne (changes requested):

PR #1108, opened 2023-07-06
fix: Problem when returning to the app

sirambd:
When you leave the app and come back very quickly, the realm instance isn't closed yet, so we'll reuse it as long as we can.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
@@ -106,7 +106,7 @@ object RealmDatabase {
 
     //region Close Realms
     fun closeOldRealms() {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
```diff
@@ -106,8 +106,10 @@ object RealmDatabase {
 
     //region Close Realms
     fun closeOldRealms() {
-        oldMailboxContent.get()?.close()
-        oldUserInfo.get()?.close()
+        if (_mailboxContent == null) { // Only closes when mailbox is changed
```
KevinBoulongne, line 109:
```suggestion
        if (_mailboxContent == null) { // Only closes when Mailbox is changed
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1115, opened 2023-07-10
Update InvalidPassword view to new wording

FabianDevel:

PR #1117, opened 2023-07-10
Invalid mailboxes screen

FabianDevel:
Depends on #1115


app/src/main/java/com/infomaniak/mail/views/MenuDrawerItemView.kt
LunarX, line 1: This class is too complexe, I only approve it for now if we simplify it with inheritance in another PR

LunarX (approved):

PR #1119, opened 2023-07-10
Fix messages shortuid

sirambd:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -66,26 +67,38 @@ class ThreadController @Inject constructor(
...
                 val localMessage = MessageController.getMessage(remoteMessage.uid, realm = this)
-                val message = if (localMessage == null) remoteMessage.apply { isFromSearch = true } else remoteMessage
 
-                remoteThread.messagesIds += message.messageIds
-                searchFolder.messages.add(message)
+                remoteMessage.initLocalValues(
+                    date = localMessage?.date ?: remoteMessage.date,
+                    isFullyDownloaded = localMessage?.isFullyDownloaded ?: false,
+                    isSpam = folderRole == Folder.FolderRole.SPAM,
+                    isTrashed = folderRole == Folder.FolderRole.TRASH,
```
LunarX, line 94: Add a comment, somewhere around here, that clearly states that the search only returns message from spam or trash if we explicitly selected those folders which is the reason why we can compute these two values so loosely like that

sirambd: Already explain in kDoc

PR #1122, opened 2023-07-12
Update italian string

FabianDevel:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/JunkBottomSheetDialog.kt
```diff
@@ -65,7 +65,6 @@ class JunkBottomSheetDialog : ActionsBottomSheetDialog() {
             createDescriptionDialog(
                 title = getString(R.string.reportPhishingTitle),
                 description = getString(R.string.reportPhishingDescription),
-                confirmButtonText = R.string.buttonReport,
```
LunarX, line 68: Is this linked to the italian strings updates

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/JunkBottomSheetDialog.kt
FabianDevel: Yes because this string is deleted during strings update

LunarX (approved):

PR #1123, opened 2023-07-12
Fix incorrect folder being selected in search after opening a thread

LunarX:


FabianDevel (approved):

FabianDevel:
This PR looks good, but you could probably make a singleton of the popup to avoid creating it on each onViewCreated. 
That should be done in another PR though

PR #1124, opened 2023-07-12
Update AttachMailbox layout and error management

FabianDevel:


app/src/main/java/com/infomaniak/mail/ui/main/user/AttachMailboxFragment.kt
```diff
@@ -50,14 +54,18 @@ class AttachMailboxFragment : Fragment() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?): Unit = with(binding) {
         super.onViewCreated(view, savedInstanceState)
 
-        mailInput.showKeyboard()
+        mailInput.apply {
+            showKeyboard()
+            manageErrorOnTextChange(mailInputLayout, attachMailboxButton)
+
+            onFocusChangeListener = OnFocusChangeListener { _, hasFocus -> manageErrorOnFocusChange(hasFocus) }
```
LunarX, line 61: Maybe add the info that manageErrorOnFocusChange is only for the email field in its name, something like `manageEmailErrorOnFocusChange`

LunarX (changes requested):

LunarX (approved):

PR #1135, opened 2023-07-17
Fix input layout icon paddings

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/73

### DIFF
--- a/app/src/main/res/layout/bottom_sheet_restore_emails.xml
+++ b/app/src/main/res/layout/bottom_sheet_restore_emails.xml
@@ -43,7 +43,7 @@
             android:layout_margin="@dimen/marginStandardMedium"
             android:text="@string/restoreEmailsText" />
 
-        <com.google.android.material.textfield.TextInputLayout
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/InvalidPasswordFragment.kt
```diff
@@ -53,6 +54,7 @@ class InvalidPasswordFragment : Fragment() {
 
         enterPasswordDescription2.text = getStringWithBoldArg(R.string.enterPasswordDescription2, navigationArgs.mailboxEmail)
 
+        passwordInputLayout.fixIconPaddingIssue()
```
LunarX, line 57: Make a class that inherits from TextInputLayout and automatically applies the fix in its `init {}`

LunarX (changes requested):

KevinBoulongne (approved):

PR #1138, opened 2023-07-17
Mark Message as read when answering to it

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/utils/SharedViewModelUtils.kt
```diff
@@ -46,6 +46,7 @@ class SharedViewModelUtils @Inject constructor(
         message: Message? = null,
         started: (() -> Unit)? = null,
         stopped: (() -> Unit)? = null,
+        shouldRefresh: Boolean = true,
```
sirambd, line 49: Need a comment here

sirambd (approved):

PR #1150, opened 2023-07-19
feat: Register firebase in the app's `onCreate` lifecCycle with a delay

sirambd:


JorisBodin (changes requested):
The app will crash if there are no gplay services

JorisBodin (approved):

PR #1174, opened 2023-07-26
Inject context instead of passing it through every constructors

LunarX:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/DraftController.kt
```diff
@@ -46,7 +46,10 @@ import org.jsoup.nodes.Element
 import org.jsoup.nodes.TextNode
 import javax.inject.Inject
 
-class DraftController @Inject constructor(private val mailboxContentRealm: RealmDatabase.MailboxContent) {
+class DraftController @Inject constructor(
+    private val mailboxContentRealm: RealmDatabase.MailboxContent,
+    private val context: Context,
```
sirambd, line 51:
```suggestion
    private val appContext: Context,
```

sirambd (approved):

PR #1191, opened 2023-08-03
Fix Folders indent & chevron in MenuDrawer

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
```diff
@@ -77,7 +77,9 @@ class Folder : RealmObject {
     @Transient
     var isHistoryComplete: Boolean = DEFAULT_IS_HISTORY_COMPLETE
     @Transient
-    var isCollapsed: Boolean = false
+    var isHidden: Boolean = false // For children only
```
sirambd, line 80: Explain reasons why this state changes

sirambd (changes requested):

sirambd (approved):

PR #1213, opened 2023-08-08
Cancel ThreadsAlgo in Search

KevinBoulongne:
When starting a Search, if the ThreadsAlgo is currently running, we cancel it to avoid overlap with Realm writings.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
@@ -45,11 +45,13 @@ import kotlin.math.max
 @Singleton
 class RefreshController @Inject constructor() {
 
+    private const val FORCE_CANCEL = "this_is_a_forced_cancellation_of_the_threads_algorithm"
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -42,9 +42,15 @@ import kotlin.math.max
 
 object RefreshController {
 
+    private const val FORCE_CANCEL = "this_is_a_forced_cancellation_of_the_threads_algorithm"
+
     private var refreshThreadsJob: Job? = null
 
     //region Fetch Messages
+    fun cancelRefresh() {
+        refreshThreadsJob?.cancel(CancellationException(FORCE_CANCEL))
```
LunarX, line 51: Rather than using a weird string, can't you create your own Exception that extends CancellationException and pass it instead in order to detect it when you catch it ?

LunarX (changes requested):

LunarX (approved):

PR #1224, opened 2023-08-10
feat: Improve signature setting logic

sirambd:


FabianDevel (changes requested):
The navigationArgs of the fragment are now unused and can be removed

FabianDevel (approved):

PR #1194, opened 2023-08-07
Put back `ThreadMode`

KevinBoulongne:
Depends on #1209

PR #1233, opened 2023-08-16
Optimize Threads algo

KevinBoulongne:
Depends on #1194

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
@@ -486,12 +486,17 @@ class RefreshController @Inject constructor(private val localSettings: LocalSett
     ): Set<Thread> {
 
         val threadsToUpsert = mutableSetOf<Thread>()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -486,12 +486,17 @@ class RefreshController @Inject constructor(private val localSettings: LocalSett
...
 
         remoteMessages.forEach { remoteMessage ->
             scope.ensureActive()
 
-            val isMessageAlreadyAdded = folder.messages.firstOrNull { it == remoteMessage && !it.isOrphan() } != null
-            if (isMessageAlreadyAdded) {
+            // Get updated version of existing Message
+            val existingMessage = existingMessages[remoteMessage.uid]?.let {
+                if (it.isManaged()) it else MessageController.getMessage(it.uid, realm = this)
+            }
```
LunarX, line 497: Where do the messages that are unmanaged at this point come from again?

KevinBoulongne: From the `remoteMessage`.

PR #1258, opened 2023-08-21
Search `folderId` can't be `null` anymore

KevinBoulongne:


sirambd (changes requested):
This RP does not correct the initial problem, it risks creating a crash.

JorisBodin (approved):

PR #1260, opened 2023-08-21
Update Signatures if our `identityId` doesn't exist anymore, then fallback on the new default one

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -383,7 +396,13 @@ class DraftsActionsWorker @AssistedInject constructor(
...
                         }
                     }
-                    else -> throwErrorAsException()
+                    else -> {
+                        if (isFirstTime && error?.code == ErrorCode.IDENTITY_NOT_FOUND) {
+                            return updateSignaturesThenRetry(draft, mailboxUuid, okHttpClient)
+                        } else {
+                            throwErrorAsException()
+                        }
+                    }
```
LunarX, line 405: Would it make sens to factorize that part, maybe something like in [Infomaniak/android-kMail/tree/try-to-factorize](https://github.com/Infomaniak/android-kMail/tree/try-to-factorize)

LunarX (changes requested):

LunarX (approved):

PR #1262, opened 2023-08-22
Inject navigation args in viewModels

FabianDevel:


app/src/main/java/com/infomaniak/mail/ui/main/search/SearchViewModel.kt
```diff
@@ -58,6 +56,9 @@ class SearchViewModel @Inject constructor(
 
     private val ioCoroutineContext = viewModelScope.coroutineContext(ioDispatcher)
 
+    /** It's simply used as a default value for the API. */
```
LunarX, line 59: It's not a default value, it's a value that needs to be passed in order pass validation for an API call but which won't be taken into account by the API

LunarX (changes requested):

LunarX (approved):

PR #1266, opened 2023-08-23
Add Breadcrumb in Realm migrations

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
```diff
@@ -17,18 +17,34 @@
...
+    )
     migrationContext.deleteRealmFromFirstMigration()
 }
 
 val MAILBOX_CONTENT_MIGRATION = AutomaticSchemaMigration { migrationContext ->
+    SentryDebug.addMigrationBreadcrumb(
+        realmName = "MailboxContent",
+        oldVersion = migrationContext.oldRealm.version().version,
+        newVersion = migrationContext.newRealm.version().version,
+    )
```
sirambd, line 47: You can simplify everything, as MigrationContext already contains all the necessary information.

sirambd (changes requested):

PR #1278, opened 2023-08-25
Fix email sending due to StyledRecipient

LunarX:


app/src/main/java/com/infomaniak/mail/data/models/correspondent/Recipient.kt
```diff
@@ -33,17 +34,29 @@ open class Recipient : EmbeddedRealmObject, Correspondent {
     override var email: String = ""
     override var name: String = ""
 
+    //region UI data (Transient & Ignore)
```
KevinBoulongne, line 37: Rename all `UI data` regions in the project the same way : `//region UI data (Transient & Ignore)`

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1291, opened 2023-08-28
Rename `ikMail` JS into `kMail`

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -205,7 +205,7 @@ fun WebView.initWebViewClientAndBridge(
     navigateToNewMessageActivity: ((Uri) -> Unit)?,
 ): MessageWebViewClient {
 
-    addJavascriptInterface(WebViewUtils.jsBridge, "ikmail")
+    addJavascriptInterface(WebViewUtils.jsBridge, "mail")
```
KevinBoulongne, line 208: I'd prefer to keep it with the `ikMail` name, so we can easily detect our code in the JS part of the project.

LunarX: I agree, with the "mail" term being too generic for our variable names. I don't have any particular opinion on the name of the js bridge object though

JorisBodin (approved):

PR #1305, opened 2023-08-30
fix: Notification actions doesn't work sometimes

sirambd:
**Description**
Sometimes notifications don't work because they're based on the application token, which is only available when the app is foreground, but if you're in the background you may not get the current user info.

**Fixes**
- Now we avoid using the current user's token, but just pass him his own `httpClient` so we don't have to worry about it.
- In a second step, we reset the notification in case of error, as a temporary solution (see if you want to do it differently).

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
@@ -178,11 +178,17 @@ object ApiRepository : ApiRepositoryCore() {
         return callApi(ApiRoutes.deleteMessages(mailboxUuid), POST, mapOf("uids" to messageUids))
     }
 
-    fun moveMessages(mailboxUuid: String, messagesUids: List<String>, destinationId: String): ApiResponse<MoveResult> {
+    fun moveMessages(
+        mailboxUuid: String,
+        messagesUids: List<String>,
+        destinationId: String,
+        okHttpClient: OkHttpClient = HttpClient.okHttpClient,
+    ): ApiResponse<MoveResult> {
         return callApi(
             url = ApiRoutes.moveMessages(mailboxUuid),
             method = POST,
             body = mapOf("uids" to messagesUids, "to" to destinationId),
+            okHttpClient = okHttpClient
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -178,11 +178,17 @@ object ApiRepository : ApiRepositoryCore() {
...
+        mailboxUuid: String,
+        messagesUids: List<String>,
+        destinationId: String,
+        okHttpClient: OkHttpClient = HttpClient.okHttpClient,
+    ): ApiResponse<MoveResult> {
         return callApi(
             url = ApiRoutes.moveMessages(mailboxUuid),
             method = POST,
             body = mapOf("uids" to messagesUids, "to" to destinationId),
+            okHttpClient = okHttpClient
```
KevinBoulongne, line 191:
```suggestion
            okHttpClient = okHttpClient,
```

app/src/main/java/com/infomaniak/mail/receivers/NotificationActionsReceiver.kt
```diff
@@ -144,14 +144,21 @@ class NotificationActionsReceiver : BroadcastReceiver() {
...
-            dismissNotification(context, mailbox, notificationId)
             context.trackNotificationActionEvent(matomoValue)
-            with(ApiRepository.moveMessages(mailbox.uuid, messages.getUids(), destinationId)) {
-                if (!isSuccess()) Sentry.withScope { scope ->
-                    scope.setTag("reason", "Notif action fail because of API call")
-                    scope.setExtra("destination folder role", folderRole.name)
-                    Sentry.captureException(getApiException())
+
+            with(ApiRepository.moveMessages(mailbox.uuid, messages.getUids(), destinationId, okHttpClient)) {
+
```
KevinBoulongne, line 152:
```suggestion
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1307, opened 2023-08-30
Split Message's Body in `IO` instead of `Main`

KevinBoulongne:


LunarX (approved):

LunarX (changes requested):
There's no shimmering while the mails is being processed if you've already opened the email before

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -486,6 +475,7 @@ class ThreadAdapter(
 
         override fun areContentsTheSame(oldMessage: Message, newMessage: Message): Boolean {
             return newMessage.body?.value == oldMessage.body?.value &&
+                    newMessage.body?.splitBody == oldMessage.body?.splitBody &&
```
LunarX, line 478: Do we use this equal operator without defining the equality method because in our case comparing objects by their instances is enough?

sirambd (approved):

PR #1308, opened 2023-08-30
Add a 2 sec timeout in `splitContentAndQuote()`, instead of a 1 Mo limit

KevinBoulongne:
Depends on #1307

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
@@ -54,7 +54,7 @@ object MessageBodyUtils {
         "[name=\"quote\"]", // GMX
     )
 
-    fun splitContentAndQuote(body: Body): SplitBody = runBlocking(Dispatchers.IO) {
+    fun splitContentAndQuote(body: Body): SplitBody = runBlocking {
 
         if (body.type == Utils.TEXT_PLAIN) return@runBlocking SplitBody(body.value)
 
--- a/app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
@@ -30,7 +30,7 @@ object MessageBodyUtils {
     const val INFOMANIAK_FORWARD_QUOTE_HTML_CLASS_NAME = "forwardContentMessage"
 
     private const val BLOCKQUOTE = "blockquote"
-    private const val QUOTE_DETECTION_TIMEOUT = 2_000L
+    private const val QUOTE_DETECTION_TIMEOUT = 1_500L
 
     private val quoteDescriptors = arrayOf(
         // Do not detect this quote as long as we can't detect siblings quotes or else a single reply will be missing among the
--- a/app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
@@ -85,6 +85,8 @@ object MessageBodyUtils {
                     scope.setExtra("email", AccountUtils.currentMailboxEmail.toString())
                     Sentry.captureMessage("Timeout reached while displaying a Message's body")
                 }
+            } else {
+                Sentry.captureException(it)
             }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -29,7 +30,7 @@ object MessageBodyUtils {
     const val INFOMANIAK_FORWARD_QUOTE_HTML_CLASS_NAME = "forwardContentMessage"
 
     private const val BLOCKQUOTE = "blockquote"
-    private const val QUOTE_DETECTION_SIZE_LIMIT = 1_000_000 // 1 Meg looks like a fine threshold
+    private const val QUOTE_DETECTION_TIMEOUT = 2_000L
```
KevinBoulongne, line 33: How much time do we want?

2 seconds looks ok to me.

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -51,27 +52,32 @@ object MessageBodyUtils {
         "[name=\"quote\"]", // GMX
     )
 
-    fun splitContentAndQuote(body: Body): SplitBody {
+    fun splitContentAndQuote(body: Body): SplitBody = runBlocking(Dispatchers.IO) {
```
KevinBoulongne, line 55: @sirambd Do we want to use `Dispatchers.IO` here, or some injection or something else ?

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -51,27 +52,32 @@ object MessageBodyUtils {
...
+                }
 
-        val (content, quote) = splitContentAndQuote(originalHtmlDocument, currentQuoteDescriptor, blockquoteElement)
-        return if (quote.isNullOrBlank()) SplitBody(body.value) else SplitBody(content, body.value)
+                val (content, quote) = splitContentAndQuote(originalHtmlDocument, currentQuoteDescriptor, blockquoteElement)
+                if (quote.isNullOrBlank()) SplitBody(body.value) else SplitBody(content, body.value)
+            }
+        }.getOrElse {
+            SplitBody(body.value)
+        }
```
LunarX, line 80: If we get another exception than TimeoutCancellationException we will never know it

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -51,27 +52,32 @@ object MessageBodyUtils {
...
 
-        // Heuristic to give up on mail too large for "perfect" preprocessing.
-        if (body.value.toByteArray().size > QUOTE_DETECTION_SIZE_LIMIT) return SplitBody(body.value)
+        return@runBlocking runCatching {
 
-        // The original parsed html document in full
-        val originalHtmlDocument = Jsoup.parse(body.value)
-        // Initiated to the original document and it'll be processed by Jsoup to remove quotes.
-        val htmlDocumentWithoutQuote = originalHtmlDocument.clone()
+            withTimeout(QUOTE_DETECTION_TIMEOUT) {
```
LunarX, line 61: During my testing, it waits for the block of code to finish before throwing the exception, which doesn't help make the loading faster. Is it normal?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
LunarX: Also, we should probably send a Sentry to know how many emails triggered the Timeout. For all we know, it might not happen anymore since the last update that removed the detection of outlook quotes

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
sirambd: There's no api call here, no file read/write, so there's no need for Dispatchers.IO

LunarX (approved):

sirambd (approved):

PR #1316, opened 2023-09-04
Update Matomo Threads count

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -102,7 +102,7 @@ class ThreadViewModel @Inject constructor(
             return@liveData
         }
 
-        context.trackUserInfo("nbMessagesInThread", thread.messages.count())
+        sendRelevantMatomoAndSentry(thread)
```
LunarX, line 105: Maybe rename to say it sends matomo and sentry about thread's stats or something like that

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -111,6 +112,23 @@ object SentryDebug {
...
+    //  If this doesn't trigger after a certain amount of time, you can remove it.
+    fun sendEmptyThread(thread: Thread) {
+        Sentry.withScope { scope ->
+            scope.level = SentryLevel.ERROR
+            scope.setExtra("currentUserId", "[${AccountUtils.currentUserId}]")
+            scope.setExtra("currentMailboxEmail", "[${AccountUtils.currentMailboxEmail}]")
+            scope.setExtra("folder.name", thread.folder.name)
+            scope.setExtra("folder.id", thread.folder.id)
+            scope.setExtra("thread.uid", "[${thread.uid}]")
+            scope.setExtra("thread.subject", "[${thread.subject}]")
```
LunarX, line 128: Not sure we want to log private info like these

LunarX (approved):

PR #1334, opened 2023-09-11
Fix actions from Search

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -864,6 +864,28 @@ class MainViewModel @Inject constructor(
...
 
+    fun getActionFolderRole(thread: Thread?): FolderRole? {
+        return getActionFolderRole(thread?.let(::listOf))
+    }
+
+    fun getActionFolderRole(message: Message): FolderRole? {
+        return getActionFolderRole(message.threads, message)
+    }
+
+    private fun getActionFolderRole(threads: List<Thread>?, message: Message? = null): FolderRole? {
```
LunarX, line 883: Add comment to precise that you can only pass list of threads with the same parent folder

PR #1335, opened 2023-09-11
Button to clear Contact input

FabianDevel:


KevinBoulongne (approved):

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -66,10 +66,9 @@
                 <FrameLayout
                     android:layout_width="0dp"
                     android:layout_height="wrap_content"
-                    android:layout_marginEnd="@dimen/alternativeMargin"
+                    android:layout_marginEnd="@dimen/marginStandardSmall"
```
LunarX, line 69: Won't changing this make the ripple look bad because of how close it will from the horizontal edges of the text? Do we want to modify the spacing of the icons here?

LunarX (changes requested):

app/src/main/res/layout/fragment_new_message.xml
FabianDevel: The ripple still looks decent I think, even if its a little closer to the end of the icon
Yes we want to modify the spacing to be consistent with the rest of the app (16dp margins), and iOS

LunarX (approved):

PR #1338, opened 2023-09-12
Add Matomo tracker when opening Draft

FabianDevel:


LunarX (changes requested):
The two events should be mutually exclusive. It's supposed to be triggered once each time you open one of your already existing drafts and report who last edited that draft (this very app or anyone else)

LunarX (changes requested):
Recreating the activity will trigger the matomo again

LunarX (approved):

PR #1344, opened 2023-09-12
Add missing log in case of API returning null UUID

FabianDevel:


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -312,6 +312,9 @@ class DraftsActionsWorker @AssistedInject constructor(
 
         val apiResponse = ApiController.json.decodeFromString<ApiResponse<Attachment>>(response.body?.string() ?: "")
         if (apiResponse.isSuccess() && apiResponse.data != null) {
+            apiResponse.data?.uuid?.let {
+                SentryLog.e(ATTACHMENT_TAG, "Attachment uuid null from API, this should not happen")
+            }
```
KevinBoulongne, line 317:
```suggestion
            if (apiResponse.data?.uuid == null) {
                SentryLog.e(ATTACHMENT_TAG, "Attachment uuid null from API, this should not happen")
            }
```

KevinBoulongne: Shouldn't it be that instead ?

KevinBoulongne (approved):

PR #1348, opened 2023-09-14
Hide failed Messages + Leave Thread if all Messages are hidden

KevinBoulongne:


LunarX (changes requested):
Instead of hiding messages that failed to load, display an error message in place of the body

PR #1355, opened 2023-09-14
Handle deleted Messages

KevinBoulongne:
Depends on #1354

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
@@ -87,7 +87,7 @@ class ThreadViewModel @Inject constructor(
             }
         }
 
-        messageController.getSortedMessagesAsync(threadUid)
+        messageController.getSortedAndNotDeletedMessagesAsync(threadUid)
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -87,7 +88,7 @@ class ThreadViewModel @Inject constructor(
             }
         }
 
-        messageController.getSortedMessagesAsync(threadUid)
+        messageController.getSortedAndNotDeletedMessagesAsync(threadUid)
             ?.map { results -> results.list.map { splitBody(it) } }
             ?.asLiveData()
             ?.let { emitSource(it) }
```
FabianDevel, line 94:
```suggestion
            ?.let(::emitSource)
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
FabianDevel: 😢

FabianDevel (approved):

PR #1369, opened 2023-09-18
Add loader in description dialog

FabianDevel:


LunarX (commented):
Make it so the dialog only becomes non-cancelable when the loader starts to preserve the cancelable property during the major part of the interaction with the user

LunarX (approved):

PR #1375, opened 2023-09-20
Delete popup specific behavior

FabianDevel:
Depends on #1377

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -52,6 +52,7 @@ import com.infomaniak.mail.utils.ContactUtils.getPhoneContacts
 import com.infomaniak.mail.utils.ContactUtils.mergeApiContactsIntoPhoneContacts
 import com.infomaniak.mail.utils.NotificationUtils.Companion.cancelNotification
 import com.infomaniak.mail.utils.SharedUtils.Companion.updateSignatures
+import com.infomaniak.mail.utils.Utils.isPermanentDeleteFolder
 import com.infomaniak.mail.utils.Utils.runCatchingRealm
 import dagger.hilt.android.lifecycle.HiltViewModel
 import io.realm.kotlin.ext.copyFromRealm
@@ -426,13 +427,12 @@ class MainViewModel @Inject constructor(
         message: Message? = null,
         isSwipe: Boolean = false,
     ) = viewModelScope.launch(ioCoroutineContext) {
+
         val mailbox = currentMailbox.value!!
         val threads = getActionThreads(threadsUids).ifEmpty { return@launch }
         var trashId: String? = null
         var undoResource: String? = null
-
-        val role = getActionFolderRole(threads, message)
-        val shouldPermanentlyDelete = role == FolderRole.DRAFT || role == FolderRole.SPAM || role == FolderRole.TRASH
+        val shouldPermanentlyDelete = isPermanentDeleteFolder(getActionFolderRole(threads, message))
 
         val messages = getMessagesToDelete(threads, message)
         val uids = messages.getUids()
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -92,7 +92,7 @@ class MainViewModel @Inject constructor(
     val isMovedToNewFolder = SingleLiveEvent<Boolean>()
     val toggleLightThemeForMessage = SingleLiveEvent<Message>()
     val deletedMessages = SingleLiveEvent<Set<String>>()
-    val deleteThreadsTrigger = MutableLiveData<Unit>()
+    val deleteThreadOrMessageTrigger = SingleLiveEvent<Unit>()
     val flushFolderTrigger = SingleLiveEvent<Unit>()
     val newFolderResultTrigger = MutableLiveData<Unit>()
     val reportPhishingTrigger = SingleLiveEvent<Unit>()
@@ -446,7 +446,7 @@ class MainViewModel @Inject constructor(
             }
         }
 
-        deleteThreadsTrigger.postValue(Unit)
+        deleteThreadOrMessageTrigger.postValue(Unit)
         if (apiResponse.isSuccess()) {
             refreshFoldersAsync(
                 mailbox = mailbox,
@@ -456,7 +456,7 @@ class MainViewModel @Inject constructor(
                 stopped = ::stoppedDownload,
             )
         } else if (isSwipe) {
-            // We need to make the swiped Thread come back, so we reassign the LiveData
+            // We need to make the swiped Thread come back, so we reassign the LiveData with Realm values
             reassignCurrentThreadsLive()
         }
 
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -55,6 +55,7 @@ import com.infomaniak.mail.utils.SharedUtils.Companion.updateSignatures
 import com.infomaniak.mail.utils.Utils.runCatchingRealm
 import dagger.hilt.android.lifecycle.HiltViewModel
 import io.realm.kotlin.ext.copyFromRealm
+import io.realm.kotlin.notifications.ResultsChange
 import kotlinx.coroutines.*
 import kotlinx.coroutines.flow.*
 import java.util.Date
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/infomaniak/mail/utils/AlertDialogUtils.kt
```diff
@@ -102,11 +102,11 @@ object AlertDialogUtils {
 
         MaterialAlertDialogBuilder(context)
             .setView(root)
-            .setPositiveButton(confirmButtonText, null)
+            .setPositiveButton(confirmButtonText) { _, _ -> onPositiveButtonClicked() }
```
LunarX, line 105: Won't this call `onPositiveButtonClicked()` twice?

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -369,20 +370,34 @@ fun List<Message>.getFoldersIds(exception: String? = null) = mapNotNull { if (it
...
+        callback()
+        null
     }
 }
 
+inline var Fragment.deleteThreadDialog
+    get() = (activity as? MainActivity)?.deleteThreadDialog
+    set(value) {
+        (activity as? MainActivity)?.deleteThreadDialog = value
+    }
```
LunarX, line 399: Can we share the dialog instance using the main view model instead of the activity?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -367,12 +381,23 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
...
-                deleteWithConfirmationPopup(folderRole, count = 1) { deleteThread(thread.uid) }
-                shouldKeepItem
+                deleteWithConfirmationPopup(
+                    folderRole = folderRole,
+                    count = 1,
+                    displayLoader = false,
+                    onDismiss = {
+                        // Notify only if the user cancelled the popup (e.g. the thread is not deleted),
+                        // otherwise it will notify the next item in the list and make it slightly blink
+                        if (threadListAdapter.dataSet.indexOfFirst { it is Thread && it.uid == thread.uid } == position) {
```
LunarX, line 391: Why do we need the condition? Can we not just `notifyItemChanged(position)` if `onDismiss` triggered?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -330,11 +330,20 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
...
+                    role == FolderRole.DRAFT || role == FolderRole.SPAM || role == FolderRole.TRASH
+                }
+
+                val shouldKeepItem = performSwipeActionOnThread(swipeAction, item, position, isPermanentDeleteFolder)
 
                 threadListAdapter.apply {
                     blockOtherSwipes()
-                    notifyItemChanged(position) // Animate the swiped element back to its original position
+
+                    if (swipeAction == SwipeAction.DELETE && isPermanentDeleteFolder) {
```
LunarX, line 342: Is this condition `!shouldKeepItem`?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -435,6 +450,9 @@ class MainViewModel @Inject constructor(
                 started = ::startedDownload,
                 stopped = ::stoppedDownload,
             )
+        } else if (isSwipe) {
+            // We need to make the swiped Thread come back, so we reassign the LiveData with Realm values
+            reassignCurrentThreadsLive()
```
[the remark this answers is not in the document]
FabianDevel: No we can't because we need to get the right data back from realm, and it's the only way retrigger the realm query

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -155,14 +157,25 @@ class MainViewModel @Inject constructor(
...
+        }
+    }
 
     private fun observeFolderAndFilter() = MediatorLiveData<Pair<Folder?, ThreadFilter>>().apply {
-        value = currentFolder.value to currentFilter.value!!
-        addSource(currentFolder) { value = it to value!!.second }
-        addSource(currentFilter) { value = value?.first to it }
+        postValue(currentFolder.value to currentFilter.value!!)
+        addSource(currentFolder) { postValue(it to value!!.second) }
+        addSource(currentFilter) { postValue(value?.first to it) }
```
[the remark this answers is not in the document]
FabianDevel: It will crash otherwise because you can't use set value on background thread

app/src/main/java/com/infomaniak/mail/utils/AlertDialogUtils.kt
FabianDevel: No because we override this clickListener in the  onShowListener

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -330,11 +330,20 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
                     else -> throw IllegalStateException("Only SwipeDirection.LEFT_TO_RIGHT and SwipeDirection.RIGHT_TO_LEFT can be triggered")
                 }
 
-                val shouldKeepItem = performSwipeActionOnThread(swipeAction, item)
+                val isPermanentDeleteFolder = item.folder.role.let { role ->
+                    role == FolderRole.DRAFT || role == FolderRole.SPAM || role == FolderRole.TRASH
+                }
```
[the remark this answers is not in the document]
FabianDevel: It's not really useful because it's the only place where we need this code

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
FabianDevel: No we can't, because onDismiss is called if we delete a thread, thus notifying will redraw the new position (the thread after the one deleted), then the api call will return, and we will redraw it once more, so it will blink

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
FabianDevel: I don't think we want UI stuff like this in a viewModel

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
FabianDevel: No it's not the same

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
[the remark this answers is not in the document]
LunarX: Doesn't the data stay the same? Why could the data have changed?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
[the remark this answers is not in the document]
LunarX: We also check for these folder roles at least in the MainViewModel inside `deleteThreadsOrMessage()` and in Extensions inside `deleteWithConfirmationPopup()`

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
LunarX: Ah yeah dismiss also triggers when positive button is clicked, it's not the first time it surprises me. Then, can we replace `indexOfFirst` and the complicated predicate with `indexOf`?

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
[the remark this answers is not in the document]
FabianDevel: The data in realm is the same, but the data of the adapter is false, and our logic is to react to realm for the data of the adapter, so we need to make the realm live data trigger so the adapter get the update with the good data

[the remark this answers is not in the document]
LunarX: Why could the adapter data have changed? You just swiped the item to delete it, right? If so nothing should have changed and if something did indeed change in realm (because of like a notification we juste received), then realm would've already notified the item by itself without reassigning anything

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
LunarX: @sirambd do you have any opinion on the matter?

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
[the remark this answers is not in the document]
FabianDevel: Because we manually deleted an item from the adapter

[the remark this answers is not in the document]
sirambd: They've already talked to me about this, and it seems to me that it's because you only remove the adapter element, so it's no longer synchronized with realm.
So when the action couldn't be performed, you'd have to synchronize with realm again. 
In other words, listen to realm again.

But if there's another way of doing it, I'd have to compare the impact with how it was done on **iOs**.
For now, we can keep it that way.

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
sirambd: a priori the logic will change in another PR so we can mergeà priori la logique va changer dans une autre PR donc on peut merger

LunarX (approved):

PR #1380, opened 2023-09-21
Cancel AI request when leaving the fragment

LunarX:
Depends on #1379


app/src/main/java/com/infomaniak/mail/ui/main/newMessage/AiPropositionFragment.kt
```diff
@@ -49,6 +49,8 @@ class AiPropositionFragment : Fragment() {
     private val newMessageViewModel: NewMessageViewModel by activityViewModels()
     private val aiViewModel: AiViewModel by activityViewModels()
 
+    var requestJob: Job? = null
```
KevinBoulongne, line 52:
```suggestion
    private var requestJob: Job? = null
```

KevinBoulongne (changes requested):

PR #1385, opened 2023-09-22
Scroll to correct Message when all expanded Messages are done loading

KevinBoulongne:
Depends on #1408


LunarX (changes requested):
Doesn't seem to work for threads that were never opened and does not scroll consistently on already opened threads

sirambd (approved):

PR #1397, opened 2023-09-27
Clean code before fixing `MESSAGE_NOT_FOUND`

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -86,7 +86,7 @@ class FetchMessagesManager @Inject constructor(
                 )
                 return
             }
-            return@let threads
+            return@let threads.toList()
```
FabianDevel, line 89:
```suggestion

            return@let threads.toList()
```

FabianDevel (changes requested):

PR #1413, opened 2023-10-03
Ai contextual menu actions

LunarX:

PR #1415, opened 2023-10-03
Ai discover bottom sheet

LunarX:
Depends on #1413


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/ui/bottomSheetDialogs/AiDiscoveryBottomSheetDialog.kt
```diff
@@ -0,0 +1,50 @@
...
+            setText(R.string.buttonTry)
+            setOnClickListener {
+                aiViewModel.aiPromptOpeningStatus.value = AiPromptOpeningStatus(true)
+                dismiss()
+            }
+        }
+
+        secondaryActionButton.setOnClickListener {
+            dismiss()
+        }
```
FabianDevel, line 48:
```suggestion
        secondaryActionButton.setOnClickListener { dismiss() }
```

app/src/main/res/values-de/strings.xml
```diff
@@ -52,6 +52,8 @@
     <string name="aiButtonInsert">Einfügen</string>
     <string name="aiButtonRefine">Verfeinern</string>
     <string name="aiButtonRetry">Wiederholen</string>
+    <string name="aiDiscoveryDescription">Vereinfachen Sie das Schreiben von Nachrichten mit künstlicher Intelligenz.</string>
+    <string name="aiDiscoveryTitle">Schreiben Sie Ihre E-Mails mit AI</string>
```
FabianDevel, line 56: A bit strange to use AI then künstlicher Intelligenz
If we want to use AI, we should probably use artifizielle Intelligenz
Otherwise, maybe use KI

FabianDevel (changes requested):

app/src/main/res/values-de/strings.xml
JorisBodin: At infomaniak we use "Künstliche Intelligenz", so yes "KI".

PR #1447, opened 2023-10-06
Add SubBody in Realm

KevinBoulongne:
Depends on #1446


sirambd (changes requested):
I approve, but we're waiting for Monday's meeting to see what will be voted on.

PR #1462, opened 2023-10-12
Clean XML code & imports

KevinBoulongne:


app/src/main/res/navigation/main_navigation.xml
```diff
@@ -521,7 +521,7 @@
     <activity
         android:id="@+id/loginActivity"
         android:name="com.infomaniak.mail.ui.login.LoginActivity"
-        android:label="activity_login"
+        android:label="LoginActivity"
```
LunarX, line 524: All these names might be used by matomo to record which pages have been navigated to

LunarX (changes requested):

PR #1477, opened 2023-10-13
Split `fetchCredentials()` into more meaningful functions

KevinBoulongne:
Depends on #1464

PR #1479, opened 2023-10-13
Fix already-sync snackbar that wasn't appearing

KevinBoulongne:
Depends on #1477

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
@@ -154,7 +154,7 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
 
         syncAutoConfig.setOnClickListener {
             trackSyncAutoConfigEvent("openFromMenuDrawer")
-            requireActivity().launchSyncAutoConfigActivityForResult()
+            launchSyncAutoConfigActivityForResult()
             closeDrawer()
         }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
```diff
@@ -153,7 +153,7 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
         }
 
         syncAutoConfig.setOnClickListener {
-            startActivity(Intent(context, SyncAutoConfigActivity::class.java))
+            (activity as MainActivity).navigateToSyncAutoConfigActivity()
```
FabianDevel, line 156: What if we create a 'launchSyncAutoConfigActivity' fonction in Extension.kt along with all the other launchActivity functions, so you won't have to cast both times

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
```diff
@@ -155,7 +154,7 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
 
         syncAutoConfig.setOnClickListener {
             trackSyncAutoConfigEvent("openFromMenuDrawer")
-            startActivity(Intent(context, SyncAutoConfigActivity::class.java))
+            requireActivity().launchSyncAutoConfigActivityForResult()
```
FabianDevel, line 157: it seems you only use this function in fragments, why not putting the requireActivity() diurectly in the extension, changing it to Fragment extension

FabianDevel (changes requested):

FabianDevel (approved):

PR #1490, opened 2023-10-16
Fix NPE because of Thread too long

FabianDevel:


LunarX (changes requested):
These 3 `post()` are just here to gain us some time before accessing values. They should not be there. Imo, if we simply want to gain time, we need to create a coroutine add a delay and emnsureActive. But yeah these 3 parts of the code should not be relying on a random delay to work

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -522,7 +524,16 @@ class NewMessageFragment : Fragment() {
...
         }
 
-        fromMailAddress.post { addressListPopupWindow.width = fromMailAddress.width }
+        fromMailAddress.post {
+            runCatching {
+                addressListPopupWindow.width = fromMailAddress.width
+            }.onFailure { exception ->
+                Sentry.withScope {
+                    Sentry.captureException(exception)
+                    Sentry.captureMessage("Binding null in post(), this is not normal", SentryLevel.WARNING)
```
KevinBoulongne, line 533: This will send 2 sentries at once.
We need to choose which one we want to send.

KevinBoulongne (approved):

PR #1502, opened 2023-10-18
Add TitleAlertDialog

KevinBoulongne:

PR #1504, opened 2023-10-18
Show AppReview dialog

KevinBoulongne:
Depends on #1502


app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -443,6 +450,19 @@ class MainActivity : BaseActivity() {
...
 
+    private fun handleInAppReview() = with(localSettings) {
+        if (showAppReviewDialog && appLaunches != 0 && appLaunches % 50 == 0) {
+            titleDialog.show(
+                title = R.string.reviewAlertTitle,
+                onPositiveButtonClicked = {
+                    showAppReviewDialog = false
+                    launchInAppReview()
+                },
+                onNegativeButtonClicked = { openUrl(getString(R.string.urlUserReportAndroid)) },
```
LunarX, line 461: Do we instantly stop the users in their flow, without asking, by sending them out of the app and in their browser if they clic on 'no'? They just started the app with probably a thing in mind they had to do, maybe a good amount of them simply don't want to interact with us and just want to quickly go check what they came here for

KevinBoulongne: Seen on kChat.

FabianDevel, line 461: This URL should be in build.gradle, not a string in Loco

app/src/main/res/values-de/strings.xml
```diff
@@ -344,6 +344,7 @@
     <string name="restoreEmailsNoBackup">Kein Backup gefunden</string>
     <string name="restoreEmailsText">Alle E-Mails werden wieder in Ihrem Posteingang landen.</string>
     <string name="restoreEmailsTitle">Wiederherstellung Ihrer E-Mails</string>
+    <string name="reviewAlertTitle">Gefällt Ihnen Infomaniak Mail?</string>
```
FabianDevel, line 347:
```suggestion
    <string name="reviewAlertTitle">Möchten Sie Infomaniak Mail?</string>
```

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -443,6 +450,19 @@ class MainActivity : BaseActivity() {
         }
     }
 
+    private fun handleInAppReview() = with(localSettings) {
+        if (showAppReviewDialog && appLaunches != 0 && appLaunches % 50 == 0) {
```
FabianDevel, line 454: Is it wanted to be that hard to show this view ? 
Because if the user doesn't exactly launch the app on the % 50th, and just come back from background, it won't be shown
So virtually this could never be presented

FabianDevel (changes requested):

FabianDevel (approved):

PR #1509, opened 2023-10-20
Don't blink when updating Message status (favorite, seen, etc…)

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -533,9 +533,7 @@ class ThreadAdapter(
 
         override fun areContentsTheSame(oldMessage: Message, newMessage: Message): Boolean {
             return newMessage.body?.value == oldMessage.body?.value &&
-                    newMessage.splitBody == oldMessage.splitBody &&
-                    newMessage.isSeen == oldMessage.isSeen &&
-                    newMessage.isFavorite == oldMessage.isFavorite
```
LunarX, line 538: If two version of a message have the same `isSeen` value and the same `isFavorite` value they should be considered the same, no?

LunarX (changes requested):
Also you have a CI issue

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
KevinBoulongne: Yes, that's why we remove this check.

LunarX (approved):

PR #1534, opened 2023-11-03
Stop detecting simple blockquotes to collapse replies

LunarX:
We now also detect outlook replies again thanks to the modifications

Depends on #1540


app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -78,73 +75,22 @@ object MessageBodyUtils {
...
-    ): Pair<String, String?> {
-        return when {
-            currentQuoteDescriptor == BLOCKQUOTE -> {
-                for (quotedContentElement in htmlDocumentWithQuote.select(currentQuoteDescriptor)) {
-                    if (quotedContentElement.toString() == blockquoteElement.toString()) {
-                        quotedContentElement.remove()
-                        break
-                    }
+            val foundQuotes = htmlDocument.select(quoteDescriptor)
+            if (foundQuotes.isNotEmpty()) {
```
FabianDevel, line 85: This check is useless with a forEach

FabianDevel (changes requested):

FabianDevel (approved):

PR #1538, opened 2023-11-07
Don't delete Attachment upload local file if another Attachment is the same

FabianDevel:


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -267,6 +267,7 @@ class DraftsActionsWorker @AssistedInject constructor(
 
     private fun Draft.uploadAttachments(): Result {
         val attachmentsToUpload = getNotUploadedAttachments(draft = this)
+        val attachmentsNames = attachmentsToUpload.map { it.name }.toMutableList()
```
KevinBoulongne, line 270: Don't loop here, do it directly in `getNotUploadedAttachments()` to save a loop.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
FabianDevel: Done in the next PR

KevinBoulongne (approved):

PR #1542, opened 2023-11-08
Add Confetti library

KevinBoulongne:


FabianDevel (changes requested):
Wouldn't it be better as a module ?

PR #1549, opened 2023-11-10
Improve email rendering by breaking wide lines

LunarX:
A not so uncommonly occurring issue with our mail rendering is that some really wide continuous lines of text (mostly URLs) are not softwrapped correctly. This results in excessively zoomed out emails which become hard too read or even completely unreadable for the most extreme cases. 

This PR solves this issue by adding "zero width spaces" inside long strings which will enable softwrapping at these key positions. This strategy is inspired by what Google seems to be doing in its webmail and it seems very likely to also be the comportment they have on their mobile app.


valentinperignon (approved):
Check the logic of the code, still needs a review from @Infomaniak/android

PR #1553, opened 2023-11-15
Use Android formatter instead of the Core one for Quotas display

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/models/Quotas.kt
```diff
@@ -31,18 +31,21 @@ import kotlin.math.ceil
 class Quotas : EmbeddedRealmObject {
 
     @SerialName("size")
-    private var _size: Int = 0
+    private var _size: Long = 0L
 
     private val size: Long
-        get() = _size.toLong() * FormatterFileSize.KIBI_BYTE.toLong()
+        get() {
+            val converted = _size * 1_000L / 1_024L // Convert size from base 1024 to base 1000
```
FabianDevel, line 38: Divide by 1024 first, even if it is now a long, it's still theoretically less plausible to overflow

FabianDevel (changes requested):

FabianDevel (approved):

PR #1560, opened 2023-11-21
Proposition: Shorten new message fragment

LunarX:
Because NewMessageFragment was too long we decided to try and shorten it by placing some of its code in other files. Here is my proposition. What's your opinion on it?

I identified four main group of functions/functionalities that I could extract to other files that I called "managers". The four themes are: ai, external recipients, recipient fields and editor buttons


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageManager.kt
```diff
@@ -0,0 +1,55 @@
...
+    protected val newMessageViewModel: NewMessageViewModel get() = _newMessageViewModel!!
+    private var _binding: FragmentNewMessageBinding? = null
+    protected val binding: FragmentNewMessageBinding get() = _binding!!
+    private var _fragment: NewMessageFragment? = null
+    protected val fragment: NewMessageFragment get() = _fragment!!
+
+    val viewLifecycleOwner get() = fragment.viewLifecycleOwner
+    val childFragmentManager get() = fragment.childFragmentManager
+    val resources get() = fragment.resources
+    val context get() = fragment.requireContext()
```
tevincent, line 36: A specific reason to use a `get()` here ? Since you only need to have the value once and you don't need to recompute the value every time, a `by lazy` might be better suited here:
`val context get() by lazy { fragment.requireContext() }`

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
```diff
@@ -0,0 +1,246 @@
...
+
+    private var _aiViewModel: AiViewModel? = null
+    private val aiViewModel: AiViewModel get() = _aiViewModel!!
+
+    private val animationDuration by lazy { resources.getInteger(R.integer.aiPromptAnimationDuration).toLong() }
+    private val scrimOpacity by lazy { ResourcesCompat.getFloat(context.resources, R.dimen.scrimOpacity) }
+    private val backgroundColor by lazy { context.getColor(R.color.backgroundColor) }
+    private val black by lazy { context.getColor(RCore.color.black) }
+
+    private var aiPromptFragment: AiPromptFragment? = null
```
tevincent, line 69: You can also do this here:
```
    private val aiPromptFragment by lazy { 
        childFragmentManager.findFragmentByTag(AI_PROMPT_FRAGMENT_TAG) as? AiPromptFragment ?: AiPromptFragment()
    }
```

Then you don't need to check for nullability and use the !! operator.

tevincent, line 69: Could be something like this to avoid the !! operator:

```
private val aiPromptFragment by lazy {
        childFragmentManager.findFragmentByTag(AI_PROMPT_FRAGMENT_TAG) as? AiPromptFragment ?: AiPromptFragment()
    }
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
```diff
@@ -0,0 +1,246 @@
...
+    private fun onAiPromptClosed(withoutTransition: Boolean) = with(binding) {
+
+        fun removeFragmentAndHideScrim() {
+            aiPromptFragment?.let {
+                childFragmentManager
+                    .beginTransaction()
+                    .remove(it)
+                    .commitNow()
+            }
+            aiPromptFragment = null
```
tevincent, line 151: Any reason to force the `aiPromptFragment` to be `null` ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
```diff
@@ -0,0 +1,246 @@
...
+            viewLifecycleOwner.lifecycleScope.launch {
+                delay(animationDuration)
+                removeFragmentAndHideScrim()
+            }
+        }
+
+        if (!withoutTransition) animateAiPrompt(false)
+        setAiPromptVisibility(false)
+        newMessageConstraintLayout.descendantFocusability = FOCUS_BEFORE_DESCENDANTS
+    }
```
tevincent, line 170: We can avoid that inner function by doing this maybe:
``` kotlin
private fun onAiPromptClosed(withoutTransition: Boolean) = with(binding) {
        aiPromptFragmentContainer.hideKeyboard()
            if (withoutTransition) delay(animationDuration)
            childFragmentManager.beginTransaction().remove(aiPromptFragment).commitNow()
            scrim.isGone = true
        }

        if (!withoutTransition) animateAiPrompt(false)
        setAiPromptVisibility(false)
        newMessageConstraintLayout.descendantFocusability = FOCUS_BEFORE_DESCENDANTS
    }
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
```diff
@@ -0,0 +1,246 @@
...
+                scrim.alpha = alpha
+                activity.window.statusBarColor = UiUtils.pointBetweenColors(backgroundColor, black, alpha)
+            }
+            start()
+        }
+    }
+
+    private fun setAiPromptVisibility(isVisible: Boolean) {
+
+        fun updateNavigationBarColor() {
```
tevincent, line 195: Any reasons to add that `inner function` here ? Since you're only using this once.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -512,25 +380,6 @@ class NewMessageFragment : Fragment() {
...
-                newMessageViewModel.externalRecipientCount.value = null to 2
-                return
-            }
-
-            if (quantityForThisField == 1) externalRecipientEmail = singleEmail
-        }
-
-        newMessageViewModel.externalRecipientCount.value = externalRecipientEmail to externalRecipientQuantity
-    }
-
```
tevincent, line 533: Line 400, you could do something like this:

```kotlin
val selectedSignature = with(signatures) {
            first {it.id == newMessageViewModel.selectedSignatureId || it.isDefault }
        }
```

to avoid the `!!` operator and to also avoid going twice through the list.

tevincent (changes requested):
Let me know what you think about my propositions.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageManager.kt
LunarX: Yeah my problem with that part is that `fragment` can be reassigned through `initValues()` and so I can't rely on a `by lazy` because it wouldn't see the new values of `fragment`

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
LunarX: `aiPromptFragment` needs to be set to `null` inside `removeFragmentAndHideScrim()` anyway so I'm afraid I needs to stay nullable.

Also, the lazy value wouldn't work because when calling `childFragmentManager.findFragmentByTag()` the value can change between two calls and caching its with a `by lazy` value would prevent it from ever updating

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
LunarX: Looking at the commits, it looks like @FabianDevel assigned it to `null` here to free its reference and avoid memory leaks. When navigating down the navigation tree I guess.

I do wonder if we could have a single instance of `AiPromptFragment()` though. But I think that's what I did go for when I implemented it back then and it actually caused an issue that said that the Fragment was already added when activity would get recreated. Maybe that part could use some comments?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
LunarX: Could there be an advantage in avoiding to launch a coroutine if we don't need it? Because else, that's a nice idea. 

If we do get rid of the double function call, I still like the idea of have the lines wrapped inside a more meaningful function like `removeFragmentAndHideScrim()` though, what do you think?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
LunarX: Sometimes we took the habit of creating some inner functions (in opposition to class methods) here and there just so they stay closer to where they are used. No better reason.

Making these two lines into a function is to group things together more neatly though I'd say

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
LunarX: Unfortunately, we can't simplify it this way because we first need to make sure there's no selected signature before falling back on the default one. 

This could nonetheless be simplified into a single loop but here the order of magnitude of the size of the list is of about ~10 values which is probably why we've let it go through. Do we want to optimize it further? Maybe? I don't know if it's worth the effort and the lisibility

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
FabianDevel: Yes it was because of leaks, as AiPropositionFragment also use this viewModel, but the AiPromptFragment is destroyed when we navigate to the proposition

tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageManager.kt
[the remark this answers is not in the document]
FabianDevel: In fact I just realize that these manager should probably be FragmentScoped than ActivityScoped, as they were a part of NewMessageFragment.
It would greatly simplify all the fragment management

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageRecipientFieldsManager.kt
```diff
@@ -0,0 +1,184 @@
...
+import com.infomaniak.mail.utils.MergedContactDictionary
+import com.infomaniak.mail.utils.copyRecipientEmailToClipboard
+import dagger.hilt.android.scopes.ActivityScoped
+import javax.inject.Inject
+
+@ActivityScoped
+class NewMessageRecipientFieldsManager @Inject constructor() : NewMessageManager() {
+
+    private var _externalsManager: NewMessageExternalsManager? = null
+    private val externalsManager: NewMessageExternalsManager get() = _externalsManager!!
```
FabianDevel, line 41:
```suggestion
    private inline val externalsManager: NewMessageExternalsManager get() = _externalsManager!!
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageRecipientFieldsManager.kt
```diff
@@ -0,0 +1,184 @@
...
+        externalsManager: NewMessageExternalsManager,
+    ) {
+        super.initValues(newMessageViewModel, binding, fragment)
+        _externalsManager = externalsManager
+
+        viewLifecycleOwner.lifecycle.addObserver(LifecycleEventObserver { _: LifecycleOwner, event: Lifecycle.Event ->
+            if (event == Lifecycle.Event.ON_DESTROY) {
+                _externalsManager = null
+            }
+        })
```
FabianDevel, line 58: instead of setting observer in each Manager, there should be a way to factorize this in the parent, with an abstract "onViewDestroy" function to call here

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageExternalsManager.kt
```diff
@@ -0,0 +1,136 @@
...
+import com.infomaniak.mail.utils.MergedContactDictionary
+import com.infomaniak.mail.utils.Utils
+import dagger.hilt.android.scopes.ActivityScoped
+import javax.inject.Inject
+
+@ActivityScoped
+class NewMessageExternalsManager @Inject constructor() : NewMessageManager() {
+
+    private var _informationDialog: InformationAlertDialog? = null
+    private val informationDialog: InformationAlertDialog get() = _informationDialog!!
```
FabianDevel, line 38:
```suggestion
    private inline val informationDialog: InformationAlertDialog get() = _informationDialog!!
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageEditorManager.kt
```diff
@@ -0,0 +1,143 @@
...
+class NewMessageEditorManager @Inject constructor(
+    @ActivityContext private val activityContext: Context,
+) : NewMessageManager() {
+
+    private val activity get() = activityContext as Activity
+
+    private var _aiManager: NewMessageAiManager? = null
+    private val aiManager: NewMessageAiManager get() = _aiManager!!
+    private var _filePicker: FilePicker? = null
+    private val filePicker: FilePicker get() = _filePicker!!
```
FabianDevel, line 52:
```suggestion
    private inline val filePicker: FilePicker get() = _filePicker!!
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageEditorManager.kt
```diff
@@ -0,0 +1,143 @@
...
+
+@ActivityScoped
+class NewMessageEditorManager @Inject constructor(
+    @ActivityContext private val activityContext: Context,
+) : NewMessageManager() {
+
+    private val activity get() = activityContext as Activity
+
+    private var _aiManager: NewMessageAiManager? = null
+    private val aiManager: NewMessageAiManager get() = _aiManager!!
```
FabianDevel, line 50:
```suggestion
    private inline val aiManager: NewMessageAiManager get() = _aiManager!!
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageEditorManager.kt
```diff
@@ -0,0 +1,143 @@
...
+import dagger.hilt.android.scopes.ActivityScoped
+import javax.inject.Inject
+import com.google.android.material.R as RMaterial
+
+@ActivityScoped
+class NewMessageEditorManager @Inject constructor(
+    @ActivityContext private val activityContext: Context,
+) : NewMessageManager() {
+
+    private val activity get() = activityContext as Activity
```
FabianDevel, line 47:
```suggestion
    private inline val activity get() = activityContext as Activity
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageRecipientFieldsManager.kt
FabianDevel, line 41: shouldn't this be obtain by injection ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageEditorManager.kt
FabianDevel, line 50: Shouldn't this be obtained by injection ?

FabianDevel (changes requested):
Architecture changes suggestion : Make the parent class fragmentScoped

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageRecipientFieldsManager.kt
LunarX: If we do end up using it in more places than this single one, we could go this route but for now I prefer this way as it better shows that externalManager needs to init its own values before being accessed by editorManager. Same for the other manager

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageManager.kt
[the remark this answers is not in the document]
LunarX: In the end, passing the fragment would be more bothersome when we simply already have access to the fragment here, so we chose to keep this solution. The abstract class won't be instantiated so there's no need to make it injectable. Finally, I changed to the managers from ActivityScoped to FragmentScoped because here it doesn't change anything and it's more intuitive for these objects to be linked to the Fragment as they are some kind of extension of the fragment.

FabianDevel (approved):

JorisBodin (approved):

PR #1565, opened 2023-11-27
Support long press on emails and phone numbers inside emails

LunarX:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -316,7 +322,23 @@ class ThreadFragment : Fragment() {
...
             },
-            promptLink = linkContextualMenuAlertDialog::show
+            promptLink = { data, type ->
+                if (type == ThreadAdapter.ContextMenuType.PHONE) {
+                    // When adding a phone number to contacts, Google decodes this value in case it's url-encoded. But I could not
+                    // reproduce this issue when manually creating a url-encoded href. If this is triggered, fix it by also
+                    // decoding it at that step.
+                    if (data.contains('%')) Sentry.withScope { scope ->
+                        scope.level = SentryLevel.ERROR
+                        Sentry.captureMessage("Google was right, phone numbers can appear url-encoded. Needs to be fixed")
```
FabianDevel, line 332:
```suggestion
                        Sentry.captureMessage("Google was right, phone numbers can be url-encoded. Needs to be fixed")
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -425,8 +425,12 @@ class ThreadAdapter(
...
 
     private fun ThreadViewHolder.bindBody(message: Message, hasQuote: Boolean) = with(binding) {
-        bodyWebView.setupLinkContextualMenu { url -> threadAdapterCallbacks?.promptLink?.invoke(url) }
-        fullMessageWebView.setupLinkContextualMenu { url -> threadAdapterCallbacks?.promptLink?.invoke(url) }
+        bodyWebView.setupLinkContextualMenu { data, type ->
+            threadAdapterCallbacks?.promptLink?.invoke(data, type)
+        }
+        fullMessageWebView.setupLinkContextualMenu { data, type ->
+            threadAdapterCallbacks?.promptLink?.invoke(data, type)
+        }
```
FabianDevel, line 433: I'd rather check if the callback is null, as it prevents us from setting an empty clickListener

```suggestion
        threadAdapterCallbacks?.promptLink?.let { 
            bodyWebView.setupLinkContextualMenu(it)
            fullMessageWebView.setupLinkContextualMenu(it)
        }
```

app/src/main/java/com/infomaniak/mail/ui/alertDialogs/ContextualMenuAlertDialog.kt
```diff
@@ -0,0 +1,62 @@
...
+) : BaseAlertDialog(activityContext) {
+
+    protected abstract val items: List<Pair<Int, (String, SnackBarManager) -> Unit>>
+
+    val binding: DialogLinkContextualMenuBinding by lazy { DialogLinkContextualMenuBinding.inflate(activity.layoutInflater) }
+    override val alertDialog: AlertDialog by lazy { initDialog() }
+
+    private lateinit var snackBarManager: SnackBarManager
+
+    private var lastData = ""
```
FabianDevel, line 39: I don't get why you name it that way

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/alertDialogs/ContextualMenuAlertDialog.kt
```diff
@@ -0,0 +1,62 @@
...
+abstract class ContextualMenuAlertDialog(
+    @ActivityContext private val activityContext: Context,
+) : BaseAlertDialog(activityContext) {
+
+    protected abstract val items: List<Pair<Int, (String, SnackBarManager) -> Unit>>
+
+    val binding: DialogLinkContextualMenuBinding by lazy { DialogLinkContextualMenuBinding.inflate(activity.layoutInflater) }
+    override val alertDialog: AlertDialog by lazy { initDialog() }
+
+    private lateinit var snackBarManager: SnackBarManager
```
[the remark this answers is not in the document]
LunarX: Honestly, I started wondering the same

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
LunarX: Here `promptLink` is not nullable so the check would be on the `threadAdapterCallbacks ` wrapper, but yeah I still understand what you mean. 

But the `setupLinkContextualMenu` will only ever be called once at the very start when binding the view. Could it be possible that the wrapper would be null by this point and only be set a bit later? Therefore we would never have assigned the clic listener. 

Also, wasn't one of the point of having this nullable wrapper so that we can set it to null at some point and if click listeners still trigger afterwards, they will stop accessing this data that got freed?

Overall, even if easier to read, I feel like this would lead to more bugs dans the current way it's written

FabianDevel: As the callbacks wrapper is created in the `init`of the adapter, he will be set in onBindViewHolder so the listener will always be assigned
It is true though that we want to avoid leak by setting it to null when the whole ThreadFragment's view is destroyed, but it seems the callback of promptLink has nothing that can leak here as the dialog are activityScoped

So I'd say theoretically I do not see any problems, but I guess we can keep it this way if we really want to be safer. I'm not sure enough to confidantly say this is 100% safe

FabianDevel (approved):

PR #1567, opened 2023-11-29
Add LargeIconSize to dimans.xml

FabianDevel:


tevincent (approved):

LunarX (changes requested):
I don't think grouping the size of this avatar-icon in the illustration and the size of these checkboxes under the same dimension makes much sens. 

I don't know if we want to update the illustration automatically if we would happen to change a dimension in the code as it was not really designed to be much more than a stale illustration.

LunarX (approved):

PR #1568, opened 2023-11-29
Create custom view for attachment

FabianDevel:
Depends on #1567

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/views/AttachmentDetailsView.kt
+++ b/app/src/main/java/com/infomaniak/mail/views/AttachmentDetailsView.kt
@@ -21,9 +21,9 @@ import android.content.Context
 import android.text.format.Formatter
 import android.util.AttributeSet
 import android.view.LayoutInflater
-import android.widget.FrameLayout
 import androidx.annotation.DimenRes
 import androidx.annotation.StyleRes
+import androidx.constraintlayout.widget.ConstraintLayout
 import coil.load
 import com.infomaniak.lib.core.utils.getAttributes
 import com.infomaniak.lib.core.utils.setMarginsRelative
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout/view_attachment_details.xml
```diff
@@ -0,0 +1,68 @@
...
+            android:maxWidth="220dp"
+            tools:text="billet-055.pdf" />
+
+        <TextView
+            android:id="@+id/fileSize"
+            style="@style/Label.Secondary"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            tools:text="36 ko" />
+    </LinearLayout>
```
tevincent, line 66: It's not possible to get rid of the `LinearLayout` ? Usually, we don't need another level of intrication when using a `ConstraintLayout`.

FabianDevel: It is possible to get rid of it, but I kept it because it allows to "factorize" the margins and the dimensions of the two separate textViews
With the constraint layout I thought that it would be easy to forget to update a margin or padding in one or the other Views.
It also make the entire file a lot easier to read
I think the performance tradeoff here should be really low as the view is very simple

What do you think ?

FabianDevel: I just realize that this is not the latest version of this file
this was a work-in-progress that was an extraction in another file of the existing view, without refactoring

app/src/main/res/layout/view_attachment_details.xml
```diff
@@ -0,0 +1,70 @@
...
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    tools:parentTag="android.widget.LinearLayout"
```
tevincent, line 21: Your parent tag should be a `ConstraintLayout` since you're custom view extend from `ConstraintLayout`.

tevincent (changes requested):

tevincent (approved):

app/src/main/java/com/infomaniak/mail/views/AttachmentDetailsView.kt
```diff
@@ -0,0 +1,95 @@
...
+            override fun getStyleResources() = DisplayStyleResources(
+                iconSize = R.dimen.largeIconSize,
+                marginSize = RCore.dimen.marginStandard,
+                fileNameStyle = R.style.Body,
+                fileSizeStyle = R.style.Body_Secondary,
+            )
+        };
+
+        abstract fun getStyleResources(): DisplayStyleResources
+    }
```
LunarX, line 87: Can't we go we specify the parameters inside the Enum rather than using a data class?

LunarX (changes requested):

app/src/main/res/layout/item_attachment.xml
```diff
@@ -90,9 +50,7 @@
             app:layout_constraintBottom_toBottomOf="parent"
             app:layout_constraintDimensionRatio="1"
             app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintStart_toEndOf="@id/textsLayout"
-            app:layout_constraintTop_toTopOf="parent"
-            tools:visibility="visible" />
```
LunarX, line 95: I don't think you want to remove this constraint "top to top", no? 

Also, can we keep the tools:visibility if you have no preferences on the subject? I like it better if we can see it here

app/src/main/res/values/attrs.xml
```diff
@@ -141,4 +141,11 @@
         <attr name="showCloseIcon" format="boolean" />
     </declare-styleable>
 
+    <declare-styleable name="AttachmentDetailsView">
+        <attr name="displayStyle" format="enum">
+            <enum name="chip" value="0" />
+            <enum name="bottomSheet" value="1" />
```
LunarX, line 147: Maybe "small vs large" instead of "chip vs bottomSheet"

LunarX (changes requested):

app/src/main/res/values/attrs.xml
FabianDevel: After team agreement : we keep chip vs bottomsheet

LunarX (approved):

PR #1572, opened 2023-11-30
T04: Clean Thread code before adding Tablet mode

KevinBoulongne:
Depends on #1575

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
@@ -122,21 +122,16 @@ class ThreadFragment : Fragment() {
         super.onViewCreated(view, savedInstanceState)
 
         initAdapter()
-
+        setupDialogs()
         permissionUtils.registerDownloadManagerPermission()
-        bindAlertToViewLifecycle(descriptionDialog)
 
-        linkContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
-        emailContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
-        phoneContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
-
-        mainViewModel.toggleLightThemeForMessage.observe(viewLifecycleOwner, threadAdapter::toggleLightMode)
+        observeLightThemeToggle()
         observeThreadLive()
 
         threadViewModel.openThread().observe(viewLifecycleOwner) { result ->
 
             if (result == null) {
-                leaveThread()
+                closeThread()
                 return@observe
             }
 
@@ -165,7 +160,9 @@ class ThreadFragment : Fragment() {
 
     private fun setupUi(folderRole: FolderRole?) = with(binding) {
 
-        toolbar.setNavigationOnClickListener { leaveThread() }
+        toolbar.setNavigationOnClickListener {
+            closeThread()
+        }
 
         val defaultTextColor = context.getColor(R.color.primaryTextColor)
         appBar.addOnOffsetChangedListener { appBarLayout, verticalOffset ->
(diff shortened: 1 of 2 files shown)


FabianDevel (changes requested):

LunarX (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -293,58 +252,61 @@ class ThreadFragment : Fragment() {
...
-            isThemeTheSameMap = result.isThemeTheSameMap
-
-            stateRestorationPolicy = StateRestorationPolicy.PREVENT_WHEN_EMPTY
-            contacts = mainViewModel.mergedContactsLive.value ?: emptyMap()
-        }
+    private fun setupDialogs() {
+        bindAlertToViewLifecycle(descriptionDialog)
+        linkContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
+        emailContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
+        phoneContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
```
tevincent, line 265:
```suggestion
        mainViewModel.snackBarManager.apply {
              linkContextualMenuAlertDialog.initValues(this)
              emailContextualMenuAlertDialog.initValues(this)
              phoneContextualMenuAlertDialog.initValues(this)
        }
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -293,58 +252,61 @@ class ThreadFragment : Fragment() {
...
+            threadSubject.movementMethod = LinkMovementMethod.getInstance()
+
+            iconFavorite.apply {
+                setIconResource(if (thread.isFavorite) R.drawable.ic_star_filled else R.drawable.ic_star)
+                val color = if (thread.isFavorite) {
+                    context.getColor(R.color.favoriteYellow)
+                } else {
+                    context.getAttributeColor(RMaterial.attr.colorPrimary)
+                }
+                iconTint = ColorStateList.valueOf(color)
```
tevincent, line 290: This logic shouldn't be in the fragment but in the `ViewModel` instead. Instead of returning a simple thread, you can also return the icon and the color and tint to use.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -381,44 +343,36 @@ class ThreadFragment : Fragment() {
...
-        }
-        toolbarSubject.setOnLongClickListener {
-            context.copyStringToClipboard(subject, R.string.snackbarSubjectCopiedToClipboard, mainViewModel.snackBarManager)
-            true
+            val (subject, spannedSubject) = computeSubject(
+                thread = result.thread ?: return@observe,
+                emailDictionary = result.mergedContacts ?: emptyMap(),
+                aliases = result.mailbox?.aliases ?: emptyList(),
+                externalMailFlagEnabled = result.mailbox?.externalMailFlagEnabled ?: false,
+            )
```
tevincent, line 354: This also need to be in the `ViewModel`.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: `apply` is used to modify the object.
Here, it's not modified, it's only used.
It should be `with` instead.

Nevertheless, we tend to avoid using too much `this`, it can be confusing.
I propose this, if it's ok for you:
```kt
private fun setupDialogs() = with(mainViewModel) {
    bindAlertToViewLifecycle(descriptionDialog)
    linkContextualMenuAlertDialog.initValues(snackBarManager)
    emailContextualMenuAlertDialog.initValues(snackBarManager)
    phoneContextualMenuAlertDialog.initValues(snackBarManager)
}
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: This is the way we currently do probably everywhere in the app.
I'm open to suggestions to change it, but we should probably discuss it with the team before, because if we change this, it should be app-wide.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: Same answer as https://github.com/Infomaniak/android-kMail/pull/1572?w=1#discussion_r1420239922

FabianDevel (approved):

tevincent (approved):

PR #1573, opened 2023-11-30
T05: Barebone Tablet mode

KevinBoulongne:
Depends on #1572


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -474,29 +464,23 @@ class ThreadFragment : Fragment() {
         val shouldLoadDistantResources = shouldLoadDistantResources(message.uid)
 
         if (message.getRecipientsForReplyTo(replyAll = true).second.isEmpty()) {
-            safeNavigateToNewMessageActivity(
+            goToNewMessageActivity(
                 draftMode = DraftMode.REPLY,
                 previousMessageUid = message.uid,
                 shouldLoadDistantResources = shouldLoadDistantResources,
+                arrivedFromExistingDraft = false,
```
FabianDevel, line 471: Useless

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -196,6 +199,24 @@ class MainViewModel @Inject constructor(
...
+    val currentThreadUid = MutableLiveData<String?>(null)
+
+    val isInThread get() = currentThreadUid.value != null
+    val closeThreadTrigger = SingleLiveEvent<Unit>()
+
+    val downloadAttachmentsArgs = SingleLiveEvent<Triple<String, String, AttachmentType>>()
+    val newMessageArgs = SingleLiveEvent<NewMessageActivityArgs>()
+    val replyBottomSheetArgs = SingleLiveEvent<Pair<String, Boolean>>()
+    val threadActionsBottomSheetArgs = SingleLiveEvent<Triple<String, String, Boolean>>()
+    val messageActionsBottomSheetArgs = SingleLiveEvent<MessageActionsArgs>()
```
FabianDevel, line 212: We should create dataclass to encapsulate each args, put them in a separate file

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -290,7 +291,8 @@ fun Fragment.navigateToThread(
             )
         }
     } else {
-        safeNavigate(R.id.threadFragment, ThreadFragmentArgs(thread.uid).toBundle())
+        val uid = thread?.uid ?: threadUid ?: return@runCatchingRealm
```
FabianDevel, line 294: Put a com to explicit a bit this part

app/src/main/res/navigation/main_navigation.xml
```diff
@@ -54,6 +51,14 @@
             android:id="@+id/action_threadListFragment_to_invalidPasswordFragment"
             app:destination="@id/invalidPasswordFragment" />
 
+        <!-- Region Notification: Click to open Thread -->
+        <argument
+            android:name="openThreadUid"
+            android:defaultValue="@null"
+            app:argType="string"
+            app:nullable="true" />
+        <!-- Endregion -->
```
FabianDevel, line 60: Move the arguments before the actions

app/src/main/res/navigation/main_navigation.xml
```diff
@@ -54,6 +51,14 @@
             android:id="@+id/action_threadListFragment_to_invalidPasswordFragment"
             app:destination="@id/invalidPasswordFragment" />
 
+        <!-- Region Notification: Click to open Thread -->
```
FabianDevel, line 54: Uniformize region name with the others

gradle/libs.versions.toml
```diff
@@ -38,6 +39,7 @@ lottie = { module = "com.airbnb.android:lottie", version.ref = "lottie" }
 navigation-safeargs = { module = "androidx.navigation:navigation-safe-args-gradle-plugin", version.ref = "navigation" }
 realm-kotlin-base = { module = "io.realm.kotlin:library-base", version.ref = "realmKotlin" }
 sentry-android-fragment = { module = "io.sentry:sentry-android-fragment", version.ref = "sentryAndroidFragment" }
+slidingpanelayout = { module = "androidx.slidingpanelayout:slidingpanelayout", version.ref = "slidingpanelayout" }
```
FabianDevel, line 42: Rename this to snakeCase but keep the camel for the version

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: Will be done in another PR.

app/src/main/res/navigation/main_navigation.xml
KevinBoulongne: Will be done in another PR.

app/src/main/res/navigation/main_navigation.xml
KevinBoulongne: Will be done in another PR.

gradle/libs.versions.toml
KevinBoulongne: Ok for the camel, but not ok for the snake.

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -164,6 +163,23 @@ class ThreadListFragment : Fragment(), SwipeRefreshLayout.OnRefreshListener {
...
+                        navigateToThread(mainViewModel, threadUid = openThreadUid)
+                        arguments?.remove(navigationArgs::openThreadUid.name)
+                        threadListObserver?.let(threadListAdapter::unregisterAdapterDataObserver)
+                    }
+                }
+            }
+        }
+        threadListObserver?.let(threadListAdapter::registerAdapterDataObserver)
+    }
+
```
LunarX, line 182: Optimize the opening of notifications like the commit in `opti-opening-notifications` does so we don't wait on the adapter for no reason. Probably best to put it in its own PR though

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -196,6 +199,24 @@ class MainViewModel @Inject constructor(
     }
     //endregion
 
+    //region Current Thread
+    val currentThreadUid = MutableLiveData<String?>(null)
+
+    val isInThread get() = currentThreadUid.value != null
```
LunarX, line 205: `isInThread` is unused in this PR, is it normal?

LunarX (changes requested):
Not finished reviewing but here some first comments for now

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -74,31 +76,38 @@ class ThreadViewModel @Inject constructor(
...
-        messageController.getSortedAndNotDeletedMessagesAsync(threadUid)
-            ?.map { results -> results.list.map { splitBody(it) } }
-            ?.asLiveData()
-            ?.let { emitSource(it) }
+            splitBodies.clear()
+
+            messageController.getSortedAndNotDeletedMessagesAsync(threadUid)
+                ?.map { results -> results.list.map { splitBody(it) } }
+                ?.collect(messagesLive::postValue)
+        }
```
tevincent, line 107: This method is a bit complicated to understand. Maybe something like this? 

```kotlin
fun reassignMessagesLive(threadUid: String) {
        messagesLiveJob?.cancel()
        messagesLiveJob = viewModelScope.launch(ioCoroutineContext) {

            suspend fun splitBody(message: Message): Message = withContext(ioDispatcher) {
                val splitBodies = mutableMapOf<String, SplitBody>()

                if (message.body == null) return@withContext message

                message.apply {
                    body?.let {
                        val isNotAlreadySplit = !splitBodies.contains(message.uid)
                        if (isNotAlreadySplit) splitBodies[message.uid] = MessageBodyUtils.splitContentAndQuote(it)
                        splitBody = splitBodies[message.uid]
                    }
                }

                return@withContext message
            }

            messageController.getSortedAndNotDeletedMessagesAsync(threadUid)
                ?.map { results -> results.list.map { splitBody(it) } }
                ?.collect(messagesLive::postValue)
        }
    }
```

Also, you don't need that splitBodies variable to be a class variable. It could be within the inner function  ? Like in my suggestion ?

app/src/main/res/layout/fragment_search.xml
```diff
@@ -15,181 +15,195 @@
...
     android:layout_width="match_parent"
-    android:layout_height="match_parent">
+    android:layout_height="match_parent"
+    tools:context=".ui.main.search.SearchFragment">
 
-    <com.google.android.material.appbar.AppBarLayout
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content">
+    <androidx.coordinatorlayout.widget.CoordinatorLayout
+        android:layout_width="600dp"
```
tevincent, line 27: Are we sure we want 600dp on all resolution and all size of tablet ?

app/src/main/res/layout/fragment_search.xml
```diff
@@ -15,181 +15,195 @@
...
-            android:layout_height="wrap_content"
-            android:background="@color/backgroundColor"
-            android:clipToPadding="false"
-            android:paddingHorizontal="@dimen/marginStandardMedium"
-            android:scrollbars="none">
-
-            <LinearLayout
-                android:layout_width="wrap_content"
+            <HorizontalScrollView
+                android:id="@+id/horizontalScrollViewDummyId"
```
tevincent, line 57: Is that id necessary ?

app/src/main/res/layout/fragment_thread.xml
```diff
@@ -22,107 +22,122 @@
     android:layout_height="match_parent"
     tools:context=".ui.main.thread.ThreadFragment">
 
-    <com.google.android.material.appbar.MaterialToolbar
-        android:id="@+id/toolbar"
+    <androidx.constraintlayout.widget.ConstraintLayout
```
tevincent, line 25: Here you have a `ConstraintLayout` with two `ConstraintLayout`s inside. `ConstraintLayout`s are meant to avoid views intrication so here you might be able to just get rid of the two inner `ConstraintLayout`s ?

app/src/main/res/layout/fragment_thread.xml
```diff
@@ -22,107 +22,122 @@
...
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toBottomOf="@id/threadCoordinatorLayout"
+            app:menu="@menu/message_menu" />
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
tevincent, line 136: A `ConstraintLayout` for an empty view ?

app/src/main/res/layout/fragment_thread_list.xml
```diff
@@ -15,275 +15,289 @@
...
-                            android:text="@string/buttonInstall"
-                            app:icon="@drawable/ic_app_update"
-                            app:iconPadding="@dimen/marginStandardSmall"
-                            app:iconSize="@dimen/marginStandardMedium"
-                            app:layout_constraintEnd_toEndOf="parent"
-                            app:layout_constraintTop_toTopOf="parent" />
-
-                        <com.google.android.material.divider.MaterialDivider
-                            android:id="@+id/installUpdateDivider"
+                        <androidx.constraintlayout.widget.ConstraintLayout
```
tevincent, line 201: Maybe you can only have the `ConstraintLayout` here instead of a `ConstraintLayout` inside a `FrameLayout` ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -265,6 +248,18 @@ class ThreadFragment : Fragment() {
         phoneContextualMenuAlertDialog.initValues(mainViewModel.snackBarManager)
     }
 
+    private fun observeThreadOpening() = with(binding) {
+
+        mainViewModel.currentThreadUid.distinctUntilChanged().observeNotNull(viewLifecycleOwner) { threadUid ->
```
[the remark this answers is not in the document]
KevinBoulongne: We can use only one LiveData, but the logic will still be split because it won't be observed in the same Fragment.
Here, it's the ThreadFragment, but the closing of the Thread is in the ThePaneFragment.

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -373,7 +373,6 @@ class MainActivity : BaseActivity() {
     private fun setupSnackBar() {
         fun getAnchor(): View? = when (navController.currentDestination?.id) {
             R.id.threadListFragment -> findViewById(R.id.newMessageFab)
-            R.id.threadFragment -> findViewById(R.id.quickActionBar)
```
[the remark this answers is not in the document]
KevinBoulongne: True.
And in the Search, the Snackbar was totally on the quickActionBar.

It's fixed here: https://github.com/Infomaniak/android-kMail/pull/1597

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -265,6 +246,18 @@ class ThreadFragment : Fragment() {
...
+
+            // Display Thread view
+            emptyView.isGone = true
+            threadView.isVisible = true
```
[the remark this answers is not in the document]
KevinBoulongne: Done !

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -367,25 +358,28 @@ class ThreadFragment : Fragment() {
...
+            updateUi(threadUid, folderRole = mainViewModel.getActionFolderRole(result.thread))
+            updateAdapter(result)
         }
     }
 
-    private fun updateUi(folderRole: FolderRole?) = with(binding) {
+    private fun updateUi(threadUid: String, folderRole: FolderRole?) = with(binding) {
 
         iconFavorite.setOnClickListener {
             trackThreadActionsEvent(ACTION_FAVORITE_NAME, isFavorite)
```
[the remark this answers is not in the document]
KevinBoulongne: Done !

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -273,11 +273,18 @@ fun Fragment.safeNavigateToNewMessageActivity(args: Bundle? = null, currentClass
...
 }
 
-fun Fragment.navigateToThread(
+/**
+ * The `thread` is the default parameter for this function. But if we are coming from a Notification,
+ * we only have the `threadUid`. Since the `thread` is only used to know if it's `isOnlyOneDraft` or
+ * not, and that it can't be `true` if coming from a Notification, we can bypass the usage of the
+ * `thread` and directly use `threadUid` instead.
+ */
+fun TwoPaneFragment.navigateToThread(
```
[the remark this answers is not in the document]
KevinBoulongne: All `navigate` functions are purposely in the same place to be easier to find.

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -273,11 +273,18 @@ fun Fragment.safeNavigateToNewMessageActivity(args: Bundle? = null, currentClass
...
     mainViewModel: MainViewModel,
-    thread: Thread,
+    thread: Thread? = null,
+    threadUid: String? = null,
```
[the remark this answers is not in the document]
KevinBoulongne: Done here: https://github.com/Infomaniak/android-kMail/pull/1595?w=1

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: Oversight, it's fixed.

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
```diff
@@ -0,0 +1,137 @@
...
+import com.infomaniak.lib.core.utils.safeNavigate
+import com.infomaniak.mail.R
+import com.infomaniak.mail.ui.MainViewModel
+import com.infomaniak.mail.ui.main.search.SearchFragment
+import com.infomaniak.mail.ui.main.thread.DetailedContactBottomSheetDialogArgs
+import com.infomaniak.mail.ui.main.thread.ThreadFragment
+import com.infomaniak.mail.ui.main.thread.actions.*
+import com.infomaniak.mail.utils.safeNavigateToNewMessageActivity
+
+open class TwoPaneFragment : Fragment() {
```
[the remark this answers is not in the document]
KevinBoulongne: "two-pane" here is an compound adjective created with a number and a noun, so it should use the singular form of the noun.

And it's someone named _"Bodin"_ that said it, so it's definitively true :
https://yolainebodin.com/the-language-nook/english/compound-adjective-with-number-and-noun

=> _"Careful! Remember to keep the noun in the singular form even if the number makes you feel there should be an s for the plural because, as you know, it is now an adjective and there is no s at the end of adjectives in English!"_

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
```diff
@@ -0,0 +1,137 @@
...
+
+    private val slidingPaneLayout: SlidingPaneLayout
+        get() = when (this) {
+            is ThreadListFragment -> binding.threadListSlidingPaneLayout
+            is SearchFragment -> binding.searchSlidingPaneLayout
+            else -> throw IllegalStateException()
+        }
+
+    fun areBothShown() = !slidingPaneLayout.isSlideable
+    fun isOnlyLeftShown() = slidingPaneLayout.let { it.isSlideable && !it.isOpen }
```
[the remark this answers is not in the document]
KevinBoulongne: These are helpers that will be used in a following PR.

app/src/main/res/layout/fragment_search.xml
KevinBoulongne: No clue, you should see with @LunarX , he added it.

app/src/main/res/layout/fragment_thread.xml
KevinBoulongne: We know we want to put something in here, but we don't have any design yet.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
KevinBoulongne: True, `splitBodies` doesn't need to be a class variable, but we can't put it in `fun splitBody()`.
If we do that, it will reset each time `splitBody()` is called, defeating its purpose.

Instead, we can put it in `reassignMessagesLive ()`, just after launching the coroutine.

app/src/main/res/layout/fragment_thread_list.xml
KevinBoulongne: This is simplified in a following PR : https://github.com/Infomaniak/android-kMail/pull/1579

KevinBoulongne: Noted, will be handled in a foreseeable future, since it's not tied to the Tablet mode feature.

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
FabianDevel, line 35: This class is not meant to be instanciated, it should be abstract

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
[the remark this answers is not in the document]
LunarX: Actually place it directly inside TwoPaneFragment, it's only gonna be used by this class

app/src/main/res/layout/fragment_search.xml
LunarX: Yeah, by giving it an id, the scroll view will retain how much it has scrolled when the activity is recreated. Without the idea it's reset to 0 every time. We could add a comment nontheless

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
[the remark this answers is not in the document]
KevinBoulongne: Done !

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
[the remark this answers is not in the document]
KevinBoulongne: Done !

app/src/main/res/layout/fragment_thread_list.xml
```diff
@@ -15,275 +15,289 @@
...
 
-    <com.google.android.material.appbar.AppBarLayout
-        android:id="@+id/appBarLayout"
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
-        app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintStart_toStartOf="parent"
-        app:layout_constraintTop_toTopOf="parent">
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="600dp"
```
LunarX, line 27: Do we really want to only start showing the two panes above 600 + 550 dp and not before? Also isn't 600dp a bit big for some smaller tablette widths?

LunarX: Also this should be in a dimension if it is shared with another layout

LunarX (changes requested):

app/src/main/res/layout/fragment_search.xml
KevinBoulongne: Done.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -437,13 +422,7 @@ class ThreadFragment : Fragment() {
...
             startActivity(openWithIntent(requireContext()))
         } else {
-            safeNavigate(
-                ThreadFragmentDirections.actionThreadFragmentToDownloadAttachmentProgressDialog(
-                    attachmentResource = resource!!,
-                    attachmentName = name,
-                    attachmentType = getFileTypeFromMimeType(),
-                )
-            )
+            mainViewModel.downloadAttachmentsArgs.value = Triple(resource!!, name, getFileTypeFromMimeType())
```
LunarX, line 425: Does it make sens to turn this into a `SingleLiveEvent` rather than defining a method to call from here that directly executes what needs to be executed? Because it's only triggered from here and is only called once because it's a `SingleLiveEvent`, no?

LunarX: We need to think about it but maybe it could even make sens to extract it to a separate use case file depending on the situation

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -182,25 +170,21 @@ class ThreadFragment : Fragment() {
...
-        updateUi(folderRole)
     }
 
-    private fun initAdapter() = with(binding.messagesList) {
+    private fun setupAdapter() = with(binding.messagesList) {
         adapter = ThreadAdapter(
             shouldLoadDistantResources = shouldLoadDistantResources(),
             onContactClicked = { contact ->
-                safeNavigate(ThreadFragmentDirections.actionThreadFragmentToDetailedContactBottomSheetDialog(contact))
+                mainViewModel.detailedContactArgs.value = contact
```
LunarX, line 179: Same comment than the previous `downloadAttachmentsArgs` one. And actually the comment works probably for all of them

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -196,6 +199,23 @@ class MainViewModel @Inject constructor(
...
+    val currentThreadUid = MutableLiveData<String?>(null)
+
+    val closeThreadTrigger = SingleLiveEvent<Unit>()
+
+    val downloadAttachmentsArgs = SingleLiveEvent<Triple<String, String, AttachmentType>>()
+    val newMessageArgs = SingleLiveEvent<NewMessageActivityArgs>()
+    val replyBottomSheetArgs = SingleLiveEvent<Pair<String, Boolean>>()
+    val threadActionsBottomSheetArgs = SingleLiveEvent<Triple<String, String, Boolean>>()
+    val messageActionsBottomSheetArgs = SingleLiveEvent<MessageActionsArgs>()
+    val detailedContactArgs = SingleLiveEvent<Recipient>()
```
LunarX, line 212: Why were all of these extracted actually?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -265,6 +244,26 @@ class ThreadFragment : Fragment() {
...
     }
 
+    private fun observeThreadOpening() = with(threadViewModel) {
+
+        mainViewModel.currentThreadUid.distinctUntilChanged().observeNotNull(viewLifecycleOwner) { threadUid ->
+
+            reassignThreadLive(threadUid)
+            reassignMessagesLive(threadUid)
+
+            openThread(threadUid).observe(viewLifecycleOwner) { result ->
```
LunarX, line 254: This `openThread()` method needs the updated Thread object to work. Shouldn't we wait for `reassignThreadLive` to have finished before calling `openThread()` so we can skip on sending some of the `threadUid` values that are passed around everywhere and use the `threadLive` latest value instead?

app/src/main/res/layout/fragment_thread_list.xml
LunarX, line 1: I will write this here so we can have a thread on github.

We should avoid having the sliding animation of the pane because this won't be compatible with the feature of swiping between emails when we implement it. If the email swipes into view from the right and when you swipe it back away from the screen it goes to another email instead of discarding the pane, it will be very confusing and counter intuitive. Also, because this would be a change of habit for the users, I don't think we want to force them to get used to this new sliding behavior for them to have to get used back to the "non sliding" behavior again when we can swipe between emails.

Currently the color of the status bar is very chaotic. I think this needs to be thought through. UI/UX should be involved if needed but this needs to be tackled for the tablet update to roll out imo. This has been postponed for long enough

LunarX (changes requested):

app/src/main/res/layout/fragment_search.xml
KevinBoulongne: Same question as https://github.com/Infomaniak/android-kMail/pull/1573?w=1#pullrequestreview-1779351753 , so I'll answer there.

app/src/main/res/layout/fragment_thread_list.xml
KevinBoulongne: Why 600dp ? Because of this : https://developer.android.com/guide/topics/large-screens/support-different-screen-sizes

The 500dp is a little more arbitrary. If we were totally following the previously linked documentation, we should put 240dp instead of 500dp, but on some tablets it's not enough (for exemple, on Pixel C).

With some trial and error, I found 500dp to be nice. The value itself is not really important, it just needs to be big enough to not show up when the tablet is in portrait, but while still being inferior to the tablet width.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -367,25 +362,15 @@ class ThreadFragment : Fragment() {
...
-
-    private fun updateUi(folderRole: FolderRole?) = with(binding) {
+        // Display Thread view
+        emptyView.isGone = true
+        threadView.isVisible = true
 
         iconFavorite.setOnClickListener {
-            trackThreadActionsEvent(ACTION_FAVORITE_NAME, isFavorite)
-            mainViewModel.toggleThreadFavoriteStatus(navigationArgs.threadUid)
+            trackThreadActionsEvent(ACTION_FAVORITE_NAME, threadViewModel.threadLive.value?.isFavorite ?: false)
```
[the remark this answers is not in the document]
KevinBoulongne: Indeed, it doesn't.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: We need to navigate, and since we are in the ThreadFragment, we are out of navigation. So we can't do it here.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: Answered here : https://github.com/Infomaniak/android-kMail/pull/1573?w=1#discussion_r1425256192

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
KevinBoulongne: Answered here : https://github.com/Infomaniak/android-kMail/pull/1573#discussion_r1425256192

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: We can extract the `//region Current Thread` to a TwoPaneViewModel, to help lighten the MainViewModel.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: How do you want to do that exactly? I don't see any elegant way.

app/src/main/res/layout/fragment_thread_list.xml
LunarX: No that I see it, the statusbar issue also applies to the bottom navbar

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
KevinBoulongne: Fixed in https://github.com/Infomaniak/android-kMail/pull/1606?w=1

tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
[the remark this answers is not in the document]
LunarX: Now, does it really make sens to send a "favorite" matomo without any favorite value if it's null?

app/src/main/res/layout/fragment_thread_list.xml
KevinBoulongne: We still need to block TwoPane layout in portrait mode.
Right now, without handling configuration changes, it's impossible.
For now we block it until screens of 1100 width.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -158,10 +145,11 @@ class ThreadFragment : Fragment() {
         _binding = null
     }
 
-    private fun setupUi(folderRole: FolderRole?) = with(binding) {
+    private fun setupUi() = with(binding) {
 
         toolbar.setNavigationOnClickListener {
-            closeThread()
+            val onlyThreadIsShown = (parentFragment as TwoPaneFragment).isOnlyRightShown()
+            if (onlyThreadIsShown) mainViewModel.closeThread()
```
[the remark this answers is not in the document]
KevinBoulongne: Done in https://github.com/Infomaniak/android-kMail/pull/1592

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: Done here https://github.com/Infomaniak/android-kMail/pull/1606?w=1

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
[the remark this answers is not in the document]
LunarX: This is actually modified inside https://github.com/Infomaniak/android-kMail/pull/1580 instead

PR #1579, opened 2023-12-01
T07: Display Folder name in Thread empty view

KevinBoulongne:
Depends on #1578

PR #1580, opened 2023-12-01
T08: Update ThreadFragment toolbar icon when rotating device

KevinBoulongne:
Depends on #1579

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
@@ -109,13 +109,19 @@ class ThreadFragment : Fragment() {
     private val permissionUtils by lazy { PermissionUtils(this) }
     private val isNotInSpam by lazy { mainViewModel.currentFolder.value?.role != FolderRole.SPAM }
 
-    private val onlyThreadIsShown get() = (parentFragment as TwoPaneFragment).isOnlyRightShown()
+    private val twoPaneFragment get() = parentFragment as TwoPaneFragment
 
     // TODO: This is probably too global as a trigger. Find something more refined?
     private val globalLayoutListener by lazy {
         OnGlobalLayoutListener {
             runCatching {
-                binding.toolbar.navigationIcon?.alpha = if (onlyThreadIsShown) 255 else 0
+                binding.toolbar.navigationIcon?.apply {
+                    if (twoPaneFragment.areBothShown()) {
+                        if (alpha != 0) alpha = 0
+                    } else {
+                        if (alpha != 255) alpha = 255
+                    }
+                }
             }
         }
     }
@@ -171,7 +177,7 @@ class ThreadFragment : Fragment() {
     private fun setupUi() = with(binding) {
 
         toolbar.setNavigationOnClickListener {
-            if (onlyThreadIsShown) mainViewModel.closeThread()
+            if (twoPaneFragment.isOnlyRightShown()) mainViewModel.closeThread()
         }
 
         val defaultTextColor = context.getColor(R.color.primaryTextColor)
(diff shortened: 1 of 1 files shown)


FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -110,6 +111,17 @@ class ThreadFragment : Fragment() {
 
     private var isFavorite = false
 
+    private val onlyThreadIsShown get() = (parentFragment as TwoPaneFragment).isOnlyRightShown()
+
+    // TODO: This is probably too global as a trigger. Find something more refined?
+    private val globalLayoutListener by lazy {
+        OnGlobalLayoutListener {
+            runCatching {
+                binding.toolbar.navigationIcon?.alpha = if (onlyThreadIsShown) 255 else 0
```
tevincent, line 120: You can add constant for 255 and 0. Plus, to avoid setting in a loop the same alpha, you could also check the value if it's different from what you want to set ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: Done !

FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -141,16 +160,24 @@ class ThreadFragment : Fragment() {
...
+    private fun removeGlobalLayoutListener() {
+        binding.toolbar.viewTreeObserver.removeOnGlobalLayoutListener(globalLayoutListener)
+    }
+
     private fun setupUi() = with(binding) {
 
         toolbar.setNavigationOnClickListener {
-            val onlyThreadIsShown = (parentFragment as TwoPaneFragment).isOnlyRightShown()
-            if (onlyThreadIsShown) mainViewModel.closeThread()
+            if (twoPaneFragment.isOnlyRightShown()) mainViewModel.closeThread()
```
LunarX, line 180: We need to always close the thread when the button is clicked and only display the button when the thread can be closed.

I don't think we need bother with this logic here

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -108,6 +109,23 @@ class ThreadFragment : Fragment() {
...
+    // TODO: This is probably too global as a trigger. Find something more refined?
+    private val globalLayoutListener by lazy {
+        OnGlobalLayoutListener {
+            runCatching {
+                binding.toolbar.navigationIcon?.apply {
+                    if (twoPaneFragment.areBothShown()) {
+                        if (alpha != MIN_ALPHA) alpha = MIN_ALPHA
+                    } else {
+                        if (alpha != MAX_ALPHA) alpha = MAX_ALPHA
+                    }
```
LunarX, line 123: We should remove the button entirely and not make its icon invisible

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: Done here: https://github.com/Infomaniak/android-kMail/pull/1592

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: Updated here : https://github.com/Infomaniak/android-kMail/pull/1592

PR #1584, opened 2023-12-01
Make sure the Thread is always displayed when it's open

KevinBoulongne:
Depends on #1606


FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
```diff
@@ -175,6 +176,20 @@ open class TwoPaneFragment : Fragment() {
...
+        ensureThreadIsOpen()
+    }
+
+    override fun onResume() {
+        super.onResume()
+        ensureThreadIsOpen()
+    }
+
+    private fun ensureThreadIsOpen() {
+        if (mainViewModel.isInThread) openThread(mainViewModel.currentThreadUid.value!!)
```
tevincent, line 190:
```suggestion
        mainViewModel.currentThreadUid.value?.let { currentThreadUid ->
           if (mainViewModel.isInThread) openThread(currentThreadUid)
        }
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
KevinBoulongne: The whole purpose of `isInThread ` is to avoid writing that :D

tevincent: Yes but `!!` are evil. Don't be so close to the dark side of the force.

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
LunarX, line 1: Why is this needed? Because it looks a bit heavy to call `openThread()` again, this often

LunarX (changes requested):

PR #1591, opened 2023-12-06
Simplify shared preferences declaration with `sharedValue`

LunarX:


app/src/main/java/com/infomaniak/mail/utils/SharedValue.kt
```diff
@@ -0,0 +1,102 @@
...
+
+import android.content.SharedPreferences
+import com.infomaniak.lib.core.api.ApiController
+import com.infomaniak.lib.core.utils.Utils
+import com.infomaniak.lib.core.utils.transaction
+import kotlinx.serialization.encodeToString
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
+
+fun SharedPreferences.sharedValue(key: String, defaultValue: Boolean): ReadWriteProperty<Any, Boolean> {
```
FabianDevel, line 28: We should put this in Core

app/src/main/java/com/infomaniak/mail/utils/SharedValue.kt
```diff
@@ -0,0 +1,102 @@
...
+        override fun setValue(thisRef: Any, property: KProperty<*>, value: List<String>) {
+            transaction { putString(key, ApiController.json.encodeToString(value)) }
+        }
+    }
+}
+
+inline fun <reified E : Enum<E>> SharedPreferences.sharedValue(
+    key: String,
+    defaultValue: E
+): ReadWriteProperty<Any, E> {
```
FabianDevel, line 92: this can be onelined

FabianDevel (changes requested):
I haven't finished yet

app/src/main/java/com/infomaniak/mail/utils/SharedValue.kt
```diff
@@ -0,0 +1,102 @@
...
+}
+
+fun SharedPreferences.sharedValue(key: String, defaultValue: List<String>): ReadWriteProperty<Any, List<String>> {
+    return object : ReadWriteProperty<Any, List<String>> {
+        override fun getValue(thisRef: Any, property: KProperty<*>): List<String> {
+            return getString(key, null)?.let(ApiController.json::decodeFromString) ?: defaultValue
+        }
+
+        override fun setValue(thisRef: Any, property: KProperty<*>, value: List<String>) {
+            transaction { putString(key, ApiController.json.encodeToString(value)) }
```
FabianDevel, line 84: I don't like that the json object we use here in in ApiController

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/utils/SharedValue.kt
KevinBoulongne: ApiController is just a static object holding properties, this `json` property is perfectly fine.

PR #1586, opened 2023-12-01
T14: Temporary fix of `statusBarColor`

KevinBoulongne:
Depends on #1585

PR #1592, opened 2023-12-07
T15: Tablet easter egg

KevinBoulongne:
Depends on #1586

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
@@ -23,6 +23,7 @@ import android.content.Intent
 import android.os.Bundle
 import android.os.CountDownTimer
 import android.view.View
+import android.view.ViewGroup
 import android.webkit.WebView
 import androidx.activity.addCallback
 import androidx.activity.result.contract.ActivityResultContracts.*
@@ -510,6 +511,8 @@ class MainActivity : BaseActivity() {
     fun navigateToSyncAutoConfigActivity() {
         syncAutoConfigActivityResultLauncher.launch(Intent(this, SyncAutoConfigActivity::class.java))
     }
+
+    fun getConfettiContainer(): ViewGroup = binding.easterEggConfettiContainer
 
     companion object {
         const val DRAFT_ACTION_KEY = "draftAction"
--- a/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
@@ -77,7 +77,6 @@ import javax.inject.Inject
 @AndroidEntryPoint
 class MainActivity : BaseActivity() {
 
-    // This binding is not private because it's used in ThreadListFragment (`(activity as? MainActivity)?.binding`)
     val binding by lazy { ActivityMainBinding.inflate(layoutInflater) }
     private val mainViewModel: MainViewModel by viewModels()
 
@@ -89,8 +88,6 @@ class MainActivity : BaseActivity() {
     private val registerFirebaseBroadcastReceiver by lazy { RegisterFirebaseBroadcastReceiver() }
 
     private var previousDestinationId: Int? = null
-    private var easterEggConfettiCount = 0
-    private var easterEggConfettiTime = 0L
 
     private val navController by lazy {
         (supportFragmentManager.findFragmentById(R.id.mainHostFragment) as NavHostFragment).navController
@@ -514,23 +511,6 @@ class MainActivity : BaseActivity() {
         syncAutoConfigActivityResultLauncher.launch(Intent(this, SyncAutoConfigActivity::class.java))
     }
 
-    fun onEasterEggConfettiClicked(matomoValue: String) {
-
-        val currentTime = System.currentTimeMillis()
-
-        if (easterEggConfettiTime == 0L || currentTime - easterEggConfettiTime > EASTER_EGG_CONFETTI_TRIGGER_DELAY) {
-            easterEggConfettiTime = currentTime
-            easterEggConfettiCount = 1
-        } else {
-            easterEggConfettiCount++
-        }
-
-        if (easterEggConfettiCount == EASTER_EGG_CONFETTI_TRIGGER_TAPS) {
-            easterEggConfettiCount = 0
-            ConfettiUtils.triggerEasterEggConfetti(binding.easterEggConfettiContainer, matomoValue)
-        }
-    }
-
     companion object {
         const val DRAFT_ACTION_KEY = "draftAction"
         const val SYNC_AUTO_CONFIG_KEY = "syncAutoConfigKey"
@@ -538,7 +518,5 @@ class MainActivity : BaseActivity() {
         const val SYNC_AUTO_CONFIG_ALREADY_SYNC = "syncAutoConfigAlreadySync"
 
         private const val FULLY_SLID = 1.0f
-        private const val EASTER_EGG_CONFETTI_TRIGGER_TAPS = 3
-        private const val EASTER_EGG_CONFETTI_TRIGGER_DELAY = 1_000L
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
@@ -165,7 +165,9 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
         }
 
         appVersionName.setOnClickListener {
-            (requireActivity() as MainActivity).onEasterEggConfettiClicked(matomoValue = "MenuDrawer")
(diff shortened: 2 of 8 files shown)


app/src/main/java/com/infomaniak/mail/utils/ConfettiUtils.kt
```diff
@@ -113,13 +133,17 @@ object ConfettiUtils {
...
+            displaySnow(colored = true)
+        } else {
+            when ((0..2).random()) {
+                0 -> displayTada()
+                1 -> displaySnow()
+                else -> if (isInPortrait()) {
+                    displaySingleGeneva()
+                } else {
+                    displayDoubleGeneva()
+                }
```
FabianDevel, line 146: oneline this

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -111,11 +112,21 @@ class ThreadFragment : Fragment() {
...
-
     private val globalLayoutListener = OnGlobalLayoutListener {
         runCatching {
-            binding.toolbar.navigationIcon?.alpha = if (bothPanesAreShown) 0 else 255
+            with(binding.toolbar) {
+
+                val bothPanesAreShown = (parentFragment as TwoPaneFragment).areBothShown()
+
+                val (alpha, callback) = if (bothPanesAreShown) {
+                    0 to { (requireActivity() as MainActivity).onEasterEggConfettiClicked(matomoValue = "Tablet") }
```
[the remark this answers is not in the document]
FabianDevel: Yes but the function onEasterEggConfettiClicked and the variable declared in MainActivity could be moved to ConfettiUtils
We only need to pass the view as parameters from mainActivity
All would be better encapsulated

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -77,7 +77,6 @@ import javax.inject.Inject
 @AndroidEntryPoint
 class MainActivity : BaseActivity() {
 
-    // This binding is not private because it's used in ThreadListFragment (`(activity as? MainActivity)?.binding`)
```
LunarX, line 80: binding is still public, why remove the comment?

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
```diff
@@ -165,7 +165,9 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
         }
 
         appVersionName.setOnClickListener {
-            (requireActivity() as MainActivity).onEasterEggConfettiClicked(matomoValue = "MenuDrawer")
+            (activity as? MainActivity)?.binding?.easterEggConfettiContainer?.let {
+                ConfettiUtils.onEasterEggConfettiClicked(container = it, matomoValue = "MenuDrawer")
+            }
```
LunarX, line 170: We should do something more reusable than relying on the MainActivity and aime for weaker coupling

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
LunarX, line 1: I think we should place this easter egg elsewhere but not here. We should simply remove the back arrow when we don't need it so the title is not strangely spaced away from the left edge of the ThreadFragment

app/src/main/java/com/infomaniak/mail/utils/ConfettiUtils.kt
```diff
@@ -28,9 +29,48 @@ import com.infomaniak.lib.confetti.R as RConfetti
...
-    fun triggerEasterEggConfetti(container: ViewGroup, matomoValue: String): Unit = with(container.context) {
+    private const val AQUA = -0xFF0001
+    private const val FUCHSIA = -0xFF01
+    private const val LIME = -0xFF0100
+    private const val MAROON = -0x800000
+    private const val NAVY = -0xFFFF80
+    private const val OLIVE = -0x7F8000
+    private const val PURPLE = -0x7FFF80
+    private const val SILVER = -0x3F3F40
+    private const val TEAL = -0xFF7F80
```
LunarX, line 45: If you need to specify hex colors, don't specify hexadecimal colors as negative numbers this makes it unintelligible when you try to read the numbers, use the standard notation `0xAARRGGBB`.

That being said, here, you should use a resources to store these values inside a color array

app/src/main/java/com/infomaniak/mail/utils/ConfettiUtils.kt
```diff
@@ -28,9 +29,48 @@ import com.infomaniak.lib.confetti.R as RConfetti
...
+
+    private val COLORED_SNOW_COLORS = arrayOf(
+        Color.BLACK, Color.DKGRAY, Color.GRAY, Color.LTGRAY, Color.WHITE, Color.RED,
+        Color.GREEN, Color.BLUE, Color.YELLOW, Color.CYAN, Color.MAGENTA,
+        AQUA, FUCHSIA, LIME, MAROON, NAVY, OLIVE, PURPLE, SILVER, TEAL,
+    ).toIntArray()
```
LunarX, line 51: Not needed if you can use a ressource color array

app/src/main/java/com/infomaniak/mail/utils/ConfettiUtils.kt
```diff
@@ -113,13 +156,13 @@ object ConfettiUtils {
...
         }
 
-        when ((0..2).random()) {
-            0 -> displayTada()
-            1 -> displaySnow()
-            else -> if (isInPortrait()) {
-                displaySingleGeneva()
-            } else {
-                displayDoubleGeneva()
+        if (matomoValue == "Tablet") {
```
LunarX, line 159: Not sure we should do conditions on matomo values

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/ConfettiUtils.kt
FabianDevel: I agree

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
KevinBoulongne: This comment dates back to when it was the 1st public binding.
Now there is like 4 or 5 public bindings, so it's a normal behavior.
Hence the comment loses its interest.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
FabianDevel: In a way, the fact that this is strangely spaced away is actually why users could find this strange and find the easter egg

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
```diff
@@ -165,7 +165,9 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
         }
 
         appVersionName.setOnClickListener {
-            (requireActivity() as MainActivity).onEasterEggConfettiClicked(matomoValue = "MenuDrawer")
+            (activity as? MainActivity)?.binding?.easterEggConfettiContainer?.let {
```
FabianDevel, line 168: make a getConfettiContainer() function in mainActivity, it would reduce binding access and also it is used 2 times so factorizing it would be good

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
KevinBoulongne: In fact, this binding is now private : https://github.com/Infomaniak/android-kMail/pull/1606

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -105,12 +105,11 @@ class ThreadFragment : Fragment() {
     private val mainViewModel: MainViewModel by activityViewModels()
     private val threadViewModel: ThreadViewModel by viewModels()
 
+    private val twoPaneFragment inline get() = parentFragment as TwoPaneFragment
```
LunarX, line 108: This could be abstracted to lower coupling but maybe ThreadFragment is really really always meant to be displayed inside a TwoPaneFragment so yeah...

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
KevinBoulongne: ThreadFragment is really really always meant to be displayed inside a TwoPaneFragment :)

FabianDevel (approved):

LunarX (approved):

PR #1599, opened 2023-12-11
Xmas easter egg

JorisBodin:


KevinBoulongne (approved):

LunarX (changes requested):
Shouldn't we put all easter egg animations in a single file (use case) and call them where they are needed to store them together and have them be groupped thematically?

JorisBodin:
> Shouldn't we put all easter egg animations in a single file (use case) and call them where they are needed to store them together and have them be groupped thematically?

In another PR

PR #1563, opened 2023-11-22
In-App update

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/104

PR #1600, opened 2023-12-12
Add worker to update app

FabianDevel:
Same as #1590, seems this was not merged correctly in Master
My guess is, it was merged in its parent #1563, at the same time the parent was being rebased on Master, which dismissed the merge of #1590 by pushing the local branch without this merge

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/workers/AppUpdateWorker.kt
+++ b/app/src/main/java/com/infomaniak/mail/workers/AppUpdateWorker.kt
@@ -21,6 +21,7 @@ import android.content.Context
 import androidx.concurrent.futures.CallbackToFutureAdapter
 import androidx.hilt.work.HiltWorker
 import androidx.work.*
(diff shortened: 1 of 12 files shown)


app/src/main/java/com/infomaniak/mail/workers/AppUpdateWorker.kt
```diff
@@ -0,0 +1,112 @@
...
+            workInfos.forEach {
+                workManager.cancelWorkById(it.id)
+                SentryLog.d(TAG, "Work cancelled")
+            }
+        }
+    }
+
+    companion object {
+        private const val TAG = "AppUpdateWorker"
+        private const val INITIAL_DELAY = 10L //10s
```
KevinBoulongne:
```suggestion
        private const val INITIAL_DELAY = 10L // 10s
```

KevinBoulongne (approved):

PR #1602, opened 2023-12-12
Escape subject and recipients when inserting them in reply/forward

LunarX:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/DraftController.kt
```diff
@@ -188,27 +191,72 @@ class DraftController @Inject constructor(
...
+        formatRecipientList(recipientList)?.let { recipients -> addAndEscapeTextLine("$prefix $recipients") }
+    }
+
+    private fun formatRecipientList(recipientList: List<Recipient>): String? {
+        return if (recipientList.isNotEmpty()) recipientList.joinToString { it.quotedDisplay() } else null
+    }
+
+    private fun Element.addAlreadyEscapedBody(previousFullBody: String) {
+        append(previousFullBody)
+    }
```
FabianDevel, line 255: Is this function really useful ? Doesn't seems like it
If any, a commentary would be enough to tell it was already escaped

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/DraftController.kt
```diff
@@ -188,27 +191,72 @@ class DraftController @Inject constructor(
...
+            addAndEscapeTextLine("---------- $messageForwardHeader ---------")
+            addAndEscapeTextLine("$fromTitle ${message.fromName()}")
+            addAndEscapeTextLine("$dateTitle ${message.date.toDate()}")
+            addAndEscapeTextLine("$subjectTitle ${message.subject}")
+            addAndEscapeRecipientLine(toTitle, message.to)
+            addAndEscapeRecipientLine(ccTitle, message.cc)
+            addAndEscapeTextLine("")
+            addAndEscapeTextLine("")
+
+            addAlreadyEscapedBody(previousFullBody)
```
FabianDevel, line 229: Is it normal that we add the body in a function assembling an header ?

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/DraftController.kt
LunarX: This is mainly because we need to make sure both places where this is called always do the same things and they must never diverge. It just so happens that the thing that needs to be done is quiet simple, at least for now. 

Also yes the function name gives a bit more insight, but as you said, it could be conveyed using a comments as well

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/DraftController.kt
LunarX: Maybe naming it "footer" would be more fitting as this is actually what comes at the end of emails

PR #1603, opened 2023-12-12
Extract reply/forward header creation code in their own use case file

LunarX:
Cleaning this code up made me realize how little our processing of emails' html is centralized. This needs to be fixed in a following PR. We access the field `value` from `Body` objects way too often when we never should access it raw like this but rather have the final html be:
* created with subbodies 
* wrapped if it was a text/plain
* sanitized

This result should be the only value we ever access

Depends on #1602

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
@@ -35,7 +35,7 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class ReplyForwardHeaderManager @Inject constructor(private val appContext: Context) {
+class ReplyForwardFooterManager @Inject constructor(private val appContext: Context) {
 
     fun createForwardFooter(message: Message, attachmentsToForward: List<Attachment>): String {
         val previousBody = getHtmlDocument(message)?.let { document ->
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
@@ -90,9 +90,7 @@ class ReplyForwardFooterManager @Inject constructor(private val appContext: Cont
         }
     }
 
-    private fun Message.fromName(): String {
-        return sender?.quotedDisplay() ?: appContext.getString(R.string.unknownRecipientTitle)
-    }
+    private fun Message.fromName(): String = sender?.quotedDisplay() ?: appContext.getString(R.string.unknownRecipientTitle)
 
     private fun getHtmlDocument(message: Message): Document? {
         val html = message.body?.let { body ->
@@ -121,7 +119,7 @@ class ReplyForwardFooterManager @Inject constructor(private val appContext: Cont
 
     private fun assembleForwardHtmlFooter(
         message: Message,
-        previousFullBody: String
+        previousFullBody: String,
     ): String = with(appContext) {
         val messageForwardHeader = getString(R.string.messageForwardHeader)
         val fromTitle = getString(R.string.fromTitle)
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
@@ -38,24 +38,20 @@ import javax.inject.Singleton
 class ReplyForwardFooterManager @Inject constructor(private val appContext: Context) {
 
     fun createForwardFooter(message: Message, attachmentsToForward: List<Attachment>): String {
-        val previousBody = getHtmlDocument(message)?.let { document ->
-            val attachmentsMap = message.attachments.associate { oldAttachment ->
-                val newAttachment = attachmentsToForward.find { it.originalContentId == oldAttachment.contentId }
-
-                oldAttachment.contentId to newAttachment?.contentId
-            }
-
-            document.doOnHtmlImage { imageElement ->
-                attachmentsMap[getCid(imageElement)]?.let { newContentId ->
+        val previousBody = getHtmlDocument(message)?.apply {
+            processCids(
+                message = message,
+                associateDataToCid = { oldAttachment ->
+                    val newAttachment = attachmentsToForward.find { it.originalContentId == oldAttachment.contentId }
+                    newAttachment?.contentId
+                },
+                applyAssociatedDataToImage = { newContentId, imageElement ->
                     imageElement.attr(SRC_ATTRIBUTE, "${CID_PROTOCOL}$newContentId")
                 }
-            }
-
-            return@let document.outerHtml()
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardHeaderManager.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    private fun getCid(imageElement: Element) = imageElement.attr(SRC_ATTRIBUTE).removePrefix(CID_PROTOCOL)
+
+    private fun computePreviousFullBody(previousBody: String, message: Message): String {
+        return message.body?.let { body ->
+            MessageBodyUtils.mergeSplitBodyAndSubBodies(previousBody, body.subBodies, message.uid)
+        } ?: previousBody
+    }
+
+    private fun Recipient.quotedDisplay(): String = "${("$name ").ifBlank { "" }}<$email>"
```
FabianDevel, line 111: Why not checking if name is blank directly ?
Or maybe checking is trimmed name is Empty ?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardHeaderManager.kt
```diff
@@ -0,0 +1,183 @@
...
+
+    private fun assembleForwardHtmlHeader(
+        message: Message,
+        previousFullBody: String
```
FabianDevel, line 115:
```suggestion
        previousFullBody: String,
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardHeaderManager.kt
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    private fun Element.addAlreadyEscapedBody(previousFullBody: String) {
+        append(previousFullBody)
+    }
+
+    private fun Element.addReplyBlockQuote(addInnerElements: Element.() -> Unit) {
+        val blockQuote = appendElement("blockquote").addClass("ws-ng-quote")
+        blockQuote.addInnerElements()
+    }
```
FabianDevel, line 172: What about making these functions returns the new elements / themselves so we can chain calls ?

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardHeaderManager.kt
LunarX: I don't know if we want to chain them. `apply` already kinda gives us the same way of writing things like "one row after the other"

FabianDevel: I understand, I just felt that it would have followed the "native way" of creating html

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
```diff
@@ -0,0 +1,190 @@
...
+        val previousBody = getHtmlDocument(message)?.apply {
+            processCids(
+                message = message,
+                associateDataToCid = { oldAttachment ->
+                    val newAttachment = attachmentsToForward.find { it.originalContentId == oldAttachment.contentId }
+                    newAttachment?.contentId
+                },
+                applyAssociatedDataToImage = { newContentId, imageElement ->
+                    imageElement.attr(SRC_ATTRIBUTE, "${CID_PROTOCOL}$newContentId")
+                }
```
FabianDevel, line 50:
```suggestion
                },
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
```diff
@@ -0,0 +1,190 @@
...
+        val from = message.fromName()
+        val messageReplyHeader = appContext.getString(R.string.messageReplyHeader, date, from)
+
+        val previousBody = getHtmlDocument(message)?.apply {
+            processCids(
+                message = message,
+                associateDataToCid = Attachment::name,
+                applyAssociatedDataToImage = { name, imageElement ->
+                    imageElement.replaceWith(TextNode("<$name>"))
+                }
```
FabianDevel, line 69:
```suggestion
                },
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
```diff
@@ -0,0 +1,190 @@
...
+                message = message,
+                associateDataToCid = { oldAttachment ->
+                    val newAttachment = attachmentsToForward.find { it.originalContentId == oldAttachment.contentId }
+                    newAttachment?.contentId
+                },
+                applyAssociatedDataToImage = { newContentId, imageElement ->
+                    imageElement.attr(SRC_ATTRIBUTE, "${CID_PROTOCOL}$newContentId")
+                }
+            )
+        }?.outerHtml() ?: ""
```
FabianDevel, line 52:
```suggestion
        val previousBody = getHtmlDocument(message)?.let { document ->
            document.processCids(
                message = message,
                associateDataToCid = { oldAttachment ->
                    val newAttachment = attachmentsToForward.find { it.originalContentId == oldAttachment.contentId }
                    newAttachment?.contentId
                },
                applyAssociatedDataToImage = { newContentId, imageElement ->
                    imageElement.attr(SRC_ATTRIBUTE, "${CID_PROTOCOL}$newContentId")
                },
            )
            
            document.outerHtml()
        } ?: ""
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardFooterManager.kt
```diff
@@ -0,0 +1,190 @@
...
+
+        val previousBody = getHtmlDocument(message)?.apply {
+            processCids(
+                message = message,
+                associateDataToCid = Attachment::name,
+                applyAssociatedDataToImage = { name, imageElement ->
+                    imageElement.replaceWith(TextNode("<$name>"))
+                }
+            )
+        }?.outerHtml() ?: ""
```
FabianDevel, line 71:
```suggestion
        val previousBody = getHtmlDocument(message)?.let { document ->
            document.processCids(
                message = message,
                associateDataToCid = Attachment::name,
                applyAssociatedDataToImage = { name, imageElement ->
                    imageElement.replaceWith(TextNode("<$name>"))
                },
            )

            document.outerHtml()
        } ?: ""
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ReplyForwardHeaderManager.kt
LunarX: Honestly the "native way" is only due to it being Java code

FabianDevel (approved):

PR #1608, opened 2023-12-14
Add transitions in menu drawer

FabianDevel:
Depends on #1607

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
@@ -211,13 +211,10 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
     }
 
     private fun addMenuDrawerTransition(shouldFade: Boolean = true) {
-        val transition = if (shouldFade) {
-            TransitionSet()
-                .addTransition(ChangeBounds())
-                .addTransition(Fade(Fade.IN))
-        } else {
-            ChangeBounds()
-        }.setDuration(MENU_DRAWER_TRANSITION_DURATION)
+        val transition = TransitionSet()
+            .addTransition(ChangeBounds())
+            .also { if (shouldFade) it.addTransition(Fade(Fade.IN)) }
+            .setDuration(MENU_DRAWER_TRANSITION_DURATION)
 
         TransitionManager.beginDelayedTransition(binding.drawerContentScrollView, transition)
     }
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
@@ -20,6 +20,8 @@ package com.infomaniak.mail.ui.main.menu
 import android.annotation.SuppressLint
 import android.content.Intent
 import android.os.Bundle
+import android.transition.ChangeBounds
+import android.transition.TransitionManager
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
```diff
@@ -199,6 +206,22 @@ class MenuDrawerFragment : MenuFoldersFragment(), MailboxListFragment {
...
+        val transition = if (shouldFade) {
+            TransitionSet()
+                .addTransition(ChangeBounds())
+                .addTransition(Fade(Fade.IN))
+        } else {
+            ChangeBounds()
+        }.setDuration(MENU_DRAWER_TRANSITION_DURATION)
+
+        TransitionManager.beginDelayedTransition(binding.drawerContentScrollView, transition)
+    }
```
LunarX, line 223: Would this gain from being written something along the lines of: 
``` kt
TransitionSet()
    .addTransition(ChangeBounds())
    .also { if (shouldFade) it.addTransition(Fade(Fade.IN)) }
    .setDuration(MENU_DRAWER_TRANSITION_DURATION)
```

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuFoldersFragment.kt
```diff
@@ -53,17 +53,29 @@ abstract class MenuFoldersFragment : Fragment() {
...
+            onCollapseClicked = ::onFolderCollapse,
+            onCollapseTransition = ::onCollapseTransition,
+        )
     }
 
     protected abstract fun onFolderSelected(folderId: String)
 
     protected abstract fun onFolderCollapse(folderId: String, shouldCollapse: Boolean)
 
+    protected abstract fun onCollapseTransition()
```
LunarX, line 77: Do we prefer to make it abstract so everyone needs to acknowledge it or do we simply implement it so it returns Unit by default and the children that want to override it take care of overriding it?

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
LunarX, line 1: I'm afraid we might be calling it at a wrong time because it only expends with the correct animation but doesn't shrink with the correct animation*

LunarX (changes requested):
I'm not a fan of the fact that the animation doesn't work when we shrink the content of the recyclerview, that the items of the recyclerview move with a different interpolator than the ChangeBounds() we added and that the ChangeBounds() makes items quickly fly through the screen because they were last seen outside the screen and need to come back to where they are now.

If the team wants to add the feature knowing these issues anyway, why not

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuFoldersFragment.kt
FabianDevel: Well I like to make it abstract because if you extends it you don't risk to forget about it
On the other hand this is not needed everywhere

I personnaly choose abstract as you can see, but I don't have a strong opinion on it

app/src/main/java/com/infomaniak/mail/ui/main/menu/MenuDrawerFragment.kt
LunarX: That's definitely what I meant, I don't know about performances either

LunarX (approved):
Still the issue of different speed is so annoying but at least now it works much better overall. So let's go for it then

PR #1612, opened 2023-12-20
T23: Smooth sliding

KevinBoulongne:
Depends on #1613


LunarX (changes requested):
Instead of delaying the loading of the thread shouldn't we rather fix the fact that data is loaded on the main thread? This feels like the core underlying issue here

Also if we plan to prevent the pane from sliding this branch and the complexity of the code that comes with it is not needed anymore at all

PR #1613, opened 2023-12-21
T22: Progressively update statusBar & navigationBar colors when sliding pane

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
```diff
@@ -104,6 +114,26 @@ abstract class TwoPaneFragment : Fragment() {
         }
     }
 
+    override fun onPanelOpened(panel: View) = Unit
+
+    override fun onPanelClosed(panel: View) = Unit
```
LunarX, line 119: On the menu drawer, I think sometimes the panel never showed 0.0 or 1.0  inside the "sliding" listener. For example, the maximum value returned by sliding was always like 0.9 something. So I needed the opened or the closed listener to set the 100% color. Is this not an issue here?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
KevinBoulongne: After adding some logs and doing some testing, it looks like this listener does trigger 0.0 & 1.0 every time 🎉

LunarX (approved):

PR #1620, opened 2024-01-03
Change values() to entries after kotlin 1.9

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/115
Depends on #1622


sirambd (approved):

LunarX (changes requested):
Your branch needs to target https://github.com/Infomaniak/android-kMail/pull/1622 because right now it doesn't build

LunarX (approved):

PR #1622, opened 2024-01-03
feat: Migrate room to ksp

sirambd:


FabianDevel (changes requested):
Shouldn't we also remove kapt in this gradle Files ?

FabianDevel (approved):

PR #1623, opened 2024-01-03
Add missing Matomo

FabianDevel:


app/src/main/java/com/infomaniak/mail/MatomoMail.kt
```diff
@@ -183,7 +185,7 @@ object MatomoMail : MatomoCore {
     }
 
     fun Fragment.trackOnBoardingEvent(name: String) {
-        trackEvent("onBoarding", name)
+        trackEvent("onboarding", name)
```
LunarX, line 188: Do we want to change the category name after using it for so long?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
LunarX, line 1: If we're adding the missing matomo, shouldn't we also add the missing `notificationAction` values of `reply` and `open`?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
FabianDevel: yes

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
FabianDevel: It's already done, the docs was never changed, i'll update it

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
LunarX: why here, compared to other times where we don't?

FabianDevel: Because we are dissociated with iOS here, which isn't good 
And changing category has less impact than changing event names because we make our queries on event names
Also (and the fact that it is needed to do it fast now), we are entering a new year, and as we query on a full year time, it isn't a problem if it's different from the past year

LunarX (approved):

PR #1643, opened 2024-01-15
Display Sync bottomSheet at 1st boot (instead of 5th boot)

KevinBoulongne:


FabianDevel (changes requested):
It will clash with the future "Permissions Onboarding" feature, we should talk about this to come to a solution

KevinBoulongne:
> It will clash with the future "Permissions Onboarding" feature, we should talk about this to come to a solution

We merge it like that for now, and in this future feature, it will be move to the onboarding.

PR #1644, opened 2024-01-15
Use initials instead of grey image as Avatars' placeholder

KevinBoulongne:
Depends on https://github.com/Infomaniak/android-core/pull/119


Core
LunarX, line 1: So what's the choice for the other projects? We change their behavior as well?

LunarX, line 1: This new fix makes the user avatar blinking much more apparent. We should use this opportunity to fix this at the same time because merging this feels worse without the second problem being fixed as well

LunarX: You can easily reproduce it if you try and find your own email address when searching contacts to write to

KevinBoulongne: Fixed in https://github.com/Infomaniak/android-kMail/pull/1653

PR #1645, opened 2024-01-15
Clean Avatar code

KevinBoulongne:


app/src/main/res/layout/cardview_thread_item.xml
```diff
@@ -44,6 +44,7 @@
             android:id="@+id/threadCard"
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
+            android:paddingStart="0dp"
```
LunarX: not sure we want to suppress the link this way, if we need to specify every padding that's 0dp we've got we're gonna need to add a lot of empty lines

LunarX (approved):

PR #1679, opened 2024-01-29
Add possibility to reply to calendar events

LunarX:
The only way to have a not too ugly result was to handle the user's status change locally and not fetch the API to not trigger the adapter and bind the whole message again causing it to blink after each click.

This also means that right now, after the user has replied and left the thread, if he loses his internet connexion and comes back in the thread, the status of the user will be the one we had inside Realm when he last opened the thread and not the one he answered with. I think that this can be considered uncommon enough to let it slide

---

Nevermind, the adapter now detects if only the attendance state of attendee have changed and if so, only the precise UI elements associated to this get to bind instead of the whole message

Depends on #1670

PR #1680, opened 2024-01-29
Calendar event matomo

LunarX:
Depends on #1679

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
+++ b/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
@@ -1,6 +1,6 @@
 /*
  * Infomaniak Mail - Android
- * Copyright (C) 2022-2023 Infomaniak Network SA
+ * Copyright (C) 2022-2024 Infomaniak Network SA
  *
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/MatomoMail.kt
```diff
@@ -248,6 +249,10 @@ object MatomoMail : MatomoCore {
         trackEvent("inAppUpdate", name)
     }
 
+    fun View.trackCalendarEventEvent(name: String, value: Float? = null) {
```
FabianDevel, line 252: For the value, directly pass a boolean instead of a float, then use the extension toMailActionValue on it, to abstract the cast

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
LunarX: I passed a Boolean instead of a Float and made the `toFloat()` call inside the function rather than outside of it

FabianDevel (approved):

PR #1685, opened 2024-02-02
Fix Search with Attachment filter

FabianDevel:


app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -195,7 +195,7 @@ class Message : RealmObject {
         isTrashed: Boolean,
         isFromSearch: Boolean,
         draftLocalUuid: String?,
-        latestCalendarEventResponse: CalendarEventResponse?,
+        latestCalendarEventResponse: CalendarEventResponse? = null,
```
LunarX, line 198: Should we not force it to be passed so people don't forget to give it a value if they call this method?

FabianDevel: Seems reasonnable

KevinBoulongne (approved):

PR #1687, opened 2024-02-02
Directly open Attachment when clicking on it

FabianDevel:


app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
```diff
@@ -275,6 +278,18 @@ fun Fragment.safeNavigateToNewMessageActivity(
...
+    attachment: Attachment,
+    attachmentIntentType: AttachmentIntentType,
+    currentClassName: String = AttachmentActionsBottomSheetDialog::class.java.name,
+) {
+    safeNavigate(
+        resId = R.id.downloadAttachmentProgressDialog,
+        args = attachment.createDownloadDialogNavArgs(attachmentIntentType),
+        currentClassName = currentClassName,
+    )
+}
```
tevincent, line 292: Do we still want to add more extensions to that file ? Maybe in a `NavigationExtensions` file ? We also need to move all `safeNavigate` extensions there.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/utils/Extensions.kt
FabianDevel: You're right, I added there because of habit and to put it with the other navigateToX() functions, I'll move them all in a new File

tevincent (approved):

PR #1684, opened 2024-02-01
Add SuperCollapsedBlock

KevinBoulongne:
Depends on #1706
Depends on #1683
Depends on https://github.com/Infomaniak/android-core/pull/125

PR #1708, opened 2024-02-20
Reset Calendar data when switching Threads

KevinBoulongne:
Depends on #1684


tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -101,7 +103,9 @@ class ThreadViewModel @Inject constructor(
     ).map { it.obj }.asLiveData(ioCoroutineContext)
 
     fun resetMessagesCache() {
-        cachedSplitBodies = mutableMapOf()
+        treatedMessagesForCalendarEvent.clear()
+        isCalendarEventExpandedMap.clear()
```
tevincent, line 107: We might need to rename this method because it's not obvious that the calendar is reset when we reset a message. Maybe `resetMessagesRelatedCache`?

tevincent (changes requested):

PR #1718, opened 2024-02-27
Restore Thread state when coming back

KevinBoulongne:
Depends on #1723

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
@@ -45,7 +45,7 @@ import javax.inject.Inject
 abstract class TwoPaneFragment : Fragment() {
 
     val mainViewModel: MainViewModel by activityViewModels()
-    protected val twoPaneViewModel: TwoPaneViewModel by activityViewModels()
+    private val twoPaneViewModel: TwoPaneViewModel by activityViewModels()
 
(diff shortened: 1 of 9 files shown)


LunarX (changes requested):
I don't think we should have to bother creating and using thread backups. This feels like excessive maneuvering.

Can't we instead have the values we care about inside the viewmodel at all time and every view that relies on it will directly go get them there. I feel like that's how the viewmodel is meant to be used

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
```diff
@@ -133,15 +133,20 @@ abstract class TwoPaneFragment : Fragment() {
         }
     }
 
-    fun navigateToThread(thread: Thread) = with(twoPaneViewModel) {
+    fun navigateToThread(thread: Thread) {
         if (thread.isOnlyOneDraft) {
             trackNewMessageEvent(OPEN_FROM_DRAFT_NAME)
-            openDraft(thread)
+            twoPaneViewModel.openDraft(thread)
```
tevincent, line 139: The ViewModel shouldn't do UI related stuff.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/TwoPaneFragment.kt
KevinBoulongne: There no UI stuff here 🤔

tevincent (approved):

PR #1732, opened 2024-03-01
Swap `with(binding) { … }` with `with(threadViewModel) { … }`

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -350,18 +350,18 @@ class ThreadFragment : Fragment() {
...
-        threadViewModel.threadLive.observe(viewLifecycleOwner) { thread ->
+        threadLive.observe(viewLifecycleOwner) { thread ->
 
             if (thread == null) {
                 twoPaneViewModel.closeThread()
                 return@observe
             }
 
-            threadSubject.movementMethod = LinkMovementMethod.getInstance()
+            binding.threadSubject.movementMethod = LinkMovementMethod.getInstance()
```
LunarX, line 362: Shouldn't this line be called only once during the onViewCreated and not every time we observe the thread?

KevinBoulongne: Prbbly, yes.

LunarX (approved):

PR #1734, opened 2024-03-01
Fix Attachments not being uploaded when activity is destroyed by system

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/144

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -53,7 +53,6 @@ import com.infomaniak.mail.MatomoMail.trackNewMessageEvent
 import com.infomaniak.mail.R
 import com.infomaniak.mail.data.LocalSettings
 import com.infomaniak.mail.data.LocalSettings.ExternalContent
-import com.infomaniak.mail.data.models.Attachment.AttachmentDisposition.INLINE
 import com.infomaniak.mail.data.models.draft.Draft.DraftAction
 import com.infomaniak.mail.data.models.draft.Draft.DraftMode
 import com.infomaniak.mail.data.models.signature.Signature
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -470,10 +468,14 @@ class NewMessageFragment : Fragment() {
...
 
-    private fun observeNewAttachments() = with(binding) {
-        newMessageViewModel.importedAttachments.observe(viewLifecycleOwner) { (attachments, importationResult) ->
-            attachmentAdapter.addAll(attachments)
-            attachmentsRecyclerView.isGone = attachmentAdapter.itemCount == 0
+    private fun observeNewAttachments() = with(newMessageViewModel) {
+        importedAttachments.observe(viewLifecycleOwner) { (newAttachments, importationResult) ->
+            attachmentAdapter.addAll(newAttachments)
+            newAttachments.filter { newAttachment ->
+                draft.attachments.none { it.uploadLocalUri == newAttachment.uploadLocalUri }
```
tevincent, line 475: Maybe we can do the filter in the `viewModel`. We're using only using `draft`, which is in the `viewModel` to filter anyway. 

That way, the view only need to display the value and all the processing is in the `viewModel`.

Also, it seems that `ImportationResult.SUCCESS` is never used.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
FabianDevel: The check of Success value is unnecessary, because in both result value, we want to add the attachment anyway

tevincent: Yes. It's weird to have a status we return and we never use it anywhere.

FabianDevel:
This lead to a full revamp of the logic to display draft in #1770 (and prior PRs)

PR #1735, opened 2024-03-04
UI now correctly updates when marking Thread as seen

KevinBoulongne:
Depends on #1733

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
@@ -34,6 +34,7 @@ import androidx.annotation.DrawableRes
 import androidx.core.view.isGone
 import androidx.core.view.isInvisible
 import androidx.core.view.isVisible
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
```diff
@@ -556,7 +554,9 @@ class ThreadListAdapter @Inject constructor(
     override fun createDiffUtil(oldList: List<Any>, newList: List<Any>): DragDropSwipeDiffCallback<Any>? = null
 
     override fun updateList(itemList: List<Thread>) {
-        globalCoroutineScope.launch {
+
+        formatListJob?.cancel()
+        formatListJob = globalCoroutineScope.launch {
```
LunarX, line 559: I think we might need an ensureActive() for it to work

LunarX (changes requested):

FabianDevel (approved):

LunarX (approved):

PR #1737, opened 2024-03-04
Change the way we show Notifications to avoid notif sound for every notif received

tevincent:


app/src/main/java/com/infomaniak/mail/utils/NotificationUtils.kt
```diff
@@ -257,13 +291,13 @@ class NotificationUtils @Inject constructor(
     }
 
     companion object : NotificationUtilsCore() {
+        private const val DEFAULT_SMALL_ICON = R.drawable.ic_logo_notification
+        private const val DELAY_DEBOUNCE_NOTIF_MS = 500L
 
         private val TAG: String = NotificationUtils::class.java.simpleName
```
KevinBoulongne, line 297:
```suggestion

        private val TAG: String = NotificationUtils::class.java.simpleName

        private const val DEFAULT_SMALL_ICON = R.drawable.ic_logo_notification
        private const val DELAY_DEBOUNCE_NOTIF_MS = 500L
```

KevinBoulongne (approved):

PR #1749, opened 2024-03-08
Various cleaning in NewMessage

tevincent:

PR #1750, opened 2024-03-08
Fix wrong selected mailbox

tevincent:
depends on https://github.com/Infomaniak/android-kMail/pull/1749

[select_mailbox.webm](https://github.com/Infomaniak/android-kMail/assets/149579879/dc4036d6-5bba-4b76-83b0-c57a9fcbbf73)

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/MailboxesAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/MailboxesAdapter.kt
@@ -1,6 +1,6 @@
 /*
  * Infomaniak Mail - Android
- * Copyright (C) 2022-2023 Infomaniak Network SA
+ * Copyright (C) 2022-2024 Infomaniak Network SA
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/menu/MailboxesAdapter.kt
```diff
@@ -92,12 +108,18 @@ class MailboxesAdapter(
...
-                trackerCallback()
-                onValidMailboxClicked?.invoke(mailbox.mailboxId)
+        fun getClickListener(isCurrentMailbox: Boolean): OnClickListener? {
+            return if (!isCurrentMailbox) {
+                OnClickListener {
+                    trackerCallback()
+                    onValidMailboxClicked?.invoke(mailbox.mailboxId)
+                }
+            } else {
+                null
```
LunarX, line 118: If you're gonna make it an inner function, you don't need to pass in the `isCurrentMailbox` argument, it's already accessible inside `getClickListener()`

LunarX (changes requested):
Why does this PR depend on #1749? They seem pretty different

app/src/main/java/com/infomaniak/mail/ui/main/menu/MailboxesAdapter.kt
LunarX, line 1: While you're at it, you could update the file copyright date, I just saw this by chance

app/src/main/java/com/infomaniak/mail/ui/main/menu/MailboxesAdapter.kt
LunarX: I actually especially like it there cause you save yourself from passing 3 different arguments to the function

LunarX (changes requested):

LunarX (approved):

PR #1764, opened 2024-03-15
Save Draft in Realm when leaving NewMessage screen

KevinBoulongne:
Depends on #1762

PR #1765, opened 2024-03-15
Observe AI result at the right moment

KevinBoulongne:
Depends on #1764


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -287,6 +286,7 @@ class NewMessageFragment : Fragment() {
             editorManager.setupEditorFormatActionsToggle()
 
             externalsManager.observeExternals(newMessageViewModel.arrivedFromExistingDraft())
+            aiManager.observeEverything()
```
LunarX, line 289: Do `observeAiPromptStatus()` and `observeAiFeatureFlagUpdates()` really need to be moved in this later part?

I guess you moved `observeAiOutput()` here so it would only update the body of the mail after the first snapshot has been recorded right?

LunarX (changes requested):

PR #1768, opened 2024-03-18
Hide Chevron until Draft loading is finished

KevinBoulongne:
Depends on #1767


app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
```diff
@@ -166,6 +166,8 @@ class RecipientFieldView @JvmOverloads constructor(
 
     fun hideLoader() = with(binding) {
         textInput.isVisible = true
+        chevronLoaderVisibilityLayout.isVisible = true
```
tevincent, line 169: No need to call that layout `visibilitySomething` right?

tevincent (changes requested):

tevincent (approved):

FabianDevel (changes requested):
CI doesn't build because of fDroid

FabianDevel (approved):

PR #1783, opened 2024-03-26
D1: Prepare to use LiveData instead of Draft model for UI

KevinBoulongne:
Depends on #1782

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
@@ -188,8 +188,8 @@ class NewMessageViewModel @Inject constructor(
             markAsRead(currentMailbox, realm)
 
             realm.writeBlocking { draftController.upsertDraft(draft, realm = this) }
-            draft.saveDraftSnapshot()
-            draft.moveDataFromDraftToLiveData()
+            draft.saveSnapshot()
+            draft.initLiveDataFromDraft()
 
             draftInRAM = draft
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -158,53 +159,50 @@ class NewMessageViewModel @Inject constructor(
...
-            true
+            return@runCatching newDraft
         }.getOrElse {
             Sentry.captureException(it)
-            false
+            return@getOrElse null
         }
 
-        if (isSuccess) {
+        if (draft != null) {
```
tevincent, line 185: `draft?.let`?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -158,53 +159,50 @@ class NewMessageViewModel @Inject constructor(
...
-                otherFieldsAreAllEmpty.postValue(false)
-                initializeFieldsAsOpen.postValue(true)
-            }
+            draftInRAM = draft
 
-            initResult.postValue(draftInRAM to signatures)
+            initResult.postValue(InitResult(draft, signatures))
         }
 
+        val isSuccess = draft != null
```
tevincent, line 199: `emit(draft != null)`?

tevincent (changes requested):

LunarX (changes requested):
Also, what happens when the draft remote id changes compared to the one we keep in the UI

tevincent (approved):

FabianDevel (changes requested):

FabianDevel (changes requested):
We shpould save the draftRessource in savedStateHandle, because if the user put the app in background before we put the fetched draft in realm, and the system kill the app, then we won't be able to fetch the draft anymore as the resource will be null

PR #1785, opened 2024-03-27
D2: `subject` field now uses its own LiveData

KevinBoulongne:
Depends on #1783

PR #1786, opened 2024-03-27
D3: `body` field now uses its own LiveData

KevinBoulongne:
Depends on #1785


tevincent (approved):

tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -618,15 +615,16 @@ class NewMessageViewModel @Inject constructor(
         isFinishing: Boolean,
         isTaskRoot: Boolean,
         rawSubject: String,
+        rawUiBody: String,
```
FabianDevel, line 618: calling it raw is sort of implying that it will have some computing on the value, which is not the case here, so i'm not a big fan

FabianDevel (approved):

PR #1789, opened 2024-03-27
D6: `from` field now uses its own LiveData

KevinBoulongne:
Depends on #1788

PR #1790, opened 2024-03-27
D7: `attachments` field now uses its own LiveData

KevinBoulongne:
Depends on #1789

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
+++ b/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
@@ -100,6 +100,10 @@ object MatomoMail : MatomoCore {
     }
 
     fun Fragment.trackAttachmentActionsEvent(name: String) {
+        context?.trackAttachmentActionsEvent(name)
+    }
+
+    fun Context.trackAttachmentActionsEvent(name: String) {
         trackEvent("attachmentActions", name)
     }
 
--- a/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
+++ b/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
@@ -100,10 +100,6 @@ object MatomoMail : MatomoCore {
     }
 
     fun Fragment.trackAttachmentActionsEvent(name: String) {
-        context?.trackAttachmentActionsEvent(name)
-    }
-
-    fun Context.trackAttachmentActionsEvent(name: String) {
         trackEvent("attachmentActions", name)
     }
 
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/AttachmentAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/AttachmentAdapter.kt
@@ -83,8 +83,7 @@ class AttachmentAdapter(
     }
 
     fun submitList(newList: List<Attachment>) {
-        attachments.clear()
-        attachments.addAll(newList.filterNot { it.disposition == AttachmentDisposition.INLINE })
+        setAttachments(newList.filterNot { it.disposition == AttachmentDisposition.INLINE })
         notifyDataSetChanged()
     }
 
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/AttachmentAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/AttachmentAdapter.kt
@@ -31,12 +31,12 @@ import com.infomaniak.mail.utils.Utils.runCatchingRealm
 
 class AttachmentAdapter(
     private val shouldDisplayCloseButton: Boolean = false,
-    private val onDelete: ((position: Int, itemCountLeft: Int) -> Unit)? = null,
+    private val onDelete: ((position: Int) -> Unit)? = null,
     private val onAttachmentClicked: ((Attachment) -> Unit)? = null,
     private val onAttachmentOptionsClicked: ((Attachment) -> Unit)? = null,
 ) : Adapter<AttachmentViewHolder>() {
 
-    private var attachments: MutableList<Attachment> = mutableListOf()
+    private val attachments: MutableList<Attachment> = mutableListOf()
 
     override fun onCreateViewHolder(parent: ViewGroup, viewType: Int): AttachmentViewHolder {
         return AttachmentViewHolder(ItemAttachmentBinding.inflate(LayoutInflater.from(parent.context), parent, false))
@@ -68,7 +68,7 @@ class AttachmentAdapter(
                         attachments.removeAt(index)
                         notifyItemRemoved(index)
 
-                        onDelete?.invoke(index, itemCount)
+                        onDelete?.invoke(index)
                     }
                 }
             }
@@ -77,11 +77,15 @@ class AttachmentAdapter(
 
     override fun getItemCount(): Int = runCatchingRealm { attachments.count() }.getOrDefault(0)
 
-    fun setAttachments(newList: List<Attachment>) = runCatchingRealm { attachments = newList.toMutableList() }
+    fun setAttachments(newList: List<Attachment>) = runCatchingRealm {
+        attachments.clear()
+        attachments.addAll(newList)
+    }
 
-    fun addAll(newAttachments: List<Attachment>) {
-        attachments.addAll(newAttachments.filterNot { it.disposition == AttachmentDisposition.INLINE })
-        notifyItemRangeInserted(attachments.lastIndex, newAttachments.count())
+    fun submitList(newList: List<Attachment>) {
+        attachments.clear()
+        attachments.addAll(newList.filterNot { it.disposition == AttachmentDisposition.INLINE })
+        notifyDataSetChanged()
     }
 
     private fun ItemAttachmentBinding.toggleEndIconVisibility(shouldDisplayCloseButton: Boolean) {
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -436,12 +436,10 @@ class NewMessageFragment : Fragment() {
 
             if (shouldRegisterToImportedAttachments) observeImportAttachments()
 
-            val shouldTransition = attachmentAdapter.itemCount != 0 && it.isEmpty()
-
             attachmentAdapter.submitList(it)
 
             binding.attachmentsRecyclerView.isVisible = if (it.isEmpty()) {
-                if (shouldTransition) TransitionManager.beginDelayedTransition(binding.root)
+                if (attachmentAdapter.itemCount != 0) TransitionManager.beginDelayedTransition(binding.root)
                 false
             } else {
                 true
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -436,12 +436,8 @@ class NewMessageFragment : Fragment() {
 
             attachmentAdapter.submitList(attachments)
 
-            binding.attachmentsRecyclerView.isVisible = if (attachments.isEmpty()) {
-                if (attachmentAdapter.itemCount != 0) TransitionManager.beginDelayedTransition(binding.root)
-                false
-            } else {
-                true
-            }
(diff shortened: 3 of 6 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -431,16 +427,27 @@ class NewMessageFragment : Fragment() {
...
+            binding.attachmentsRecyclerView.isVisible = if (it.isEmpty()) {
+                if (shouldTransition) TransitionManager.beginDelayedTransition(binding.root)
+                false
+            } else {
+                true
+            }
 
-            if (importationResult == ImportationResult.FILE_SIZE_TOO_BIG) showSnackbar(R.string.attachmentFileLimitReached)
-            updateIsSendingAllowed()
+            newMessageViewModel.updateIsSendingAllowed(attachments = it)
```
tevincent, line 444: This can be in the viewModel right? It's a bit weird to go back and forth between the fragment and viewModel.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -499,26 +506,21 @@ class NewMessageFragment : Fragment() {
...
         runCatching {
-            val attachment = draftInRAM.attachments[position]
+            val attachments = attachmentsLiveData.value?.toMutableList() ?: mutableListOf()
+            val attachment = attachments[position]
             attachment.getUploadLocalFile()?.delete()
-            LocalStorageUtils.deleteAttachmentUploadDir(requireContext(), draftInRAM.localUuid, attachment.localUuid)
-            draftInRAM.attachments.removeAt(position)
+            LocalStorageUtils.deleteAttachmentUploadDir(requireContext(), draftLocalUuid()!!, attachment.localUuid)
+            attachments.removeAt(position)
+            attachmentsLiveData.value = attachments
```
tevincent, line 519: You can move that to the viewModel.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -427,16 +424,27 @@ class NewMessageFragment : Fragment() {
         }
     }
 
-    private fun observeNewAttachments() = with(newMessageViewModel) {
-        importedAttachments.observe(viewLifecycleOwner) { (attachments, importationResult) ->
+    private fun observeAttachments() {
+        newMessageViewModel.attachmentsLiveData.observe(viewLifecycleOwner) {
+
+            val shouldTransition = attachmentAdapter.itemCount != 0 && it.isEmpty()
```
tevincent, line 430: You can move that to line to line 435 like so:

```kotlin
binding.attachmentsRecyclerView.isVisible = if (it.isEmpty()) {
   val shouldTransition = attachmentAdapter.itemCount != 0 && it.isEmpty()
   if (shouldTransition) TransitionManager.beginDelayedTransition(binding.root)
         false
      } else {
         true
   }
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -531,33 +535,35 @@ class NewMessageViewModel @Inject constructor(
         }
     }
 
-    private fun importAttachments(uris: List<Uri>, draft: Draft) = viewModelScope.launch(ioCoroutineContext) {
+    fun importAttachments(currentAttachments: List<Attachment>, uris: List<Uri>): List<Attachment> {
```
FabianDevel, line 538: we shouldn't write to the disk on the main thread

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -531,33 +535,35 @@ class NewMessageViewModel @Inject constructor(
...
     }
 
-    private fun importAttachment(localUuid: String, uri: Uri, availableSpace: Long): Pair<Attachment?, Boolean>? {
+    private fun importAttachment(uri: Uri, availableSpace: Long): Pair<Attachment?, Boolean>? {
 
         val (fileName, fileSize) = context.getFileNameAndSize(uri) ?: return null
         val attachment = Attachment()
 
-        return LocalStorageUtils.saveAttachmentToUploadDir(context, uri, fileName, localUuid, attachment.localUuid)
+        return LocalStorageUtils.saveAttachmentToUploadDir(context, uri, fileName, draftLocalUuid!!, attachment.localUuid)
```
FabianDevel, line 566: this can be called before initLiveData, so there is a risk that draftLocalUuid is null

FabianDevel: we should probably keep this ref to localUuid from the draft (or call the initLiveData sooner)

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -427,16 +426,38 @@ class NewMessageFragment : Fragment() {
...
+            if (shouldRegisterToImportedAttachments) observeImportAttachments()
+
+            attachmentAdapter.submitList(it)
+
+            binding.attachmentsRecyclerView.isVisible = if (it.isEmpty()) {
+                if (attachmentAdapter.itemCount != 0) TransitionManager.beginDelayedTransition(binding.root)
+                false
+            } else {
+                true
+            }
```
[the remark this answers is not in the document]
KevinBoulongne: They need to be done together.
The transition is about the fact that the view is going to be gone.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -488,26 +509,8 @@ class NewMessageFragment : Fragment() {
         draftsActionsWorkerScheduler.scheduleWork(newMessageViewModel.draftLocalUuid())
     }
 
-    private fun onDeleteAttachment(position: Int, itemCountLeft: Int) = with(newMessageViewModel) {
-
-        trackAttachmentActionsEvent("delete")
-
-        if (itemCountLeft == 0) {
-            TransitionManager.beginDelayedTransition(binding.root)
```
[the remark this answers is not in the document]
KevinBoulongne: Not the same itemCounts.
The old `itemCountLeft` corresponds to the new `attachments.isEmpty()`.
This logic hasn't changed.

Before, we were only acting when deleting an Attachment (the code you selected was in `fun onDeleteAttachment()`.
Now, we are acting when adding & deleting. Hence, we need to check if we are going from a list to an empty list to trigger the transition.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
KevinBoulongne: This is actually really called before `initLiveData`, so there is currently a bug.
Nice catch !

app/src/main/java/com/infomaniak/mail/ui/main/thread/AttachmentAdapter.kt
```diff
@@ -77,11 +77,15 @@ class AttachmentAdapter(
...
+        attachments.clear()
+        attachments.addAll(newList)
+    }
 
-    fun addAll(newAttachments: List<Attachment>) {
-        attachments.addAll(newAttachments.filterNot { it.disposition == AttachmentDisposition.INLINE })
-        notifyItemRangeInserted(attachments.lastIndex, newAttachments.count())
+    fun submitList(newList: List<Attachment>) {
+        attachments.clear()
+        attachments.addAll(newList.filterNot { it.disposition == AttachmentDisposition.INLINE })
```
FabianDevel, line 87: You could use setAttachments here

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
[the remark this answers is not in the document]
FabianDevel: Alright, but we could still separate the two treatment.
In cases like this we usually set the visibility directly with the condition, then make a if on the condition tu do other treatments

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
```diff
@@ -100,6 +100,10 @@ object MatomoMail : MatomoCore {
     }
 
     fun Fragment.trackAttachmentActionsEvent(name: String) {
+        context?.trackAttachmentActionsEvent(name)
+    }
+
+    fun Context.trackAttachmentActionsEvent(name: String) {
```
FabianDevel, line 106: this is not needed anymore as you kept the tracker in the fragment

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -432,16 +429,39 @@ class NewMessageFragment : Fragment() {
...
-        importedAttachments.observe(viewLifecycleOwner) { (attachments, importationResult) ->
+    private fun observeAttachments() {
+        newMessageViewModel.attachmentsLiveData.observe(viewLifecycleOwner) { attachments ->
+
+            if (shouldRegisterToImportedAttachments) observeImportAttachments()
+
+            attachmentAdapter.submitList(attachments)
+
+            binding.attachmentsRecyclerView.isVisible = if (attachments.isEmpty()) {
+                if (attachmentAdapter.itemCount != 0) TransitionManager.beginDelayedTransition(binding.root)
```
FabianDevel, line 440: I don't really get the use of this check, as you submit the adapter with "attachments" before. So if 'attachments' is empty, this will be == 0 no ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -530,33 +530,46 @@ class NewMessageViewModel @Inject constructor(
...
+        uris: List<Uri>,
+        completion: (List<Attachment>) -> Unit,
+    ) = viewModelScope.launch(ioCoroutineContext) {
+        completion(importAttachments(currentAttachments, uris))
+    }
+
+    private fun importAttachments(
+        currentAttachments: List<Attachment>,
+        uris: List<Uri>,
+    ): List<Attachment> {
```
FabianDevel, line 544: There should be enough space to oneline this

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -603,11 +616,27 @@ class NewMessageViewModel @Inject constructor(
...
+            attachmentsLiveData.value = attachments
+        }.onFailure { exception ->
+            // TODO: If we don't see this Sentry after June 2024, we can remove it.
+            SentryLog.e(TAG, " Attachment $position doesn't exist", exception)
+        }
+    }
+
+    fun updateIsSendingAllowed(
+        attachments: List<Attachment> = attachmentsLiveData.valueOrEmpty(),
+    ) {
```
FabianDevel, line 635: oneline this

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
KevinBoulongne: This code is wrong, here is the correct version :
```kt
private fun observeAttachments() = with(newMessageViewModel) {
    attachmentsLiveData.observe(viewLifecycleOwner) { attachments ->

        if (shouldRegisterToImportedAttachments) observeImportAttachments()

        val shouldTransition = attachmentAdapter.itemCount != 0 && attachments.isEmpty()
        attachmentAdapter.submitList(attachments)
        if (shouldTransition) TransitionManager.beginDelayedTransition(binding.root)
        binding.attachmentsRecyclerView.isVisible = attachments.isNotEmpty()

        updateIsSendingAllowed(attachments)
    }
}
```

FabianDevel (approved):

tevincent (approved):

PR #1796, opened 2024-04-02
D9: Use a single Realm transaction to get & update Draft when leaving

KevinBoulongne:
Depends on #1791


tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -679,33 +690,48 @@ class NewMessageViewModel @Inject constructor(
...
 
-        val draft = getLatestLocalDraft(draftLocalUuid) ?: return@launch
+        val localUuid = draftLocalUuid ?: return@launch
         val subject = subjectValue.ifBlank { null }?.take(SUBJECT_MAX_LENGTH)
 
-        draft.updateDraftFromLiveData(action, subject, uiBodyValue)
-
-        if (isFinishing && isSavingDraftWithoutChanges(draft, action, subject, uiBodyValue)) {
-            if (!arrivedFromExistingDraft) removeDraftFromRealm(draft.localUuid)
+        if (isFinishing && isSavingDraftWithoutChanges(action, subject, uiBodyValue)) {
```
LunarX, line 696: Why not try to skip the draft saving if there's no change anyway when placing the app in the background?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -679,33 +690,48 @@ class NewMessageViewModel @Inject constructor(
...
+            if (!arrivedFromExistingDraft) removeDraftFromRealm(localUuid)
             return@launch
         }
 
-        context.trackSendingDraftEvent(action, draft, currentMailbox.externalMailFlagEnabled)
+        val hasFailed = mailboxContentRealm().writeBlocking {
+            DraftController.getDraft(localUuid, realm = this)
+                ?.apply { updateDraftFromLiveData(action, subject, uiBodyValue, realm = this@writeBlocking) }
```
LunarX, line 703: shouldn't we have used a `let` actually?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -679,33 +690,48 @@ class NewMessageViewModel @Inject constructor(
...
+            return@writeBlocking false
+        }
 
-        saveDraftToLocal(draft)
+        if (hasFailed) return@launch
 
         if (isFinishing) {
-            if (isTaskRoot) showDraftToastToUser(action)
             startWorkerCallback()
+            if (isTaskRoot) showDraftToastToUser(action)
```
LunarX, line 712: Why move this line around?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
KevinBoulongne: If we do that with "Don't keep activities" activated, we're gonna loose the Draft.

LunarX: If the draft hasn't been modified (i.e. `isSavingDraftWithoutChanges`) there should not be anything to loose in all logic

FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
KevinBoulongne: `isSavingDraftWithoutChanges ` doesn't mean there is nothing.
It only means that the Draft hasn't been modified by the user, but there is still a whole Draft that exists.

It will be lost, and we'll need to create it again when resuming the app.

LunarX: we agreed there shouldn't be a need to save it

PR #1804, opened 2024-04-08
Fix a crash when an Email's body in a Notification had too many characters

LunarX:
Issue can be seen in sentry issues like the id 41593

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
@@ -251,8 +251,8 @@ class FetchMessagesManager @Inject constructor(
         private val TAG: String = FetchMessagesManager::class.java.simpleName
 
         // Based on what seems to be the limit taken into account by NotificationCompat.Builder, we truncate some fields to avoid
-        // pendingIntent that will trigger TransactionTooLargeException further down the line like sentry id 41593. We don't need
-        // the extra data, it was bound to be truncated anyway
-        private const val MAX_CHAR_LIMIT = 5 * 1024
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -249,5 +249,10 @@ class FetchMessagesManager @Inject constructor(
 
     companion object {
         private val TAG: String = FetchMessagesManager::class.java.simpleName
+
+        // Based on what seems to be the limit taken into account by NotificationCompat.Builder, we truncate some fields to avoid
+        // pendingIntent that will trigger TransactionTooLargeException further down the line like sentry id 41593. We don't need
+        // the extra data, it was bound to be truncated anyway
```
KevinBoulongne, line 255:
```suggestion
        // Based on what seems to be the limit taken into account by NotificationCompat.Builder, we truncate some fields to avoid
        // pendingIntent that will trigger TransactionTooLargeException further down the line like Sentry ID 41593. We don't need
        // the extra data, it was bound to be truncated anyway.
```

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -249,5 +249,10 @@ class FetchMessagesManager @Inject constructor(
...
+        private const val MAX_CHAR_LIMIT = 5 * 1024
```
KevinBoulongne, line 256:
```suggestion
        private const val MAX_CHAR_LIMIT = 5 * 1_024
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1806, opened 2024-04-11
Bump sentry

JorisBodin:
Depends on https://github.com/Infomaniak/android-core/pull/163


LunarX (changes requested):
Is it normal that this PR makes it so the Core submodules targets a branch which is not master?

LunarX (approved):

PR #1826, opened 2024-04-23
Draft revamp

KevinBoulongne:


FabianDevel (changes requested):
There's lack the animation when suppressing attachment
When adding an attachment we have a toast botbh when opening the file picker and saving the attachment, so it displays 2 times in a row most of the time
Sometimes there's 2 saving drafts in a row with the AI

FabianDevel (approved):
🥳

PR #1831, opened 2024-04-29
Dismiss the group notification if we dismissed the last notification

tevincent:


app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -139,6 +139,12 @@ class FetchMessagesManager @Inject constructor(
             }
         }
 
+        if (notificationManagerCompat.activeNotifications.isEmpty()) {
+            // We don't need to display a group notification saying "0 new messages" so if we don't have any active notifications,
+            // We dismiss the group notification.
+            notificationManagerCompat.cancel(mailbox.notificationGroupId)
+        }
```
KevinBoulongne, line 146:
```suggestion
        // We don't need to display a group notification saying "0 new messages" so
        // if we don't have any active notifications, we dismiss the group notification.
        if (notificationManagerCompat.activeNotifications.isEmpty()) {
            notificationManagerCompat.cancel(mailbox.notificationGroupId)
        }
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1832, opened 2024-04-29
Only save Draft when adding Attachment

KevinBoulongne:
We were saving Draft everytime we add or remove an Attachment, it's too much.
Now, we only save Draft when adding an Attachment.

Also, if we add a lot of Attachments at the same time _(via the FilePicker)_, and we delete them all while it's still saving, the DraftsActionsWorker will fail because it's won't find the files on the phone disk.
Now, it will just ignore these failures so the Draft can be saved.


FabianDevel (changes requested):
This display a toast (should be a snackbar I think as we stay in the app) when coming b ack to the activity after saving an attachment
This is annoying as we are already displaying a toast when quitting the activity to open the picker
We probably shouldn't display anything when coming back to the activity

FabianDevel (approved):

PR #1836, opened 2024-04-30
Remove SignatureAdapter memory leak

KevinBoulongne:
Depends on #1834 

Remove strong reference about SignatureAdapter in NewMessageFragment.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -103,7 +103,7 @@ class NewMessageFragment : Fragment() {
     private var quoteWebView: WebView? = null
     private var signatureWebView: WebView? = null
 
-    private val signatureAdapter = SignatureAdapter(::onSignatureClicked)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -103,7 +103,7 @@ class NewMessageFragment : Fragment() {
     private var quoteWebView: WebView? = null
     private var signatureWebView: WebView? = null
 
-    private val signatureAdapter = SignatureAdapter(::onSignatureClicked)
+    private var signatureAdapter: SignatureAdapter? = SignatureAdapter(::onSignatureClicked)
```
LunarX, line 106: If the value is made null inside onDestroyView maybe we want to instantiate it inside onCreateView

LunarX (changes requested):

FabianDevel (approved):

LunarX (commented):
Are we sure the adapter for a list actually leaks in the first place? Because maybe only recyclerview's ones do. Maybe list adapters don't have a strong reference to their ListPopupWindow

PR #1838, opened 2024-05-01
Save Draft in Activity `onStop` instead of Fragment one

KevinBoulongne:
Depends on #1832

Move the Draft saving from the Fragment's `onStop` to the Activity's `onStop` instead.
It will lessen the Toast spam of _"Saving draft"_.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -39,6 +39,7 @@ import androidx.appcompat.content.res.AppCompatResources
 import androidx.constraintlayout.widget.Group
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
+import androidx.core.widget.doAfterTextChanged
 import androidx.fragment.app.Fragment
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -415,6 +415,18 @@ class NewMessageFragment : Fragment() {
...
+            editable?.toString()?.let { newMessageViewModel.subjectTextField = it }
+        }
+    }
+
+    private fun doAfterBodyChange() {
+        binding.bodyTextField.doAfterTextChanged { editable ->
+            editable?.toString()?.let { newMessageViewModel.bodyTextField = it }
+        }
+    }
+
```
LunarX, line 429: Can you access the content when saving instead of saving it continuously?

KevinBoulongne: No, because the Activity's `onStop()` always triggers before the Fragment's `onStop()`.
It's the one from the Activity that tells the Fragment to start its own.

PR #1839, opened 2024-05-02
Save Draft in Activity `onStop` instead of Fragment one

KevinBoulongne:
Move the Draft saving from the Fragment's `onStop` to the Activity's `onStop` instead.
It will lessen the Toast spam of _"Saving draft"_.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -72,7 +72,6 @@ import com.infomaniak.mail.utils.WebViewUtils
 import com.infomaniak.mail.utils.WebViewUtils.Companion.destroyAndClearHistory
 import com.infomaniak.mail.utils.WebViewUtils.Companion.setupNewMessageWebViewSettings
 import com.infomaniak.mail.utils.extensions.*
-import com.infomaniak.mail.workers.DraftsActionsWorker
 import dagger.hilt.android.AndroidEntryPoint
 import io.sentry.Sentry
 import io.sentry.SentryLevel
@@ -82,7 +81,7 @@ import javax.inject.Inject
 class NewMessageFragment : Fragment() {
 
     private var _binding: FragmentNewMessageBinding? = null
-    private val binding get() = _binding!! // This property is only valid between onCreateView and onDestroyView
+    val binding get() = _binding!! // This property is only valid between onCreateView and onDestroyView
     private val newMessageActivityArgs by lazy {
         // When opening this fragment via deeplink, it can happen that the navigation
         // extras aren't yet initialized, so we don't use the `navArgs` here.
@@ -123,9 +122,6 @@ class NewMessageFragment : Fragment() {
 
     @Inject
     lateinit var localSettings: LocalSettings
-
-    @Inject
-    lateinit var draftsActionsWorkerScheduler: DraftsActionsWorker.Scheduler
 
     @Inject
     lateinit var informationDialog: InformationAlertDialog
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -82,7 +81,7 @@ import javax.inject.Inject
 class NewMessageFragment : Fragment() {
 
     private var _binding: FragmentNewMessageBinding? = null
-    private val binding get() = _binding!! // This property is only valid between onCreateView and onDestroyView
+    val binding get() = _binding!! // This property is only valid between onCreateView and onDestroyView
```
LunarX, line 84: I think we should avoid exposing the binding. Instead we can provide a method that returns the final data that we need to expose

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -527,24 +524,12 @@ class NewMessageFragment : Fragment() {
...
-            action = if (shouldSendInsteadOfSave) DraftAction.SEND else DraftAction.SAVE,
-            isFinishing = requireActivity().isFinishing,
-            isTaskRoot = requireActivity().isTaskRoot,
-            subjectValue = binding.subjectTextField.text.toString(),
-            uiBodyValue = binding.bodyTextField.text.toString(),
-            startWorkerCallback = ::startWorker,
-        )
-    }
+        newMessageViewModel.subjectTextField = subjectTextField.text.toString()
+        newMessageViewModel.bodyTextField = bodyTextField.text.toString()
```
LunarX, line 530: Maybe we should add a comment to understand the context of these two lines of codes?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -106,6 +106,9 @@ class NewMessageViewModel @Inject constructor(
     val uiQuoteLiveData = MutableLiveData<String?>()
     //endregion
 
+    var subjectTextField = ""
+    var bodyTextField = ""
```
LunarX, line 110: Maybe we should even make these two variables names clearer. The way it is right now, someone might think at first that these will always contain the correct subject and body which is not the case

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: lastOnStop...?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -527,24 +524,19 @@ class NewMessageFragment : Fragment() {
...
-            subjectValue = binding.subjectTextField.text.toString(),
-            uiBodyValue = binding.bodyTextField.text.toString(),
-            startWorkerCallback = ::startWorker,
-        )
-    }
+        /**
+         * When the Activity is being stopped, we save the Draft.
+         * We then need the up-to-date subject & body values.
+         * If we are in the NewMessageFragment when stopping, it's easy, we just get them from the binding.
+         * If we are not (ex: FilePicker, IA…), we get them from the ViewModel.
```
LunarX, line 533: FilePicker is another Activity which means that from the point of view of "NewMessageActivity", the NewMessageFragment is still there and reachable

LunarX (changes requested):

LunarX (approved):

PR #1840, opened 2024-05-02
Use custom FormatterFileSize

KevinBoulongne:
Depends on https://github.com/Infomaniak/android-core/pull/173

Since Android O, Google decided to use SI format (1000) instead of binary one (1024) to display bytes.

Source : https://stackoverflow.com/a/3758880

Unit tests are here : https://github.com/Infomaniak/android-core/pull/175


tevincent (changes requested):
Maybe we can add some tests to check the formatter output.

tevincent (approved):

PR #1844, opened 2024-05-06
Update Signature API and handle nullable signature

tevincent:
Update signature API to take into account the fact that we can have no default signature.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
@@ -98,9 +98,11 @@ object ApiRepository : ApiRepositoryCore() {
         return callApi(ApiRoutes.signatures(mailboxHostingId, mailboxName), GET)
     }
 
-    fun setDefaultSignature(mailboxHostingId: Int, mailboxName: String, signature: Signature): ApiResponse<Boolean> {
-        val body = mapOf("default_signature_id" to signature.id)
-        return callApi(ApiRoutes.signature(mailboxHostingId, mailboxName), PUT, body)
+    fun setDefaultSignature(mailboxHostingId: Int, mailboxName: String, signature: Signature?): ApiResponse<Boolean> {
+        // If signature is null, it means we want to have no default signature.
+        // If we want to delete the default signature, we have to pass null to the WS call.
+        val body = """{"default_signature_id":${signature?.id}}"""
+        return callApi(ApiRoutes.signature(mailboxHostingId, mailboxName), POST, body)
     }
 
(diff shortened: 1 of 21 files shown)


KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -98,7 +99,8 @@ object ApiRepository : ApiRepositoryCore() {
     }
 
     fun setDefaultSignature(mailboxHostingId: Int, mailboxName: String, signature: Signature): ApiResponse<Boolean> {
```
KevinBoulongne, line 101: We should add the possibility to set the Signature to "no signature" in the Signature settings.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -98,7 +99,8 @@ object ApiRepository : ApiRepositoryCore() {
...
-        return callApi(ApiRoutes.signature(mailboxHostingId, mailboxName, signature.id), PUT, Json.encodeToString(signature))
+        val body = mapOf("default_signature_id" to signature.id, "default_reply_signature_id" to null)
```
KevinBoulongne, line 102:
```suggestion
        val body = mapOf("default_signature_id" to signature.id)
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -242,13 +246,16 @@ class NewMessageViewModel @Inject constructor(
...
+        val defaultSignature = SignatureController.getDefaultSignature(realm, draftMode)
         val shouldPreselectSignature = draftMode == DraftMode.REPLY || draftMode == DraftMode.REPLY_ALL
         val signature = if (shouldPreselectSignature) {
-            guessMostFittingSignature(previousMessage!!, signatures)
+            defaultSignature ?: guessMostFittingSignature(previousMessage!!, signatures)
         } else {
-            SignatureController.getSignature(realm)
+            defaultSignature
+        }
+        (signature ?: SignatureController.getDefaultSignatureWithFallback(realm, draftMode))?.let {
```
KevinBoulongne, line 256: There seems to be an issue somewhere around here.
When the default signature is set to "none", the UI fallback to another one.

KevinBoulongne (changes requested):

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1847, opened 2024-05-08
Redirect to webmail signatures settings

tevincent:
https://github.com/Infomaniak/android-kMail/assets/149579879/6871edf7-2b90-4fd1-a25a-455430d9a87e

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -37,6 +37,7 @@ android {
         // buildConfigField 'String', 'MAIL_API_PREPROD', '"https://mail.preprod.dev.infomaniak.ch"'
         buildConfigField 'String', 'SHOP_URL', '"https://ik.me"'
         buildConfigField 'String', 'CHATBOT_URL', '"https://www.infomaniak.com/chatbot"'
+        buildConfigField 'String', 'MANAGE_SIGNATURES_URL', '"https://mail.infomaniak.com/0/settings/signatures"'
 
         buildConfigField 'String', 'BUGTRACKER_MAIL_BUCKET_ID', '"app_mail"'
         buildConfigField 'String', 'BUGTRACKER_MAIL_PROJECT_NAME', '"mail"'
--- a/app/src/main/res/layout/fragment_signature_setting.xml
+++ b/app/src/main/res/layout/fragment_signature_setting.xml
@@ -51,6 +51,7 @@
                 style="@style/TextButtonSecondary"
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
+                android:layout_marginHorizontal="@dimen/marginStandardVerySmall"
                 android:text="@string/manageSignatures" />
 
         </LinearLayout>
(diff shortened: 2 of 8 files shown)


app/src/main/res/values-it/strings.xml
```diff
@@ -300,6 +300,7 @@
         <item quantity="one">Indirizzo e-mail bloccato</item>
         <item quantity="other">Indirizzi e-mail bloccati</item>
     </plurals>
+    <string name="manageSignatures">Gestire le mie firme</string>
```
LunarX, line 303: Aks @valentinperignon about the conjugation of that sentence

app/src/main/res/layout/fragment_signature_setting.xml
```diff
@@ -45,6 +45,14 @@
                 app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
                 tools:itemCount="3"
                 tools:listitem="@layout/item_settings_signature" />
+
+            <com.google.android.material.button.MaterialButton
+                android:id="@+id/manageSignatures"
+                style="@style/TextButtonSecondary"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:text="@string/manageSignatures" />
```
LunarX, line 54: Either make the padding of the button so the actual text of the button is aligned horizontally with the list of signatures' actual text or make it so the ripple of the button is horizontally aligned with the start of the signatures' actual text but not something in between

app/build.gradle
```diff
@@ -37,6 +37,7 @@ android {
         // buildConfigField 'String', 'MAIL_API_PREPROD', '"https://mail.preprod.dev.infomaniak.ch"'
         buildConfigField 'String', 'SHOP_URL', '"https://ik.me"'
         buildConfigField 'String', 'CHATBOT_URL', '"https://www.infomaniak.com/chatbot"'
+        buildConfigField 'String', 'MANAGE_SIGNATURES_URL', '"https://mail.infomaniak.com/0/settings/signatures"'
```
LunarX, line 40: Do we create a whole variable for it? Or do we create an ApiRoute that depends on `MAIL_API`?

LunarX (changes requested):

app/build.gradle
tevincent: @Infomaniak/android what do you all think?

KevinBoulongne: We already have `'MAIL_API', '"https://mail.infomaniak.com"'`.
Re-use it, like the other routes are doing.

KevinBoulongne (changes requested):

app/build.gradle
KevinBoulongne: Keep it this way.

KevinBoulongne (approved):

app/src/main/res/values-it/strings.xml
tevincent: @valentinperignon sent me the new string.

LunarX (approved):

PR #1853, opened 2024-05-14
Don't log NetworkException in Sentry

KevinBoulongne:
We don't want to send Sentry about NetworkException.
So we make sure that when we send a Sentry about an Exception, it's not a NetworkException.


LunarX (changes requested):
Maybe we could create a Sentry.captureNonNetworkException() method so when autocompleting while capturing an exception we can think about using this new method and this would automatically filter out all the NetworkException

FabianDevel (approved):

LunarX (approved):

PR #1854, opened 2024-05-15
Fix disappearing multi-select on a single Thread when selecting the "More" action

NicolasBourdin88:
we should not close multi selection before an action is performed

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
@@ -101,10 +101,10 @@ class ThreadListMultiSelection {
                 R.id.quickActionMenu -> {
                     threadListFragment.trackMultiSelectActionEvent(OPEN_ACTION_BOTTOM_SHEET, selectedThreadsCount)
                     val direction = if (selectedThreadsCount == 1) {
-                        isMultiSelectOn = false
                         ThreadListFragmentDirections.actionThreadListFragmentToThreadActionsBottomSheetDialog(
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
```diff
@@ -101,10 +101,10 @@ class ThreadListMultiSelection {
                 R.id.quickActionMenu -> {
                     threadListFragment.trackMultiSelectActionEvent(OPEN_ACTION_BOTTOM_SHEET, selectedThreadsCount)
                     val direction = if (selectedThreadsCount == 1) {
-                        isMultiSelectOn = false
                         ThreadListFragmentDirections.actionThreadListFragmentToThreadActionsBottomSheetDialog(
                             threadUid = selectedThreadsUids.single(),
                             shouldLoadDistantResources = false,
+                            shouldCloseMultiSelection = true
```
LunarX, line 107: Maybe we could add a comma at the end of that line

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
LunarX:
```suggestion
                            shouldCloseMultiSelection = true,
```

LunarX (approved):

KevinBoulongne (changes requested):
I'd like to review this one before being merged.

KevinBoulongne (approved):

PR #1856, opened 2024-05-15
Support `mailTo` from `Intent.Send`

FabianDevel:
The Intent.Send action should normally not be used to write mailto according to the doc.
However, some apps do it, and it is supported by gmail.

So to avoid suffer from the comparison, we now support it too.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
@@ -498,7 +498,9 @@ class NewMessageViewModel @Inject constructor(
 
     private fun handleSingleSendIntent(draft: Draft, intent: Intent) = with(intent) {
 
-        if (hasExtra(Intent.EXTRA_TEXT)) {
+        if (hasExtra(Intent.EXTRA_EMAIL)) {
+            handleMailTo(draft, intent.data, intent)
+        } else if (hasExtra(Intent.EXTRA_TEXT)) {
             getStringExtra(Intent.EXTRA_SUBJECT)?.let { draft.subject = it }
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -518,8 +520,11 @@ class NewMessageViewModel @Inject constructor(
     }
 
     /**
-     * Handle `MailTo` from [Intent.ACTION_VIEW] or [Intent.ACTION_SENDTO]
-     * Get [Intent.ACTION_VIEW] data with [MailTo] and [Intent.ACTION_SENDTO] with [Intent]
+     * Handle `MailTo` from [Intent.ACTION_VIEW], [Intent.ACTION_SENDTO] or [Intent.ACTION_SEND]
+     * Get [Intent.ACTION_VIEW] data with [MailTo] and [Intent.ACTION_SENDTO] or [Intent.ACTION_SEND] with [Intent]
+     *
+     * [Intent.ACTION_SEND] shouldn't be used for mailTo as it's not meant to pass the recipient of the mail,
+     * but some apps don't follow the guideline, so we support it anyway
```
KevinBoulongne, line 527:
```suggestion
     * Handle `mailTo` from [Intent.ACTION_VIEW], [Intent.ACTION_SENDTO] or [Intent.ACTION_SEND].
     * Get [Intent.ACTION_VIEW] data with [MailTo] and [Intent.ACTION_SENDTO] or [Intent.ACTION_SEND] with [Intent].
     *
     * [Intent.ACTION_SEND] shouldn't be used for `mailTo` as it isn't meant to pass the Recipient of the mail,
     * but some apps don't follow the guidelines, so we support it anyway.
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1872, opened 2024-05-29
Increase debounce to avoid being spammed with notifications sound and vibration

tevincent:


app/src/main/java/com/infomaniak/mail/utils/NotificationUtils.kt
```diff
@@ -303,7 +303,7 @@ class NotificationUtils @Inject constructor(
 
         private val TAG: String = NotificationUtils::class.java.simpleName
 
-        private const val DELAY_DEBOUNCE_NOTIF_MS = 500L
+        private const val DELAY_DEBOUNCE_NOTIF_MS = 1500L
```
KevinBoulongne, line 306:
```suggestion
        private const val DELAY_DEBOUNCE_NOTIF_MS = 1_500L
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1876, opened 2024-05-29
Delete Thread in Realm if it's empty and we're trying to display it in ThreadsList

KevinBoulongne:
Depends on #1875

Sometimes, we display an empty Thread in the ThreadsList.

The cause isn't known yet, but in the meantime we just delete them in Realm.
It's possibly because we are out of sync, and the situation will resolve itself shortly.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -28,6 +28,7 @@ import com.infomaniak.mail.data.models.thread.Thread.ThreadFilter
 import com.infomaniak.mail.di.IoDispatcher
 import com.infomaniak.mail.utils.ErrorCode
 import com.infomaniak.mail.utils.SearchUtils.Companion.convertToSearchThreads
+import com.infomaniak.mail.utils.SentryDebug
 import io.realm.kotlin.MutableRealm
 import io.realm.kotlin.Realm
 import io.realm.kotlin.TypedRealm
@@ -214,6 +215,12 @@ class ThreadController @Inject constructor(
         private fun getThreadQuery(uid: String, realm: TypedRealm): RealmSingleQuery<Thread> {
             return realm.query<Thread>("${Thread::uid.name} == $0", uid).first()
         }
+
+        private fun getEmptyThreadsInFolderQuery(folderId: String, realm: TypedRealm): RealmQuery<Thread> {
+            val noMessages = "${Thread::messages.name}.@size == $0"
+            return FolderController.getFolder(folderId, realm)?.threads?.query(noMessages, 0)
+                ?: realm.query<Thread>("$noMessages AND ${Thread::folderId.name} == $1", 0, folderId)
+        }
         //endregion
 
         //region Get data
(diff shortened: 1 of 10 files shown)


LunarX (changes requested):
We will need to remove all of this once the bug at the root of the issue is fixed. Isn't there a way to easily find back what needs to be removed somehow?

Also, the reason for the change request: Shouldn't we still report to Sentry that this issue occurred and was detected? Maybe even with extra informations we can compute

KevinBoulongne:
> We will need to remove all of this once the bug at the root of the issue is fixed. Isn't there a way to easily find back what needs to be removed somehow?

First and foremost, we're not even sure there's a bug here.
I applied the decision that was taken in the 25th of March Monday meeting.

> Also, the reason for the change request: Shouldn't we still report to Sentry that this issue occurred and was detected? Maybe even with extra informations we can compute

👍

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
```diff
@@ -183,6 +184,15 @@ class ThreadListAdapter @Inject constructor(
...
     private fun CardviewThreadItemBinding.displayThread(thread: Thread, position: Int) {
 
+        // If we are trying to display an empty Thread, don't. Just delete it.
+        if (thread.messages.isEmpty()) {
+            // TODO: Find why we are sometimes displaying empty Threads, and fix it instead of just deleting them.
+            //  It's possibly because we are out of sync, and the situation will resolve itself shortly? ¯\_(ツ)_/¯
+            threadListAdapterCallback?.deleteThreadInRealm?.invoke(thread.uid)
+            SentryDebug.sendEmptyThread(thread, "No Message in the Thread when displaying it in ThreadList")
+            return
+        }
```
tevincent, line 194: Maybe we can move that at the end of the `RefreshController` ? Because otherwise, we can still have `Threads`with 0 message.

tevincent (changes requested):

tevincent (approved):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -307,7 +320,22 @@ class ThreadController @Inject constructor(
...
+                val emptyThreadsQuery = getEmptyThreadsInFolderQuery(folderId, realm = this)
+                val emptyThreads = emptyThreadsQuery.find()
+                // TODO: Find why we are sometimes displaying empty Threads, and fix it instead of just deleting them.
+                //  It's possibly because we are out of sync, and the situation will resolve by itself shortly?
+                if (emptyThreads.isNotEmpty()) {
+                    emptyThreads.forEach {
+                        SentryDebug.sendEmptyThread(it, "No Message in a Thread when refreshing a Folder")
+                    }
+                    delete(emptyThreadsQuery)
+                }
```
LunarX, line 337: Do we want to send a sentry for each thread as in "this thread poses an issue" or do we want to group them all in a single sentry as a "An issue with thread as been detected, here is the list of impacted threads" and save some Sentry events?

KevinBoulongne: Yes, for each.

LunarX (approved):

PR #1878, opened 2024-05-29
fix: Only retry api calls for the current user 

### LINKED ISSUE #1871: NullPointerException
Sentry Issue: [MAIL-ANDROID-160](https://sentry-mobile.infomaniak.com/organizations/sentry/issues/6847/?referrer=github_integration)

```
NullPointerException: None
    at com.infomaniak.mail.MainApplication$tokenInterceptorListener$1.getApiToken(MainApplication.kt:246)
...
(5 additional frame(s) were not displayed)
```

sirambd:
Fixes MAIL-ANDROID-160
Fix #1871

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/MainApplication.kt
+++ b/app/src/main/java/com/infomaniak/mail/MainApplication.kt
@@ -206,9 +206,7 @@ open class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycle
         AccountUtils.reloadApp = { withContext(mainDispatcher) { startActivity(getLaunchIntent()) } }
     }
 
-    private fun getLaunchIntent() = Intent(this, LaunchActivity::class.java).apply {
-        clearStack()
-    }
+    private fun getLaunchIntent() = Intent(this, LaunchActivity::class.java).clearStack()
 
     private fun configureInfomaniakCore() {
         InfomaniakCore.apply {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/utils/LogoutUser.kt
```diff
@@ -56,6 +56,8 @@ class LogoutUser @Inject constructor(
         localSettings.removeRegisteredFirebaseUser(userId = user.id)
 
         if (user.id == AccountUtils.currentUserId) {
+            AccountUtils.currentUser = null // Prevent that the user current no longer exists
```
LunarX, line 59: I'm not sure what you meant with this comment. Maybe you meant `Prevent that the current user no longer exists` but still, if you meant that sentence, the user is set to `null` which means the user is no longer. So what is the actual meaning of this comment?

app/src/main/java/com/infomaniak/mail/MainApplication.kt
```diff
@@ -266,6 +264,8 @@ open class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycle
         }
 
         override suspend fun getApiToken(): ApiToken? = AccountUtils.currentUser?.apiToken
+
+        override fun getCurrentUserId(): Int = AccountUtils.currentUserId
```
LunarX, line 268: This won't work, the value is computed only once when the tokenInterceptorListener is instanciated

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/LogoutUser.kt
LunarX:
```suggestion
            AccountUtils.currentUser = null // Inform that there is no more current user
```

app/src/main/java/com/infomaniak/mail/MainApplication.kt
LunarX: Actually not

LunarX (approved):

PR #1880, opened 2024-05-30
Place information icon of external banner to the left of the banner

LunarX:
Avoids missclicking the close icon instead according to designers

### DIFF
--- a/app/src/main/res/layout/fragment_new_message.xml
+++ b/app/src/main/res/layout/fragment_new_message.xml
@@ -476,21 +476,24 @@
             app:layout_constraintTop_toBottomOf="@+id/compositionNestedScrollView"
             tools:visibility="visible">
 
-            <TextView
-                style="@style/BodySmall"
-                android:layout_width="0dp"
-                android:layout_height="wrap_content"
-                android:layout_marginHorizontal="@dimen/marginStandardMedium"
-                android:layout_weight="1"
-                android:text="@string/externalDialogTitleRecipient"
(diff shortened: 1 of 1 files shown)


KevinBoulongne (approved):

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -476,22 +476,26 @@
...
+                android:text="@string/externalDialogTitleRecipient"
+                android:textColor="@color/primaryTextColor"
                 app:icon="@drawable/ic_external_information"
+                app:iconGravity="end"
                 app:iconTint="@color/externalTagOnBackground" />
 
+            <Space
+                android:layout_width="0dp"
+                android:layout_height="0dp"
+                android:layout_weight="1" />
```
tevincent, line 497: A `ConstraintLayout` would be better instead of a Space IMO.

tevincent (changes requested):

app/src/main/res/layout/fragment_new_message.xml
LunarX: A ConstraintLayout is overkill when items are simply placed side by side like this I think. It requires an unnecessary amount of constraint to achieve this very simple result. Directly using a LinearLayout also helps you understand how the views are gonna be arranged where as using a ConstraintLayout requires you to go and apply every constraint in your mind to understand where views are placed from one another. In this sens, it's more meaningful/easy to grasp I think

tevincent (approved):

PR #1882, opened 2024-05-30
Return a copy of the list instead of the reference to avoid crash if this list is modified by another threads.

tevincent:
Fixes MAIL-ANDROID-7Z2

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/NotificationUtils.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/NotificationUtils.kt
@@ -48,7 +48,6 @@ import io.realm.kotlin.Realm
 import io.sentry.SentryLevel
 import kotlinx.coroutines.*
 import java.util.UUID
-import java.util.concurrent.ConcurrentHashMap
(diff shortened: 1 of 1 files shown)


sirambd (approved):

app/src/main/java/com/infomaniak/mail/utils/NotificationUtils.kt
```diff
@@ -235,7 +234,7 @@ class NotificationUtils @Inject constructor(
 
             @Suppress("MissingPermission")
             notificationsByMailboxId[mailboxId]?.let { notifications ->
-                notificationManagerCompat.notify(notifications)
+                notificationManagerCompat.notify(notifications.toList())
```
KevinBoulongne, line 237: The crash was that `notificationsByMailboxId[mailboxId]` throws OOB.
How can this fix solve the OOB ? 🤔

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1886, opened 2024-06-04
Split this XXX characters long line (which is greater than 130 authorized).

KevinBoulongne:
Some lines were too long.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/api/FlatteningSubBodiesSerializer.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/api/FlatteningSubBodiesSerializer.kt
@@ -22,7 +22,7 @@ import io.realm.kotlin.serializers.RealmListKSerializer
 import io.realm.kotlin.types.RealmList
 import kotlinx.serialization.json.*
 
-// Documentation: https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/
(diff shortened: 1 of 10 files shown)


LunarX:
Not sure this requiring this check is always productive actually

tevincent (approved):

app/src/main/java/com/infomaniak/mail/data/api/FlatteningSubBodiesSerializer.kt
```diff
@@ -22,7 +22,7 @@ import io.realm.kotlin.serializers.RealmListKSerializer
 import io.realm.kotlin.types.RealmList
 import kotlinx.serialization.json.*
 
-// Documentation: https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-json/kotlinx.serialization.json/-json-transforming-serializer/
+// Documentation on KotlinLang.org : https://chk.me/kYZiqN0
```
LunarX, line 25: Not sure about that solution as well

KevinBoulongne: What do you propose instead ?

LunarX: Yeah the comment wasn't very clear now that I read it again. I think we could keep that one on a single line as it doesn't obstruct reading

PR #1889, opened 2024-06-04
Remove mutable parameter.

KevinBoulongne:
We shouldn't have mutableList as function parameter.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
@@ -104,6 +104,7 @@ class ContactAdapter(
     }
 
     fun searchContacts(text: CharSequence) {
+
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
```diff
@@ -126,13 +127,9 @@ class ContactAdapter(
...
         }
 
-        fun publishResults(results: MutableList<MatchedContact>) {
-            matchedContacts = results
-            notifyDataSetChanged()
-        }
-
         searchQuery = text.toString()
-        publishResults(performFiltering(text))
+        matchedContacts = performFiltering(text)
```
tevincent, line 131: Can we improve this? Without using a mutable.

tevincent (changes requested):

tevincent (approved):

PR #1894, opened 2024-06-05
Temp fix attachments

NicolasBourdin88:
Set manually the 'has_attachments' while loading heavy data, this fix is temporary and will be remove when the API returns the good value.

The value return by the API for the lite version of the message is sometimes false, the API can return that they're are attachments but actually not and it can occurred in the two way. The API will probably be fixed but not soon.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -305,15 +305,7 @@ class ThreadController @Inject constructor(
                     }
 
                     // TODO: Remove this when the API returns the good value for `has_attachments`.
-                    if (!hasAttachmentsInThread) {
-                        messages.flatMapTo(mutableSetOf()) { it.threads }.forEach { thread ->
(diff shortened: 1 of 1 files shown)


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -309,6 +321,18 @@ class ThreadController @Inject constructor(
...
+
+        private fun verifyAttachmentsValues(hasAttachmentsInThread: Boolean, messages: List<Message>, realm: MutableRealm) {
+            if (!hasAttachmentsInThread) {
+                messages.flatMapTo(mutableSetOf()) { it.threads }.forEach { thread ->
+                    if (thread.hasAttachments) {
+                        updateThread(thread.uid, realm) {
+                            it?.hasAttachments = false
+                        }
+                    }
+                }
```
LunarX, line 333: If the bug can happen the other way around, also fix it

LunarX (changes requested):

LunarX (approved):

PR #1896, opened 2024-06-06
Fix calendar event canceled detection

LunarX:
The previous condition wasn't really needed and in veeeery extrem case might even have been wrong. This PR makes it so the detection logic is inline with iOS and the web

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/calendar/CalendarEvent.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/calendar/CalendarEvent.kt
@@ -102,8 +102,8 @@ class CalendarEvent() : EmbeddedRealmObject {
         result = 31 * result + isFullDay.hashCode()
         result = 31 * result + start.hashCode()
         result = 31 * result + end.hashCode()
-        result = 31 * result + _status.hashCode()
         result = 31 * result + attendees.hashCode()
+        result = 31 * result + _status.hashCode()
 
         return result
     }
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/data/models/calendar/CalendarEventResponse.kt
```diff
@@ -54,7 +53,7 @@ class CalendarEventResponse() : EmbeddedRealmObject {
 
     val calendarEvent get() = userStoredEvent ?: attachmentEvent
 
-    val isCanceled get() = isUserStoredEventDeleted || attachmentEventMethod == AttachmentEventMethod.CANCEL
+    val isCanceled get() = calendarEvent?.status == CalendarEventStatus.CANCELLED
```
valentinperignon, line 56: 👍 We must check the status of an event, and only the status, to know if it has been canceled of not

valentinperignon (commented):
Perfect

app/src/main/java/com/infomaniak/mail/data/models/calendar/CalendarEvent.kt
```diff
@@ -96,8 +102,15 @@ class CalendarEvent() : EmbeddedRealmObject {
         result = 31 * result + isFullDay.hashCode()
         result = 31 * result + start.hashCode()
         result = 31 * result + end.hashCode()
+        result = 31 * result + _status.hashCode()
         result = 31 * result + attendees.hashCode()
```
KevinBoulongne, line 106:
```suggestion
        result = 31 * result + attendees.hashCode()
        result = 31 * result + _status.hashCode()
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1897, opened 2024-06-07
Nested #1: Remove NestedScrollView in Move fragmentes

KevinBoulongne:
Remove the nested scrollView, and merge the 3 recycler views into only 1.

Also, the MoveFragment is no longer a `MenuFoldersFragment`, it's only a `Fragment` now.
In the end, the `MenuFoldersFragment` will probably be deleted.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -177,6 +177,6 @@ class Folder : RealmObject, Cloneable {
         const val DEFAULT_IS_HISTORY_COMPLETE = false
 
         const val INBOX_FOLDER_ID = "eJzz9HPyjwAABGYBgQ--"
-        const val CUSTOM_FOLDER_ROLE_ORDER = 0
+        private const val CUSTOM_FOLDER_ROLE_ORDER = 0
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -147,7 +147,11 @@ class Folder : RealmObject, Cloneable {
 
     override fun hashCode(): Int = id.hashCode()
 
-    public override fun clone(): Folder = super.clone() as Folder
+    /**
+     * We've had a reference issue when modifying the `shouldDisplayDivider` of a Folder before sending it to the Adapter.
+     * Cloning it to do a deep copy resolved the issue.
+     */
+    public override fun clone() = super.clone() as Folder
 
     enum class FolderRole(
         @StringRes val folderNameRes: Int,
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -44,7 +44,7 @@ import kotlinx.serialization.UseSerializers
 import kotlin.math.max
 
 @Serializable
-class Folder : RealmObject {
+class Folder : RealmObject, Cloneable {
 
     //region Remote data
     @PrimaryKey
@@ -143,6 +143,8 @@ class Folder : RealmObject {
     override fun equals(other: Any?) = other === this || (other is Folder && other.id == id)
 
     override fun hashCode(): Int = id.hashCode()
+
+    public override fun clone(): Folder = super.clone() as Folder
 
     enum class FolderRole(
         @StringRes val folderNameRes: Int,
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -82,6 +82,8 @@ class Folder : RealmObject, Cloneable {
     var isHidden: Boolean = false // For children only (a children Folder is hidden if its parent is collapsed)
     @Transient
     var isCollapsed: Boolean = false // For parents only (collapsing a parent Folder will hide its children)
+    @Transient
+    var roleOrder: Int = CUSTOM_FOLDER_ROLE_ORDER
     //endregion
 
     //region UI data (Transient & Ignore)
@@ -129,6 +131,7 @@ class Folder : RealmObject, Cloneable {
         this.isHistoryComplete = isHistoryComplete
         this.isHidden = isHidden
         this.isCollapsed = isCollapsed
+        this.roleOrder = role?.order ?: CUSTOM_FOLDER_ROLE_ORDER
     }
 
     fun getLocalizedName(context: Context): String {
@@ -152,14 +155,14 @@ class Folder : RealmObject, Cloneable {
         val order: Int,
         val matomoValue: String,
     ) {
-        INBOX(R.string.inboxFolder, R.drawable.ic_drawer_inbox, 0, "inboxFolder"),
+        INBOX(R.string.inboxFolder, R.drawable.ic_drawer_inbox, 8, "inboxFolder"),
+        COMMERCIAL(R.string.commercialFolder, R.drawable.ic_promotions, 7, "commercialFolder"),
+        SOCIALNETWORKS(R.string.socialNetworksFolder, R.drawable.ic_social_media, 6, "socialNetworksFolder"),
+        SENT(R.string.sentFolder, R.drawable.ic_sent_messages, 5, "sentFolder"),
         DRAFT(R.string.draftFolder, R.drawable.ic_draft, 4, "draftFolder"),
-        SENT(R.string.sentFolder, R.drawable.ic_sent_messages, 3, "sentFolder"),
-        SPAM(R.string.spamFolder, R.drawable.ic_spam, 5, "spamFolder"),
-        TRASH(R.string.trashFolder, R.drawable.ic_bin, 6, "trashFolder"),
-        ARCHIVE(R.string.archiveFolder, R.drawable.ic_archive_folder, 7, "archiveFolder"),
-        COMMERCIAL(R.string.commercialFolder, R.drawable.ic_promotions, 1, "commercialFolder"),
-        SOCIALNETWORKS(R.string.socialNetworksFolder, R.drawable.ic_social_media, 2, "socialNetworksFolder"),
+        SPAM(R.string.spamFolder, R.drawable.ic_spam, 3, "spamFolder"),
+        TRASH(R.string.trashFolder, R.drawable.ic_bin, 2, "trashFolder"),
+        ARCHIVE(R.string.archiveFolder, R.drawable.ic_archive_folder, 1, "archiveFolder"),
     }
 
     companion object {
@@ -170,5 +173,6 @@ class Folder : RealmObject, Cloneable {
         const val DEFAULT_IS_HISTORY_COMPLETE = false
 
         const val INBOX_FOLDER_ID = "eJzz9HPyjwAABGYBgQ--"
+        const val CUSTOM_FOLDER_ROLE_ORDER = 0
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -35,6 +35,7 @@ import io.realm.kotlin.serializers.RealmListKSerializer
 import io.realm.kotlin.types.RealmInstant
 import io.realm.kotlin.types.RealmList
 import io.realm.kotlin.types.RealmObject
+import io.realm.kotlin.types.annotations.Ignore
 import io.realm.kotlin.types.annotations.PrimaryKey
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
@@ -81,6 +82,12 @@ class Folder : RealmObject {
     var isHidden: Boolean = false // For children only (a children Folder is hidden if its parent is collapsed)
     @Transient
     var isCollapsed: Boolean = false // For parents only (collapsing a parent Folder will hide its children)
+    //endregion
+
+    //region UI data (Transient & Ignore)
+    @Transient
+    @Ignore
+    var shouldDisplayDivider: Boolean = false
     //endregion
 
     private val _parents by backlinks(Folder::children)
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
@@ -195,6 +195,9 @@ class FolderAdapter @Inject constructor(
             }
         }
 
+        /**
+         * If there was no Folder with children, and then now there's at least one, we need to indent the whole Folders list.
+         */
         fun notifyCollapsableFolders() {
             setFoldersJob?.cancel()
             setFoldersJob = globalCoroutineScope.launch {
--- a/app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
@@ -34,6 +34,7 @@ import com.infomaniak.mail.data.models.Folder.FolderRole
 import com.infomaniak.mail.databinding.ItemFolderMenuDrawerBinding
 import com.infomaniak.mail.databinding.ItemSelectableFolderBinding
 import com.infomaniak.mail.ui.main.menu.FolderAdapter.FolderViewHolder
+import com.infomaniak.mail.utils.UiUtils
 import com.infomaniak.mail.utils.UnreadDisplay
 import com.infomaniak.mail.utils.Utils.runCatchingRealm
 import com.infomaniak.mail.views.itemViews.*
@@ -100,6 +101,8 @@ class FolderAdapter @Inject constructor(
 
     override fun onBindViewHolder(holder: FolderViewHolder, position: Int) = with(holder.binding) {
         val folder = folders[position]
+
+        root.tag = if (folder.shouldDisplayDivider) null else UiUtils.IGNORE_DIVIDER_TAG
 
         when (getItemViewType(position)) {
             DisplayType.SELECTABLE_FOLDER.layout -> (this as ItemSelectableFolderBinding).root.displayFolder(folder)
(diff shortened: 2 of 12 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -17,55 +17,59 @@
...
+import com.infomaniak.mail.utils.extensions.bindAlertToViewLifecycle
 import com.infomaniak.mail.utils.extensions.handleEditorSearchAction
 import com.infomaniak.mail.utils.extensions.setOnClearTextClickListener
 import com.infomaniak.mail.utils.extensions.setSystemBarsColors
 import dagger.hilt.android.AndroidEntryPoint
 import javax.inject.Inject
 
 @AndroidEntryPoint
-class MoveFragment : MenuFoldersFragment() {
+class MoveFragment : Fragment() {
```
LunarX, line 53: Does the common parent Fragment entirely disappear by the end of the feature, when the menu drawer will also be adapted with the new logic?

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -17,55 +17,59 @@
...
+    @Inject
+    lateinit var folderController: FolderController
 
-    private var currentFolderId: String? = null
+    @Inject
+    lateinit var injectedFolderAdapter: FolderAdapter
 
-    override val defaultFoldersList: RecyclerView by lazy { binding.defaultFoldersList }
-    override val customFoldersList: RecyclerView by lazy { binding.customFoldersList }
+    private val folderAdapter inline get() = binding.foldersRecyclerView.adapter as FolderAdapter
```
LunarX, line 69: We already have a reference on the adapter with the injected version. I don't think we need to create an alternative getter for it. This would also spare use from having to call one of them "injected"

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -75,13 +79,31 @@ class MoveFragment : MenuFoldersFragment() {
...
         observeNewFolderCreation()
         observeSearchResults()
     }
 
+    private fun setupRecyclerView() = with(binding.foldersRecyclerView) {
+
+        adapter = injectedFolderAdapter(
+            isInMenuDrawer = false,
+            onFolderClicked = ::onFolderSelected,
+        )
```
LunarX, line 96: I would have probably put it on a single line, do you prefer it on two separate lines?

app/src/main/java/com/infomaniak/mail/utils/UiUtils.kt
```diff
@@ -37,6 +37,7 @@ import com.infomaniak.mail.utils.extensions.updateNavigationBarColor
 object UiUtils {
 
     const val FULLY_SLID = 1.0f
+    const val IGNORE_DIVIDER_TAG = "ignoreDividerTag"
```
LunarX, line 40: Should this not have gone into the DividerItemDecorator class instead?

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveViewModel.kt
```diff
@@ -49,20 +49,21 @@ class MoveViewModel @Inject constructor(
...
 
-    fun getFolderIdAndCustomFolders() = liveData(ioCoroutineContext) {
+    fun getFolders() = liveData(ioCoroutineContext) {
 
-        val folderId = messageUid?.let { messageController.getMessage(it)!!.folderId }
+        currentFolderId = messageUid?.let(messageController::getMessage)?.folderId
             ?: threadController.getThread(threadsUids.first())!!.folderId
 
-        val customFolders = folderController.getCustomFolders().getCustomMenuFolders()
+        val folders = folderController.getMoveFolders().getCustomMenuFolders()
```
LunarX, line 64: `getCustomMenuFolders()` doesn't only return the list of custom folders so maybe we can rename it with a better name

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -110,51 +132,26 @@ class MoveFragment : MenuFoldersFragment() {
...
+                toggleFolderListsVisibility(newQuery?.toString(), allFolders)
                 if (newQuery?.isNotBlank() == true) {
                     moveViewModel.filterFolders(newQuery.toString(), allFolders, shouldDebounce = true)
                 }
 
-                if (!hasAlreadyTrackedFolderSearch) {
+                if (!hasAlreadyTrackedSearch) {
                     trackMoveSearchEvent("executeSearch")
-                    hasAlreadyTrackedFolderSearch = true
+                    hasAlreadyTrackedSearch = true
```
LunarX, line 154: Most of the `setSearchBarUi()` code should be executed as fast as possible a single time inside the `onViewCreated()`, no need to wait for a liveData nor update as the liveData updates

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -165,14 +162,41 @@ class MoveFragment : MenuFoldersFragment() {
...
+    /**
+     * Asynchronously validate folder name locally
+     * @return error string, otherwise null
+     */
+    private fun checkForFolderCreationErrors(folderName: CharSequence): String? = when {
+        folderName.length > 255 -> getString(R.string.errorNewFolderNameTooLong)
+        folderName.contains(Regex(INVALID_CHARACTERS_PATTERN)) -> getString(R.string.errorNewFolderInvalidCharacter)
+        folderController.getRootFolder(folderName.toString()) != null -> context?.getString(R.string.errorNewFolderAlreadyExists)
+        else -> null
+    }
```
LunarX, line 197: We should put some of this code in common with the menu drawer fragment. This is a good example but this might not be the only one, maybe there's more logic around the new folder creation

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -75,13 +79,31 @@ class MoveFragment : MenuFoldersFragment() {
...
         setSystemBarsColors()
 
+        bindAlertToViewLifecycle(inputDialog)
+        setupRecyclerView()
         setupListeners()
-        setupFolderAdapters()
+        observeFolders()
         setupCreateFolderDialog()
         observeNewFolderCreation()
         observeSearchResults()
```
LunarX, line 88: We can probably regroup the `observeFolders()` and the `observeSearchResults()` inside a single liveData if you use a single RecyclerView. This would also remove the need for the `isSearching` global boolean and make everything more robust

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
```diff
@@ -165,14 +162,41 @@ class MoveFragment : MenuFoldersFragment() {
         }
     }
 
-    private fun toggleFolderListsVisibility(isSearching: Boolean) = with(binding) {
-        searchResultsList.isVisible = isSearching
-        moveFoldersLists.isGone = isSearching
+    private fun toggleFolderListsVisibility(query: String?, allFolders: List<Folder>) {
+        isSearching = !query.isNullOrBlank()
+        if (!isSearching) folderAdapter.setFolders(allFolders, moveViewModel.currentFolderId)
```
LunarX, line 167: If we use a single live data we don't need this anymore, but as of now, the name of the method doesn't mean anything anymore. We're not toggling the visibility of two lists anymore

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveViewModel.kt
```diff
@@ -49,20 +49,21 @@ class MoveViewModel @Inject constructor(
     private val messageUid inline get() = savedStateHandle.get<String?>(MoveFragmentArgs::messageUid.name)
     private val threadsUids inline get() = savedStateHandle.get<Array<String>>(MoveFragmentArgs::threadsUids.name)!!
 
+    var currentFolderId: String? = null
```
LunarX, line 52: Do we really want to rely on this kind of "side effect" for our logic? The value is not set as long as the getFolders() hasn't been called. This would become obvious when reading if the value was returned by the getFolders() instead. This would also prevent anyone from accessing it before it's ready

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
```diff
@@ -181,21 +181,41 @@ class FolderAdapter @Inject constructor(
...
+
+                if (!isFirstTime && collapsableFolderExistenceHasChanged) {
+                    Dispatchers.Main { notifyDataSetChanged() }
+                }
             }
         }
+
+        newCurrentFolderId?.let { currentFolderId = it }
+        submitList(foldersWithDivider())
+        if (isInMenuDrawer) notifyCollapsableFolders()
```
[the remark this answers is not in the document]
KevinBoulongne: This is about collapsable state of Folders, not indent.

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
```diff
@@ -181,21 +181,41 @@ class FolderAdapter @Inject constructor(
...
+        }
 
-            if (!isFirstTime && collapsableFolderExistenceHasChanged) {
-                Dispatchers.Main { notifyDataSetChanged() }
+        /**
+         * If there was no Folder with children, and then now there's at least one, we need to indent the whole Folders list.
+         */
+        fun notifyCollapsableFolders() {
+            setFoldersJob?.cancel()
+            setFoldersJob = globalCoroutineScope.launch {
```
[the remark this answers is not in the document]
KevinBoulongne: Don't know exactly why a global scope is used, that what's already used on `master`.

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: Yes.

app/src/main/java/com/infomaniak/mail/utils/UiUtils.kt
KevinBoulongne: The `DividerItemDecorator` class is in Core, and doesn't use `IGNORE_DIVIDER_TAG`.
So I think not.

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
```diff
@@ -137,20 +147,26 @@ class Folder : RealmObject {
 
     override fun hashCode(): Int = id.hashCode()
 
+    /**
+     * We've had a reference issue when modifying the `shouldDisplayDivider` of a Folder before sending it to the Adapter.
+     * Cloning it to do a deep copy resolved the issue.
+     */
+    public override fun clone() = super.clone() as Folder
+
```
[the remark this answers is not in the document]
KevinBoulongne: We didn't get it to work without it when we tried with @tevincent, but we can try again.

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
```diff
@@ -181,21 +181,41 @@ class FolderAdapter @Inject constructor(
...
+            return newFolders.map { folder ->
+                folder.clone().apply {
+                    shouldDisplayDivider = if (folder.role == null && isFirstCustomFolder) {
+                        isFirstCustomFolder = false
+                        true
+                    } else {
+                        false
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
KevinBoulongne: Same topic as https://github.com/Infomaniak/android-kMail/pull/1897#pullrequestreview-2107501486

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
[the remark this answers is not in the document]
KevinBoulongne: That's a new feature, can you create a ticket about it ?

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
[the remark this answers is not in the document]
KevinBoulongne: This is about displaying (or not) the chevrons before Folders.
It's a MenuDrawer only feature.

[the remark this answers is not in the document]
KevinBoulongne: This notify is very specific, and will probably almost never trigger.
I'm not sure that the complexity added by merging it with the other notify is worth it.

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
```diff
@@ -181,21 +181,41 @@ class FolderAdapter @Inject constructor(
...
+            setFoldersJob?.cancel()
+            setFoldersJob = globalCoroutineScope.launch {
+                val newHasCollapsableFolder = newFolders.any { it.canBeCollapsed }
+
+                val isFirstTime = hasCollapsableFolder == null
+                val collapsableFolderExistenceHasChanged = newHasCollapsableFolder != hasCollapsableFolder
+                hasCollapsableFolder = newHasCollapsableFolder
+
+                if (!isFirstTime && collapsableFolderExistenceHasChanged) {
+                    Dispatchers.Main { notifyDataSetChanged() }
```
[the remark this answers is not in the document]
KevinBoulongne: How do you do a notify on the whole dataSet with a payload ?

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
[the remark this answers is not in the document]
LunarX: This is actually a regression

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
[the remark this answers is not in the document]
LunarX: Ah I see, ok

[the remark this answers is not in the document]
LunarX: Yeah now I better understand what this notify does, I agree

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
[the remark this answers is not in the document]
LunarX: Let's maybe just add a TODO so if we ever modify that part one day, we can handle this at the same time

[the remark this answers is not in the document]
KevinBoulongne: Fixed here : https://github.com/Infomaniak/android-kMail/pull/1899

app/src/main/java/com/infomaniak/mail/utils/UiUtils.kt
LunarX: Ah yeah it takes a lambda to determine on what condition to ignore, ok, yeah, UiUtils is probably the best place indeed

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: Most of the `setSearchBarUi()` code needs this liveData value.
What do you want to do exactly ?

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: Done in https://github.com/Infomaniak/android-kMail/pull/1899

app/src/main/java/com/infomaniak/mail/ui/main/menu/FolderAdapter.kt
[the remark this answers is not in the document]
LunarX: Well, it's actually not in the end

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: It's already triggered a single time. It won't update because the liveData can't emit updates.

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: We can't.
`observeFolders()` triggers only once, and `observeSearchResults()` triggers every time a search is done.

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: We don't want to use a single liveData.

KevinBoulongne: Fixed here : https://github.com/Infomaniak/android-kMail/pull/1901

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveViewModel.kt
KevinBoulongne: Fixed here : https://github.com/Infomaniak/android-kMail/pull/1901

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
LunarX: We still wait for no reason to execute most of this code which also has the effect of not splitting concerns as well as we could and it impacts readability and understanding. We should only execute the part that actually needs to wait on the liveData inside the observer I think

LunarX: Ah if most of the code depends on the folders and it's only called once anyway, then there's no problem. I didn't look through it precisely enough

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
KevinBoulongne: Fixed in https://github.com/Infomaniak/android-kMail/pull/1901

LunarX: The linked PR doesn't really solve the issue from my point of view, maybe we can talk about it. Also, is there a specific reason to delay the fix to a future PR instead of including it in this one? If possible, it might be more efficient to resolve issues in the current PR to avoid tracking a complication like this through multiple PRs

app/src/main/java/com/infomaniak/mail/ui/main/menu/MoveFragment.kt
LunarX: Same here but as in a following comment but, is there a is there a specific reason to delay the fix to a future PR instead of including it in this one?

KevinBoulongne: Both PRs have been merged into only 1 PR.

PR #1908, opened 2024-06-12
Fix IllegalStateException because of unmanaged object in notif

FabianDevel:


tevincent (approved):

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -352,11 +353,15 @@ class ThreadController @Inject constructor(
         }
 
         private fun verifyAttachmentsValues(hasAttachmentsInThread: Boolean, messages: List<Message>, realm: MutableRealm) {
-            messages.flatMapTo(mutableSetOf()) { it.threads }.forEach { thread ->
+            val localMessages = if (messages.firstOrNull()?.isManaged() == true) {
+                messages
+            } else {
+                MessageController.getMessagesByUids(messages.map(Message::uid), realm)
+            }
```
LunarX, line 360: Maybe we should add a comment so we remember all of the reasoning behind it

LunarX (changes requested):

LunarX (approved):

PR #1916, opened 2024-06-13
Remove NestedScrollView in Move fragment

KevinBoulongne:
Depends on #1915

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -45,10 +45,6 @@ class FolderController @Inject constructor(
 ) {
 
     //region Get data
-    fun getCustomFolders(): RealmResults<Folder> {
-        return getCustomFoldersQuery(mailboxContentRealm()).find()
-    }
-
     fun getMoveFolders(): RealmResults<Folder> {
         return getMoveFoldersQuery(mailboxContentRealm()).find()
     }
@@ -151,7 +147,9 @@ class FolderController @Inject constructor(
 
         private fun getDefaultFoldersQuery(realm: TypedRealm): RealmQuery<Folder> {
             val hasRole = "${Folder.rolePropertyName} != nil"
-            return realm.query("$isNotSearch AND $hasRole")
+            return realm
+                .query<Folder>("$isNotSearch AND $hasRole")
+                .sort(Folder::roleOrder.name, Sort.DESCENDING)
         }
 
         private fun getCustomFoldersQuery(realm: TypedRealm): RealmQuery<Folder> {
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -49,6 +49,10 @@ class FolderController @Inject constructor(
         return getCustomFoldersQuery(mailboxContentRealm()).find()
     }
 
+    fun getMoveFolders(): RealmResults<Folder> {
+        return getMoveFoldersQuery(mailboxContentRealm()).find()
+    }
+
     fun getRootsFoldersAsync(): Flow<ResultsChange<Folder>> {
         return getFoldersQuery(mailboxContentRealm(), onlyRoots = true).asFlow()
     }
@@ -160,6 +164,15 @@ class FolderController @Inject constructor(
 
         private fun getFoldersQuery(exceptionsFoldersIds: List<String>, realm: TypedRealm): RealmQuery<Folder> {
             return realm.query("NOT ${Folder::id.name} IN $0 AND $isNotSearch", exceptionsFoldersIds)
+        }
+
+        private fun getMoveFoldersQuery(realm: TypedRealm): RealmQuery<Folder> {
+            val isNotDraft = "${Folder.rolePropertyName} != '${FolderRole.DRAFT.name}'"
+            return realm
+                .query<Folder>("$isNotSearch AND $isRootFolder AND $isNotDraft")
+                .sort(Folder::name.name, Sort.ASCENDING)
+                .sort(Folder::isFavorite.name, Sort.DESCENDING)
+                .sort(Folder::roleOrder.name, Sort.DESCENDING)
         }
 
         private fun getFolderQuery(key: String, value: String, realm: TypedRealm): RealmSingleQuery<Folder> {
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -164,6 +164,8 @@ class FolderController @Inject constructor(
             return realm.query("NOT ${Folder::id.name} IN $0 AND $isNotSearch", exceptionsFoldersIds)
         }
 
+        // TODO: Check if we can remove this `excludeDrafts` parameter after
+        //  trying to merge together as much `getFoldersQuery` as possible.
         private fun getMoveFoldersQuery(realm: TypedRealm, excludeDrafts: Boolean): RealmQuery<Folder> {
             val draftsQuery = if (excludeDrafts) " AND ${Folder.rolePropertyName} != '${FolderRole.DRAFT.name}'" else ""
             return realm
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -45,8 +45,8 @@ class FolderController @Inject constructor(
 ) {
 
     //region Get data
-    fun getMoveFolders(): RealmResults<Folder> {
-        return getMoveFoldersQuery(mailboxContentRealm()).find()
+    fun getMoveFolders(excludeDrafts: Boolean = false): RealmResults<Folder> {
+        return getMoveFoldersQuery(mailboxContentRealm(), excludeDrafts).find()
     }
 
     fun getRootsFoldersAsync(): Flow<ResultsChange<Folder>> {
@@ -164,10 +164,10 @@ class FolderController @Inject constructor(
             return realm.query("NOT ${Folder::id.name} IN $0 AND $isNotSearch", exceptionsFoldersIds)
         }
 
-        private fun getMoveFoldersQuery(realm: TypedRealm): RealmQuery<Folder> {
-            val isNotDraft = "${Folder.rolePropertyName} != '${FolderRole.DRAFT.name}'"
+        private fun getMoveFoldersQuery(realm: TypedRealm, excludeDrafts: Boolean): RealmQuery<Folder> {
+            val draftsQuery = if (excludeDrafts) " AND ${Folder.rolePropertyName} != '${FolderRole.DRAFT.name}'" else ""
             return realm
-                .query<Folder>("$isNotSearch AND $isRootFolder AND $isNotDraft")
+                .query<Folder>("$isNotSearch AND ${isRootFolder}${draftsQuery}")
                 .sort(Folder::name.name, Sort.ASCENDING)
                 .sort(Folder::isFavorite.name, Sort.DESCENDING)
                 .sort(Folder::roleOrder.name, Sort.DESCENDING)
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -45,8 +45,8 @@ class FolderController @Inject constructor(
 ) {
 
     //region Get data
-    fun getMoveFolders(excludeDrafts: Boolean = false): RealmResults<Folder> {
-        return getMoveFoldersQuery(mailboxContentRealm(), excludeDrafts).find()
+    fun getMoveFolders(): RealmResults<Folder> {
+        return getMoveFoldersQuery(mailboxContentRealm(), excludeDrafts = true).find()
     }
 
     fun getRootsFoldersAsync(): Flow<ResultsChange<Folder>> {
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -177,6 +177,6 @@ class Folder : RealmObject, Cloneable {
         const val DEFAULT_IS_HISTORY_COMPLETE = false
 
         const val INBOX_FOLDER_ID = "eJzz9HPyjwAABGYBgQ--"
-        const val CUSTOM_FOLDER_ROLE_ORDER = 0
+        private const val CUSTOM_FOLDER_ROLE_ORDER = 0
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
@@ -90,6 +90,9 @@ class Folder : RealmObject, Cloneable {
     @Transient
     @Ignore
     var shouldDisplayDivider: Boolean = false
+    @Transient
+    @Ignore
+    var shouldDisplayIndent: Boolean = true
     //endregion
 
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
```diff
@@ -137,20 +153,26 @@ class Folder : RealmObject {
 
     override fun hashCode(): Int = id.hashCode()
 
+    /**
+     * We've had a reference issue when modifying the `shouldDisplayDivider` of a Folder before sending it to the Adapter.
+     * Cloning it to do a deep copy resolved the issue.
+     */
+    public override fun clone() = super.clone() as Folder
```
KevinBoulongne, line 160: Couldn't do without it, but we moved it to a coroutine in the VM, instead of Main thread in the Adapter.

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveFragment.kt
```diff
@@ -76,17 +76,26 @@ class MoveFragment : MenuFoldersFragment() {
...
         setSystemBarsColors()
 
+        bindAlertToViewLifecycle(inputDialog)
+        setupRecyclerView()
         setupListeners()
-        setupFolderAdapters()
         setupCreateFolderDialog()
-        observeNewFolderCreation()
+        observeFolders()
         observeSearchResults()
```
KevinBoulongne, line 84: image

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
```diff
@@ -34,32 +32,30 @@ import com.infomaniak.mail.data.models.Folder.FolderRole
...
 
 class MoveAdapter @Inject constructor(
     private val globalCoroutineScope: CoroutineScope,
-) : Adapter<FolderViewHolder>() {
+) : ListAdapter<Folder, FolderViewHolder>(FolderDiffCallback()) {
 
-    private val foldersDiffer = AsyncListDiffer(this, FolderDiffCallback())
-
-    private inline val folders get() = foldersDiffer.currentList
+    private var setFoldersJob: Job? = null
```
LunarX, line 48: globalCoroutineScope and setFoldersJob are unused

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
```diff
@@ -178,35 +175,20 @@ class MoveAdapter @Inject constructor(
...
-
-            val isFirstTime = hasCollapsableFolder == null
-            val collapsableFolderExistenceHasChanged = newHasCollapsableFolder != hasCollapsableFolder
-            hasCollapsableFolder = newHasCollapsableFolder
-
-            if (!isFirstTime && collapsableFolderExistenceHasChanged) {
-                Dispatchers.Main { notifyDataSetChanged() }
-            }
-        }
+        newCurrentFolderId?.let { currentFolderId = it }
```
LunarX, line 179: Do we really need to make `newCurrentFolderId` nullable and give it a default value?

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
```diff
@@ -178,35 +175,20 @@ class MoveAdapter @Inject constructor(
...
+        submitList(newFolders)
     }
 
     fun updateSelectedState(newCurrentFolderId: String) {
         val previousCurrentFolderId = currentFolderId
         currentFolderId = newCurrentFolderId
-        previousCurrentFolderId?.let(::notifyCurrentItem)
-        notifyCurrentItem(newCurrentFolderId)
+        previousCurrentFolderId?.let(::notifyFolder)
+        notifyFolder(newCurrentFolderId)
```
LunarX, line 187: Since these two notify occur at the very same place and nowhere else, we could've looped through the list a single time and detected both at the same time to avoid looping over every folder twice. Is it worth it? I don't know

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveFragment.kt
```diff
@@ -101,79 +110,90 @@ class MoveFragment : MenuFoldersFragment() {
...
+    /**
+     * Asynchronously validate folder name locally
+     * @return error string, otherwise null
+     */
+    private fun checkForFolderCreationErrors(folderName: CharSequence): String? = when {
+        folderName.length > 255 -> getString(R.string.errorNewFolderNameTooLong)
+        folderName.contains(Regex(INVALID_CHARACTERS_PATTERN)) -> getString(R.string.errorNewFolderInvalidCharacter)
+        folderController.getRootFolder(folderName.toString()) != null -> context?.getString(R.string.errorNewFolderAlreadyExists)
+        else -> null
+    }
```
LunarX, line 161: This function is duplicated, can't we like, extract it as an extension (original idea not by me)

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveFragment.kt
```diff
@@ -101,79 +110,90 @@ class MoveFragment : MenuFoldersFragment() {
...
+                    moveViewModel.filterFolders(newQuery.toString(), allFolders, currentFolderId, shouldDebounce = true)
+                } else {
+                    moveAdapter.setFolders(allFolders, currentFolderId)
                 }
 
-                if (!hasAlreadyTrackedFolderSearch) {
+                if (!hasAlreadyTrackedSearch) {
                     trackMoveSearchEvent("executeSearch")
-                    hasAlreadyTrackedFolderSearch = true
+                    hasAlreadyTrackedSearch = true
```
LunarX, line 179: This would track again after every configuration change. Is it a problem? I don't know

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveFragment.kt
LunarX: Also, the Regex could have been instanciated a single time instead of being instanciated each time the check is made at every character, if it's not to complicated to achieve. And if this can be done using a `by lazy` even better

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveViewModel.kt
```diff
@@ -49,38 +49,70 @@ class MoveViewModel @Inject constructor(
...
-            val filteredFolders = folders.filter { folder ->
+
+            var isFirstRootAndCustom = true
+            val filteredFolders = folders.mapNotNull { folder ->
                 val folderName = folder.role?.folderNameRes?.let(appContext::getString) ?: folder.name
-                folderName.standardize().contains(query.standardize())
+                val isFound = folderName.standardize().contains(query.standardize())
+                if (isFound) {
+                    folder.clone().apply {
+                        shouldDisplayDivider = if (isRootAndCustom && isFirstRootAndCustom) {
```
LunarX, line 95: `isRootAndCustom` is heavier to compute, we might as well do it this way since it doesn't impact anything else also this should be pretty easy for processors optimizations not that it matter much

```suggestion
                        shouldDisplayDivider = if (isFirstRootAndCustom && isRootAndCustom) {
```

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveViewModel.kt
```diff
@@ -49,38 +49,70 @@ class MoveViewModel @Inject constructor(
...
+            }
+        }
+    }
+
+    fun filterFolders(
+        query: String,
+        folders: List<Folder>,
+        currentFolderId: String,
+        shouldDebounce: Boolean,
+    ) = viewModelScope.launch(ioCoroutineContext) {
```
LunarX, line 81: Can't we also store `allFolders` inside the ViewModel?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
```diff
@@ -45,8 +45,8 @@ class FolderController @Inject constructor(
 ) {
 
     //region Get data
-    fun getCustomFolders(): RealmResults<Folder> {
-        return getCustomFoldersQuery(mailboxContentRealm()).find()
+    fun getMoveFolders(excludeDrafts: Boolean = false): RealmResults<Folder> {
+        return getMoveFoldersQuery(mailboxContentRealm(), excludeDrafts).find()
```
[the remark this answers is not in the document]
KevinBoulongne: Fixed.

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
KevinBoulongne: How do you want to do ?

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveFragment.kt
KevinBoulongne: Fixed.

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
LunarX: A different coroutine is always welcome but this won't help with the memory consumption issue ... are we really not just doing something bad somewhere above where we get the folders ...? This is really unfortunate if we really can't do anything about it

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
```diff
@@ -81,6 +82,17 @@ class Folder : RealmObject {
...
+    var roleOrder: Int = CUSTOM_FOLDER_ROLE_ORDER
+    //endregion
+
+    //region UI data (Transient & Ignore)
+    @Transient
+    @Ignore
+    var shouldDisplayDivider: Boolean = false
+    @Transient
+    @Ignore
+    var shouldDisplayIndent: Boolean = true
```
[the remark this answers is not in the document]
LunarX: This value is only used to trigger a difference in the adapter DiffUtil's `areContentsTheSame()`. But the boolean value is never accessed to draw something conditionally, is this really the best way to do things?

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
LunarX: Yeah, I don't know what I read the first time around, the value still needs to be nullable. Still, wouldn't we maybe prefer to split the selection of the current folder as its own method that can be called only once as early as we want independently of the setFolders method?

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
LunarX: This code was already there and is not used until the next menu drawer PR so it's fixed there

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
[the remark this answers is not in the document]
KevinBoulongne: The usage of this boolean was involuntarily moved to the next PR.
It's back here now.

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
KevinBoulongne: It's been revamped with you in #1917.

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
KevinBoulongne: We looked with @tevincent and @sirambd but couldn't find a better way.

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveViewModel.kt
```diff
@@ -18,15 +18,18 @@
...
 import com.infomaniak.mail.data.cache.mailboxContent.ThreadController
 import com.infomaniak.mail.data.models.Folder
 import com.infomaniak.mail.di.IoDispatcher
 import com.infomaniak.mail.utils.coroutineContext
 import com.infomaniak.mail.utils.extensions.appContext
-import com.infomaniak.mail.utils.extensions.getCustomMenuFolders
+import com.infomaniak.mail.utils.extensions.flattenFolderChildren
 import com.infomaniak.mail.utils.extensions.standardize
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.*
```
tevincent, line 35: Can we remove that wildcard import ?

app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
```diff
@@ -595,6 +588,20 @@ fun Fragment.bindAlertToViewLifecycle(alertDialog: BaseAlertDialog) {
...
+/**
+ * Asynchronously validate folder name locally
+ * @return error string, otherwise null
+ */
+private val invalidCharactersRegex by lazy { Regex("[/'\"]") }
+fun Context.checkForFolderCreationErrors(folderName: CharSequence, folderController: FolderController): String? {
+    return when {
+        folderName.length > 255 -> getString(R.string.errorNewFolderNameTooLong)
+        folderName.contains(invalidCharactersRegex) -> getString(R.string.errorNewFolderInvalidCharacter)
+        folderController.getRootFolder(folderName.toString()) != null -> getString(R.string.errorNewFolderAlreadyExists)
```
tevincent, line 600: Since you're returning a string error, maybe a suitable name cloud be `getFolderCreationError` ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveViewModel.kt
KevinBoulongne: Ok for me, but we should do it to the project level, so in another PR.

tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
LunarX: Do we really have to wait for the next PR? Can't we introduce it directly in this PR?

LunarX (changes requested):
Also, can't we directly merge this PR inside master, rather than creating a feature branch? Or are there feature that don't work yet in this branch?

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
KevinBoulongne: To test: directly add the divider in the list, as an item.

KevinBoulongne: The clone has been removed, but the complexity forced me to do in in another PR : #1947.

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
KevinBoulongne: Yes, there was too much changes in the next PR to do it in this one.

KevinBoulongne:
> Also, can't we directly merge this PR inside master, rather than creating a feature branch? Or are there feature that don't work yet in this branch?

I'd rather do it all on a feature branch to avoid messing with `master` in the event that we need to change stuff in here.

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
LunarX: I feel like moving fixes and change requests across PRs doesn't always help and sometimes makes it harder to follow, but at least the issue is fixed and it's cool you could find an alternative to this `clone()` method

app/src/main/java/com/infomaniak/mail/ui/main/move/MoveAdapter.kt
LunarX: I went and added a comment there

LunarX:
I feel like if we can avoid dragging around code for a long time inside feature branches when they are not needed, it would help us and it would force us to code more self contained PRs. But if you need the feature branch then there's no problem with that

LunarX (approved):

PR #1923, opened 2024-06-18
Don't replace Attachments with the 1st one when forwarding a Message

KevinBoulongne:
We've got an issue where, when forwarding an email, we replace all of its Attachments with the 1st one that we find.

It was caused by a wrong usage of `uploadLocalUri`, that we supposed couldn't be `null`.
The truth was that, when forwarding emails, Attachments don't have any `uploadLocalUri`.


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -813,8 +813,13 @@ class NewMessageViewModel @Inject constructor(
 
     private fun Draft.updateDraftAttachmentsFromLiveData() {
         val updatedAttachments = attachmentsLiveData.valueOrEmpty().map { uiAttachment ->
-            // If a localAttachment has the same `uploadLocalUri` than a UI one, it means it represents the same Attachment.
-            val localAttachment = attachments.filter { it.uploadLocalUri == uiAttachment.uploadLocalUri }
+            val localAttachment = attachments
+                /**
+                 * If a localAttachment has the same `uploadLocalUri` than a UI one, it means it represents the same Attachment.
+                 * But an Attachment only has an `uploadLocalUri` if the user added it to the Draft.
+                 * If it was added by forwarding a Message, it's doesn't have a `uploadLocalUri`, so we don't check this.
```
FabianDevel, line 820:
```suggestion
                 * If it was added by Message forwarding, it won't have any `uploadLocalUri`, so we don't check this.
```

FabianDevel (changes requested):
We could add like 2 lines to avoid doing the clear and the addAll if nothing change

If we forward a draft with a lot of attachment, that is a waste to clean the realm attachments, then add the same attachments back in realm

KevinBoulongne:
> We could add like 2 lines to avoid doing the clear and the addAll if nothing change
> 
> If we forward a draft with a lot of attachment, that is a waste to clean the realm attachments, then add the same attachments back in realm

We are not sure the Attachments in the `attachmentsLiveData` and in the Draft.attachments are already the same list to begin with, so I'm not sure we can do that.

app/src/main/java/com/infomaniak/mail/data/models/Attachment.kt
```diff
@@ -72,12 +72,21 @@ class Attachment : EmbeddedRealmObject {
...
+        mimeType: String? = null,
+    ): Attachment {
+
         this.uploadLocalUri = uri
 
+        uuid?.let { localUuid = it }
+        name?.let { this.name }
+        size?.let { this.size = it }
+        mimeType?.let { this.mimeType = it }
+
```
LunarX, line 89: I'm not a fan of having all these nullable arguments here. I think that when we actually use it to instantiate an object it should keep all of its non-null values and when we want to use it to do something else, like backup some values we could at least create another method explicitly used for this very and only purpose

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -812,9 +817,31 @@ class NewMessageViewModel @Inject constructor(
...
-            val localAttachment = attachments.filter { it.uploadLocalUri == uiAttachment.uploadLocalUri }
+
+        val uiAttachments = attachmentsLiveData.valueOrEmpty()
+
+        /**
+         * If :
+         * - we are in FORWARD mode,
+         * - all Attachments have no `uploadLocalUri` (meaning they are all from the original forwarded Message),
+         * - there quantity is the same in UI and in Realm,
+         * Then it means the Attachments list hasn't be edited by the user, so we have nothing to do here.
```
LunarX, line 828:
```suggestion
         * Then it means the Attachments list hasn't been edited by the user, so we have nothing to do here.
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -812,9 +817,31 @@ class NewMessageViewModel @Inject constructor(
...
+         */
+        if (draftMode == DraftMode.FORWARD &&
+            uiAttachments.all { it.uploadLocalUri == null } &&
```
LunarX, line 831: Having to constantly check throughout the app if we an `uploadLocalUri` or not to actually get another derived information is pretty hard to understand when reading the code with no context. Can we create some kind of property with a meaningful name that does this very check behind the scene? 

On a side note, this could go in a second PR if it impacts multiple places in the code I think

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
KevinBoulongne: Done in #1930.

LunarX (approved):

PR #1926, opened 2024-06-19
Add Sentry to check if the WebView version name is well formatted

tevincent:
Some user have the banner saying the WebView is outdated. Adding a sentry to understand what we have as version will help to understand the problem.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListViewModel.kt
@@ -25,6 +25,7 @@ import androidx.lifecycle.viewModelScope
 import androidx.webkit.WebViewCompat
 import com.infomaniak.mail.di.IoDispatcher
 import com.infomaniak.mail.utils.SearchUtils
+import com.infomaniak.mail.utils.SentryDebug
 import com.infomaniak.mail.utils.coroutineContext
 import com.infomaniak.mail.utils.extensions.appContext
 import dagger.hilt.android.lifecycle.HiltViewModel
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -307,6 +307,17 @@ object SentryDebug {
             Sentry.captureMessage("Credentials issue when trying to auto-sync user", SentryLevel.ERROR)
         }
     }
+
+    fun sendWebViewVersionName(webViewVersionName: String?, majorVersion: Int) {
+        Sentry.withScope { scope ->
+            scope.setExtra("webViewVersionName", webViewVersionName ?: "null")
```
KevinBoulongne, line 313:
```suggestion
            scope.setExtra("webViewVersionName", webViewVersionName.toString())
```

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListViewModel.kt
```diff
@@ -65,18 +66,16 @@ class ThreadListViewModel @Inject constructor(
...
-        } else {
-            false
-        }
+            val majorVersion = runCatching { versionName.substringBefore('.').toInt() }.getOrDefault(0)
+            versionName to majorVersion
+        } ?: (null to 0)
+
+        val needToUpdateWebView = canShowWebViewOutdated && webViewMajorVersion < WEBVIEW_MIN_VERSION
+
+        SentryDebug.sendWebViewVersionName(webViewVersionName, webViewMajorVersion)
```
KevinBoulongne, line 77: This Sentry will trigger at every ThreadList `onStart()`.
Is it what we want ?

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListViewModel.kt
KevinBoulongne: First, check the WebView version (always do it).
Then :
- If the version is wrong, send Sentry.
- If the version is wrong + the user never clicked "Later", display the banner.

KevinBoulongne (approved):

PR #1929, opened 2024-06-20
Move the access of `uiAttachments` liveData to `updateDraftAttachmentsFromLiveData()` parameters

KevinBoulongne:
We already have the info we need, no need to fetch them again from the LiveData.


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -811,8 +811,9 @@ class NewMessageViewModel @Inject constructor(
         }
     }
 
-    private fun Draft.updateDraftAttachmentsFromLiveData() {
-        val updatedAttachments = attachmentsLiveData.valueOrEmpty().map { uiAttachment ->
+    private fun Draft.updateDraftAttachmentsFromLiveData(uiAttachments: List<Attachment>) {
```
LunarX, line 814: We could rename the method with something along the lines of `updateDraftAttachmentsWithLiveData`

LunarX (changes requested):

LunarX (approved):

PR #1930, opened 2024-06-20
Simplify Attachment IDs usage

KevinBoulongne:
Depends on #1923

Replace `Attachment.uuid` usage with `Attachment.isAlreadyUploaded` when possible
Replace usage of `uploadLocalUri` with `localUuid` when possible

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/extensions/DraftExtensions.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/extensions/DraftExtensions.kt
@@ -53,4 +53,4 @@ suspend fun Draft.uploadAttachments(mailbox: Mailbox, draftController: DraftCont
     return true
 }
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/utils/extensions/DraftExtensions.kt
```diff
@@ -53,4 +53,4 @@ suspend fun Draft.uploadAttachments(mailbox: Mailbox, draftController: DraftCont
     return true
 }
 
-private fun Draft.getNotUploadedAttachments(): List<Attachment> = attachments.filter { it.uuid.isEmpty() }
+private fun Draft.getNotUploadedAttachments(): List<Attachment> = attachments.filter { !it.isAlreadyUploaded }
```
LunarX, line 56: Could we `filterNot`?

LunarX (changes requested):

LunarX (approved):

PR #1933, opened 2024-06-21
Add UploadStatus + Mutex to avoid concurrency when uploading Attachments

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/utils/extensions/AttachmentExtensions.kt
```diff
@@ -185,8 +184,12 @@ object AttachmentExtensions {
...
                 SentryLog.d(ATTACHMENT_TAG, "Target uploadLocalUri is: $uploadLocalUri")
 
-                // The API version of an Attachment doesn't have the `uploadLocalUri` & `localUuid`, so we need to back them up.
-                remoteAttachment.backupLocalData(oldAttachment = this@updateLocalAttachment)
+                remoteAttachment.backupLocalData(oldAttachment = this@updateLocalAttachment, UploadStatus.FINISHED)
+
+                SentryLog.d(ATTACHMENT_TAG, "Uploaded attachment name: ${remoteAttachment.name}")
+                SentryLog.d(ATTACHMENT_TAG, "Uploaded attachment uuid: ${remoteAttachment.uuid}")
+                SentryLog.d(ATTACHMENT_TAG, "Uploaded attachment localUuid: ${remoteAttachment.localUuid}")
+                SentryLog.d(ATTACHMENT_TAG, "Uploaded attachment uploadLocalUri: ${remoteAttachment.uploadLocalUri}")
```
tevincent, line 192: We might leak some personal information if we log `uploadLocalUri` right?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/utils/extensions/AttachmentExtensions.kt
KevinBoulongne: `name` was removed.

tevincent (approved):

PR #1940, opened 2024-06-26
Reduce Bimi priority (it should only be displayed if we couldn't find an avatar)

KevinBoulongne:
When we received a Bimi, we displayed it automatically.
But we should only do that if we can't find an avatar.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/views/AvatarView.kt
+++ b/app/src/main/java/com/infomaniak/mail/views/AvatarView.kt
@@ -58,11 +58,11 @@ class AvatarView @JvmOverloads constructor(
     private val binding by lazy { ViewAvatarBinding.inflate(LayoutInflater.from(context), this, true) }
 
     private var currentCorrespondent: Correspondent? = null
-    private var _bimi: Bimi? = null
+    private var isBimiShown: Boolean = false
 
     private val mergedContactObserver = Observer<MergedContactDictionary> { contacts ->
         currentCorrespondent?.let { correspondent ->
-            loadAvatarUsingDictionary(correspondent, contacts, _bimi)
+            if (!isBimiShown) loadAvatarUsingDictionary(correspondent, contacts)
         }
     }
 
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/views/AvatarView.kt
LunarX, line 1: The logic in this class is already complicated enough. When supporting the bimi, I think we should avoid separating the places that choose if we display a bimi or an avatar to multiple places and we should also avoid letting the different `loadAvatar` methods handle too much concerns, as much as possible

LunarX: I'm trying some things out in this regard

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/views/AvatarView.kt
LunarX: I ended up changing too many things for a code proposal so here is the kind of modifications I can think about: https://github.com/Infomaniak/android-kMail/pull/1941

app/src/main/java/com/infomaniak/mail/views/AvatarView.kt
```diff
@@ -58,16 +58,24 @@ class AvatarView @JvmOverloads constructor(
...
+        val displayType = getAvatarDisplayType(currentCorrespondent, currentBimi)
+
+        if (displayType == AvatarDisplayType.CUSTOM_AVATAR || displayType == AvatarDisplayType.INITIALS) {
+            loadAvatarUsingDictionary(currentCorrespondent!!, contacts)
         }
     }
 
     @Inject
-    lateinit var avatarMergedContactData: AvatarMergedContactData
+    lateinit var _avatarMergedContactData: AvatarMergedContactData
```
NicolasBourdin88, line 72: Why add an underscore

LunarX (approved):

NicolasBourdin88 (approved):

PR #1958, opened 2024-07-11
Replace current editor with rich html editor

LunarX:
This PR doesn't add any functionality yet. It only replaces the current editor and reimplements everything so nothing breaks

Depends on https://github.com/Infomaniak/android-rich-html-editor/pull/32

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
@@ -22,9 +22,15 @@ package com.infomaniak.mail.ui.newMessage
  * @param type The type of representation of [content]. Each type will lead to different processing of the content.
  */
 data class BodyContentPayload(val content: String, val type: BodyContentType) {
+
     companion object {
-        fun emptyBody(): BodyContentPayload = BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
+        fun emptyBody() = BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
     }
 }
 
-enum class BodyContentType { HTML_SANITIZED, HTML_UNSANITIZED, TEXT_PLAIN_WITH_HTML, TEXT_PLAIN_WITHOUT_HTML }
+enum class BodyContentType {
+    HTML_SANITIZED,
+    HTML_UNSANITIZED,
+    TEXT_PLAIN_WITH_HTML,
+    TEXT_PLAIN_WITHOUT_HTML,
+}
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
@@ -1,0 +1,29 @@
+/*
+ * Infomaniak Mail - Android
+ * Copyright (C) 2024 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.mail.ui.newMessage
+
+
+/**
+ * @param content The string representation of the body in either html or plain text format.
+ * @param type The type of representation of [content]. Each type will lead to different processing of the content.
+ * @param isSanitized Is only required and cannot be omitted when [type] is [BodyContentType.HTML]. This describes whether or not
+ * the HTML can skip sanitization when we know it's already been sanitized earlier.
+ */
+data class BodyContentPayload(val content: String, val type: BodyContentType, val isSanitized: Boolean? = null)
+
+enum class BodyContentType { HTML, TEXT_PLAIN_WITH_HTML, TEXT_PLAIN_WITHOUT_HTML }
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
@@ -21,10 +21,8 @@ package com.infomaniak.mail.ui.newMessage
 /**
  * @param content The string representation of the body in either html or plain text format.
  * @param type The type of representation of [content]. Each type will lead to different processing of the content.
- * @param isSanitized Is only required and cannot be omitted when [type] is [BodyContentType.HTML]. This describes whether or not
- * the HTML can skip sanitization when we know it's already been sanitized earlier.
  */
-data class BodyContentPayload(val content: String, val type: BodyContentType, val isSanitized: Boolean? = null) {
+data class BodyContentPayload(val content: String, val type: BodyContentType) {
     companion object {
         fun emptyBody(): BodyContentPayload {
             return BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
@@ -32,4 +30,4 @@ data class BodyContentPayload(val content: String, val type: BodyContentType, va
     }
 }
 
-enum class BodyContentType { HTML, TEXT_PLAIN_WITH_HTML, TEXT_PLAIN_WITHOUT_HTML }
+enum class BodyContentType { HTML_SANITIZED, HTML_UNSANITIZED, TEXT_PLAIN_WITH_HTML, TEXT_PLAIN_WITHOUT_HTML }
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
@@ -24,7 +24,7 @@ package com.infomaniak.mail.ui.newMessage
 data class BodyContentPayload(val content: String, val type: BodyContentType) {
 
     companion object {
-        fun emptyBody() = BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
+        fun emptyBody() = BodyContentPayload(content = "", type = BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
     }
 }
 
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
@@ -24,6 +24,12 @@ package com.infomaniak.mail.ui.newMessage
  * @param isSanitized Is only required and cannot be omitted when [type] is [BodyContentType.HTML]. This describes whether or not
  * the HTML can skip sanitization when we know it's already been sanitized earlier.
  */
-data class BodyContentPayload(val content: String, val type: BodyContentType, val isSanitized: Boolean? = null)
+data class BodyContentPayload(val content: String, val type: BodyContentType, val isSanitized: Boolean? = null) {
+    companion object {
+        fun emptyBody(): BodyContentPayload {
+            return BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
+        }
+    }
+}
 
 enum class BodyContentType { HTML, TEXT_PLAIN_WITH_HTML, TEXT_PLAIN_WITHOUT_HTML }
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
@@ -17,16 +17,13 @@
  */
 package com.infomaniak.mail.ui.newMessage
 
-
 /**
  * @param content The string representation of the body in either html or plain text format.
  * @param type The type of representation of [content]. Each type will lead to different processing of the content.
  */
 data class BodyContentPayload(val content: String, val type: BodyContentType) {
     companion object {
-        fun emptyBody(): BodyContentPayload {
-            return BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
-        }
+        fun emptyBody(): BodyContentPayload = BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
     }
 }
 
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
@@ -37,6 +37,8 @@ class EditorContentManager @Inject constructor() {
         }
     }
 
+    private fun RichHtmlEditorWebView.setSanitizedHtml(html: String) = setHtml(html)
+
     private fun RichHtmlEditorWebView.setUnsanitizedHtml(html: String) = setSanitizedHtml(html.sanitize())
 
     private fun RichHtmlEditorWebView.setPlainTextAndInterpretHtml(text: String) {
@@ -46,8 +48,6 @@ class EditorContentManager @Inject constructor() {
     private fun RichHtmlEditorWebView.setPlainTextAndEscapeHtml(text: String) {
         setSanitizedHtml(text.escapeHtmlCharacters().replaceNewLines())
     }
-
-    private fun RichHtmlEditorWebView.setSanitizedHtml(html: String) = setHtml(html)
 
     private fun String.escapeHtmlCharacters(): String = Html.escapeHtml(this)
 
@@ -74,6 +74,6 @@ class EditorContentManager @Inject constructor() {
     }
 
     companion object {
-        val NEW_LINES_REGEX = "(\\r\\n|\\n)".toRegex()
+        private val NEW_LINES_REGEX = "(\\r\\n|\\n)".toRegex()
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
@@ -27,13 +27,8 @@ import javax.inject.Inject
 
 @FragmentScoped
 class EditorContentManager @Inject constructor() {
-    private lateinit var editor: RichHtmlEditorWebView
 
-    fun initValues(editor: RichHtmlEditorWebView) {
-        this.editor = editor
-    }
-
-    fun setContent(bodyContentPayload: BodyContentPayload) {
+    fun setContent(editor: RichHtmlEditorWebView, bodyContentPayload: BodyContentPayload) = with(editor) {
         when (bodyContentPayload.type) {
             BodyContentType.HTML -> setHtml(bodyContentPayload.content, bodyContentPayload.isSanitized!!)
             BodyContentType.TEXT_PLAIN_WITH_HTML -> setPlainTextAndInterpretHtml(bodyContentPayload.content)
@@ -41,16 +36,20 @@ class EditorContentManager @Inject constructor() {
         }
     }
 
-    private fun setHtml(html: String, isSanitized: Boolean) {
+    private fun RichHtmlEditorWebView.setHtml(html: String, isSanitized: Boolean) {
         val sanitizedHtml = if (isSanitized) html else html.sanitize()
         setSanitizedHtml(sanitizedHtml)
     }
 
-    private fun setPlainTextAndInterpretHtml(text: String) = setSanitizedHtml(text.replaceNewLines().sanitize())
+    private fun RichHtmlEditorWebView.setPlainTextAndInterpretHtml(text: String) {
+        setSanitizedHtml(text.replaceNewLines().sanitize())
+    }
 
-    private fun setPlainTextAndEscapeHtml(text: String) = setSanitizedHtml(text.escapeHtmlCharacters().replaceNewLines())
+    private fun RichHtmlEditorWebView.setPlainTextAndEscapeHtml(text: String) {
+        setSanitizedHtml(text.escapeHtmlCharacters().replaceNewLines())
+    }
 
-    private fun setSanitizedHtml(html: String) = editor.setHtml(html)
+    private fun RichHtmlEditorWebView.setSanitizedHtml(html: String) = setHtml(html)
 
     private fun String.escapeHtmlCharacters(): String = Html.escapeHtml(this)
 
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
@@ -1,0 +1,47 @@
+/*
+ * Infomaniak Mail - Android
+ * Copyright (C) 2024 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.mail.ui.newMessage
+
+import com.infomaniak.html.cleaner.HtmlSanitizer
+import com.infomaniak.lib.richhtmleditor.RichHtmlEditorWebView
+import dagger.hilt.android.scopes.FragmentScoped
+import org.jsoup.Jsoup
+import javax.inject.Inject
+
+@FragmentScoped
+class EditorContentManager @Inject constructor() {
+    private lateinit var editor: RichHtmlEditorWebView
+
+    fun initValues(editor: RichHtmlEditorWebView) {
+        this.editor = editor
+    }
+
+    fun setHtml(htmlPayload: HtmlPayload) = with(htmlPayload) {
+        if (isSanitized) setSanitizedHtml(html) else setUnsanitizedHtml(html)
+    }
+
+    private fun setUnsanitizedHtml(html: String) {
+        setSanitizedHtml(HtmlSanitizer.getInstance().sanitize(Jsoup.parse(html)).html())
+    }
+
+    private fun setSanitizedHtml(html: String) {
+        editor.setHtml(html)
+    }
+
+    data class HtmlPayload(val html: String, val isSanitized: Boolean)
+}
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
@@ -58,6 +58,12 @@ class EditorContentManager @Inject constructor() {
         .apply { outputSettings().prettyPrint(false) }
         .getHtmlWithoutDocumentWrapping()
 
+    // Jsoup wraps parsed html inside an <html> and <body> tag. This gives us a wrapped form of the html content. While the editor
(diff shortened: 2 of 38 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -730,31 +757,52 @@ class NewMessageViewModel @Inject constructor(
...
-    fun executeDraftActionWhenStopping(
-        action: DraftAction,
-        isFinishing: Boolean,
-        isTaskRoot: Boolean,
-        subjectValue: String,
-        uiBodyValue: String,
-        startWorkerCallback: () -> Unit,
-    ) = globalCoroutineScope.launch(ioDispatcher) {
+    fun saveBodyAndSubject(subject: String, html: String) {
+        globalCoroutineScope.launch {
```
KevinBoulongne, line 761:
```suggestion
        globalCoroutineScope.launch(ioDispatcher) {
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -263,7 +275,7 @@ class NewMessageViewModel @Inject constructor(
 
         populateWithExternalMailDataIfNeeded(draft = this, intent)
 
-        body = getWholeBody()
+        body = getWholeBody("", uiSignature, uiQuote)
```
KevinBoulongne, line 278: Here, the `uiBody` isn't necessary empty _(ex: mailTo with payload)_.

Why hardcoding an empty string instead of sending the `uiBody` ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -302,11 +328,7 @@ class NewMessageViewModel @Inject constructor(
             doc.split(MessageBodyUtils.INFOMANIAK_FORWARD_QUOTE_HTML_CLASS_NAME, bodyWithQuote)
         }
 
-        draft.apply {
-            uiBody = body.htmlToText()
```
KevinBoulongne, line 306: Why no more `.htmlToText()` ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -110,8 +114,16 @@ class NewMessageViewModel @Inject constructor(
     val uiQuoteLiveData = MutableLiveData<String?>()
     //endregion
 
-    var lastOnStopSubjectValue = ""
-    var lastOnStopBodyValue = ""
+    val editorBodyLoader = SingleLiveEvent<BodyContentPayload>()
```
KevinBoulongne, line 117: A loader that contains data ?

Maybe the `loader` part of the name is a little confusing.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -576,7 +600,10 @@ class NewMessageViewModel @Inject constructor(
             bcc.addAll(splitBcc)
 
             subject = mailToIntent?.subject?.takeIf(String::isNotEmpty) ?: intent?.getStringExtra(Intent.EXTRA_SUBJECT)
-            uiBody = mailToIntent?.body?.takeIf(String::isNotEmpty) ?: intent?.getStringExtra(Intent.EXTRA_TEXT) ?: ""
+
+            val bodyContent: String? = mailToIntent?.body?.takeIf(String::isNotEmpty) ?: intent?.getStringExtra(Intent.EXTRA_TEXT)
```
KevinBoulongne, line 604:
```suggestion
            val bodyContent = mailToIntent?.body?.takeIf(String::isNotEmpty) ?: intent?.getStringExtra(Intent.EXTRA_TEXT)
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * @param content The string representation of the body in either html or plain text format.
+ * @param type The type of representation of [content]. Each type will lead to different processing of the content.
+ */
+data class BodyContentPayload(val content: String, val type: BodyContentType) {
+    companion object {
+        fun emptyBody(): BodyContentPayload {
+            return BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
+        }
```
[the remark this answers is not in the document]
LunarX: Oh

app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    private fun RichHtmlEditorWebView.setPlainTextAndInterpretHtml(text: String) {
+        setSanitizedHtml(text.replaceNewLines().sanitize())
+    }
+
+    private fun RichHtmlEditorWebView.setPlainTextAndEscapeHtml(text: String) {
+        setSanitizedHtml(text.escapeHtmlCharacters().replaceNewLines())
+    }
+
+    private fun RichHtmlEditorWebView.setSanitizedHtml(html: String) = setHtml(html)
```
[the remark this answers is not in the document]
LunarX: Yeah I changed the order in which these methods were called, but didn't notice the declarations were now not in order anymore, nice catch

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
tevincent, line 761: Any reason to use `globalCoroutineScope` instead of the `viewModelScope` ?

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -284,34 +284,38 @@
...
                             android:layout_width="match_parent"
-                            android:layout_height="wrap_content"
-                            android:gravity="top"
-                            android:hint="@string/newMessagePlaceholderTitle"
-                            android:inputType="textMultiLine|textCapSentences"
-                            android:nextFocusUp="@id/subjectTextField"
-                            android:paddingBottom="0dp"
-                            android:visibility="gone"
-                            tools:visibility="visible" />
+                            android:layout_height="wrap_content" />
```
tevincent, line 299: Do we need to have 2 levels here (FrameLayout -> Editor) instead of just one?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -792,11 +840,9 @@ class NewMessageViewModel @Inject constructor(
...
         subject = subjectValue
 
-        uiBody = uiBodyValue
-        uiSignature = uiSignatureLiveData.value
-        uiQuote = uiQuoteLiveData.value
-
-        body = getWholeBody()
+        val signature = uiSignatureLiveData.value
+        val quote = uiQuoteLiveData.value
+        body = getWholeBody(uiBodyValue, signature, quote)
```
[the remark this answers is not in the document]
LunarX: These three fields are `@Ignore` and cannot ever be saved to Realm.

These values are only ever used during snapshot comparison and during `initDraftAndViewModel()`. In this second usage, it is set at the beginning of the method (either by actually setting the objet's field or by instantiating a new instance of draft with its default value) and it is read by the end of the method.

To avoid impacting code that did not relate to the body I refrained from trying to change that part. Because I only modified the body and kept the signature and quote perfectly intact, I tried to not touch too much how things worked for signatures and quotes.

But because I changed how `saveSnapshot()` works, we could argue that I could've used this opportunity to remove uiBody, uiSignatures and uiQuotes from the Draft class.

We should probably either use variables defined inside the scope of the method or if we really need the side effect, we can have them plainly defined inside the view model and use the view model to transmit the data across parts of the `initDraftAndViewModel()`. All of this while passing in uiSignature and uiQuote as arguments of `saveSnapshot` the same way I had to do for uiBody.

What do you think of all this? Also, is this important enough that we need to refactor that part? If you agree with keeping any non-body matter intact in this PR, I could maybe create a new branch were I try and refactor that part entirely as an independent new task anyway.

Anyway, great question!

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -236,6 +244,10 @@ class NewMessageFragment : Fragment() {
     }
 
     override fun onDestroyView() {
+        binding.editor.exportHtml { html ->
+            newMessageViewModel.editorBodyLoader.postValue(BodyContentPayload(html, BodyContentType.HTML_SANITIZED))
+        }
```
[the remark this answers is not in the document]
LunarX: As @tevincent also asked me, I see that this could be hard to understand. I will add a comment to follow his advice. This block of code is needed in order to keep and reload the content of the editor across configuration changes

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -532,6 +556,12 @@ class NewMessageFragment : Fragment() {
         }
     }
 
+    private fun observeBodyLoader() {
+        newMessageViewModel.editorBodyLoader.observe(viewLifecycleOwner) {
+            editorContentManager.setContent(binding.editor, it)
```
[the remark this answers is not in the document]
LunarX: I named it

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: I think the reason why I placed an empty string is because `body` is never used in this situation, the whole line is useless. Even when there's a mailto with payload we never end up using the value of the field `body` which is set here. And setting it for no reason here, wrongly makes us believe that setting `body` using values like `uiBody` is required for the code to work properly and that we cannot stop relying on `uiBody` here.

But this is not the right way to show it anyway, you're right. We should not use the Draft's properties to pass a value like this around the `getExistingDraft()` function

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: Now that we use a rich html editor the input of the editor is not text anymore but html which is what's used here instead now

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: I tried adding something to the name so when reading where this value is used, we understand that this live data is not meant for keeping the content of the editor up to date or anything but only to initialize it.

Do you think that something more along the lines of `editorBodyInitializer` be easier to understand?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: Yeah, the code is executed (amongst other situations) when the activity stops which in turns stop the view model and its scope but the `exportHtml()`, being async, can call the callback after this point. In this situation, if we're using the `viewModelScope`, the code inside the scope will never be called because the scope has already been closed

app/src/main/res/layout/fragment_new_message.xml
LunarX: I thought it was needed for something but I must have been thinking of something else because it doesn't look it's needed here

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: I've modified the code in this regard, `body`'s usages are now limited to what's really needed

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -859,7 +904,7 @@ class NewMessageViewModel @Inject constructor(
         }
     }
 
-    private fun Draft.getWholeBody(): String = uiBody.textToHtml() + (uiSignature ?: "") + (uiQuote ?: "")
+    private fun getWholeBody(body: String, signature: String?, quote: String?): String = body + (signature ?: "") + (quote ?: "")
```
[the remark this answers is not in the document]
KevinBoulongne: The naming is fine, I just wanted to be sure we don't oversight anything 👌

tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: This is the only way for the draft's content to actually be loaded

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -730,31 +758,52 @@ class NewMessageViewModel @Inject constructor(
...
+                    break
+                }
+            }
+
+            draftSaveConfiguration.addSubjectAndBody(subject, body)
+            executeDraftAction(draftSaveConfiguration)
+        }
+    }
+
+    private suspend fun executeDraftAction(draftSaveConfiguration: DraftSaveConfiguration) = with(draftSaveConfiguration) {
```
FabianDevel, line 790: Make this function an extension of DraftSaveConfiguration

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -859,7 +905,8 @@ class NewMessageViewModel @Inject constructor(
         }
     }
 
-    private fun Draft.getWholeBody(): String = uiBody.textToHtml() + (uiSignature ?: "") + (uiQuote ?: "")
+    private fun assembleWholeBody(body: String, signature: String?, quote: String?): String =
+        body + (signature ?: "") + (quote ?: "")
```
FabianDevel, line 909: This function can be replaced by `"$body$signature$quote"` and is only used once, so it's not really worth keeping it as a function I think

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: Fabian also wanted this, so I already did it

app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
```diff
@@ -0,0 +1,30 @@
...
+ * @param content The string representation of the body in either html or plain text format.
+ * @param type The type of representation of [content]. Each type will lead to different processing of the content.
+ */
+data class BodyContentPayload(val content: String, val type: BodyContentType) {
+    companion object {
+        fun emptyBody(): BodyContentPayload = BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
+    }
+}
+
+enum class BodyContentType { HTML_SANITIZED, HTML_UNSANITIZED, TEXT_PLAIN_WITH_HTML, TEXT_PLAIN_WITHOUT_HTML }
```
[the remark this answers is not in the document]
LunarX: Well maybe, there should be when the enum has no argument and only a few values. Obviously most enums either have arguments or don't fit in a single line so most cannot ever be like this no matter what. 

Anyway, I changed this enum as you proposed, I think Kevin also likes them much more when they're defined on multiple lines

app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
```diff
@@ -0,0 +1,79 @@
...
+    private fun String.escapeHtmlCharacters(): String = Html.escapeHtml(this)
+
+    private fun String.replaceNewLines(): String = replace(NEW_LINES_REGEX, "<br>")
+
+    private fun String.sanitize(): String = HtmlSanitizer.getInstance()
+        .sanitize(Jsoup.parse(this))
+        .apply { outputSettings().prettyPrint(false) }
+        .extractHtmlWithoutDocumentWrapping()
+
+    private fun Document.extractHtmlWithoutDocumentWrapping(): String {
```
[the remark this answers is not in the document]
LunarX: "get" instead of "extract" is much more simple and straight forward anyway so I like it better. I went with `getHtmlWithoutDocumentWrapping` as you proposed

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -110,8 +114,16 @@ class NewMessageViewModel @Inject constructor(
...
-    var lastOnStopBodyValue = ""
+    val editorBodyLoader = SingleLiveEvent<BodyContentPayload>()
+
+    // 1. Navigating to AiPropositionFragment causes NewMessageFragment to export its body to `subjectAndBodyChannel`.
+    // 2. Inserting the AI proposition navigates back to NewMessageFragment.
+    // 3. When saving or sending the draft now, the channel still holds the previous body as it wasn't consumed.
+    // channelExpirationIdTarget lets us identify and discard outdated messages, ensuring the correct message is processed.
+    private var channelExpirationIdTarget = 0
+    private val _subjectAndBodyChannel: Channel<Triple<String, String, Int>> = Channel(capacity = CONFLATED)
+    private val subjectAndBodyChannel: ReceiveChannel<Triple<String, String, Int>> = _subjectAndBodyChannel
```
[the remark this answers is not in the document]
LunarX: Yeah, this is definitely needed!

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -730,31 +758,52 @@ class NewMessageViewModel @Inject constructor(
...
+    fun discardOldBodyAndSubjectChannelMessages() {
+        channelExpirationIdTarget++
+    }
+
+    fun waitForBodyAndSubjectToExecuteDraftAction(draftSaveConfiguration: DraftSaveConfiguration) {
+        subjectAndBodyJob?.cancel()
+        subjectAndBodyJob = globalCoroutineScope.launch(ioDispatcher) {
+            val subject: String
+            val body: String
+            while (true) {
```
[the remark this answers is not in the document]
LunarX: Isn't explaining that receive is blocking the same as explaining that `val` instances can only be read and not set? If you see what I mean. Shouldn't we be careful about over explaining things? 

I think that small comments like this one really can't hurt so I added it anyway. But I raised the question regardless to open the discussion if you have an opinion about it

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -295,6 +308,19 @@ class NewMessageFragment : Fragment() {
...
+
+            addCss(context.readRawResource(R.raw.style))
+            addCss(context.readRawResource(R.raw.editor_style))
+
+            isEmptyFlow
+                .filterNotNull()
+                .onEach { isEditorEmpty -> newMessagePlaceholder.isVisible = isEditorEmpty }
+                .launchIn(lifecycleScope)
+        }
+
```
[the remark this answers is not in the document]
LunarX: Definitely!

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -730,31 +758,52 @@ class NewMessageViewModel @Inject constructor(
...
 
-    fun executeDraftActionWhenStopping(
-        action: DraftAction,
-        isFinishing: Boolean,
-        isTaskRoot: Boolean,
-        subjectValue: String,
-        uiBodyValue: String,
-        startWorkerCallback: () -> Unit,
-    ) = globalCoroutineScope.launch(ioDispatcher) {
+    fun saveBodyAndSubject(subject: String, html: String) {
```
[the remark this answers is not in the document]
LunarX: Here, we're indeed only saving the body and subject for a potential future use. Maybe we should talk about it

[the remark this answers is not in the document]
LunarX: We went with "store" instead of "save" as "save" looking too much like "saveDaft" was the actual issue

app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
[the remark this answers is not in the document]
FabianDevel: There's still a "lot" of enum that could fit on 1 line but aren't formatted like this

And even if I agree that we could discuss the logic of putting enum on 1 line when it can fit, I still think it shouldn't be onelined as it's a 'class', and I don't like a whole class being onelined 😄

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LunarX: After discussing about it, this object being only created right there for this only purpose makes it more suited for being passed as an argument

app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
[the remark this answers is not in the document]
FabianDevel: I read "extract" but I understood that `withoutDocumentWrapping` was relative to `html`and not `extract`. 
Thus, I was confused how you could extract something from a HTML that wasn't wrapped

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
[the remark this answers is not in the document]
FabianDevel: This is a bit of an over explaination indeed, but I think it can be needed as neither of us apart from Siramd had heard of the `ReceiveChannel` before because it's not that common to use it, and the fact that even when reading the documentation, I didn't quickly understood that it was blocking.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
```diff
@@ -61,6 +61,9 @@ class NewMessageAiManager @Inject constructor(
     private var _aiViewModel: AiViewModel? = null
     private inline val aiViewModel: AiViewModel get() = _aiViewModel!!
 
+    private var _editorContentManager: EditorContentManager? = null
+    private inline val editorContentManager: EditorContentManager get() = _editorContentManager!!
+
```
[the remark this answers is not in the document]
FabianDevel: We thought about making it an interface that we would just implement with a anonym object inside `NewMessageFragment`, as it represent the same memory usage as a `FragmentScoped`injection.

But in the end, we chose that injecting the class was pretty and convenient and it was more useful than having a more "semantically logic" object.

FabianDevel (changes requested):
There seems to be a blinking of the from and to fields when recreating the draft the first time (create a draft, then add an attachment for example)

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -300,13 +337,31 @@ class NewMessageFragment : Fragment() {
...
+            val isPlaceholderVisible = combine(
+                isEmptyFlow.filterNotNull(),
+                newMessageViewModel.isShimmering,
+            ) { isEditorEmpty, isShimmering -> isEditorEmpty && !isShimmering }
+
+            isPlaceholderVisible
+                .onEach { isVisible -> newMessagePlaceholder.isVisible = isVisible }
+                .launchIn(lifecycleScope)
+        }
+    }
```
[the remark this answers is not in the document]
LunarX: I tried doing this but I didn't find any way of doing it easily like you did. Apparently I wasn't looking at it the correct way because it wasn't even that hard to simplify

LunarX:
> There seems to be a blinking of the from and to fields when recreating the draft the first time (create a draft, then add an attachment for example)
It may be only if the activity is killed and recreated, in which case it's not really important

After testing, it indeed does not happen when the activity is still there and gets reused

FabianDevel (approved):

FabianDevel (changes requested):
The anchor of the snackbar isn't right when clicking on a unsupported attachment

LunarX:
> The anchor of the snackbar isn't right when clicking on a unsupported attachment

Fixed in: https://github.com/Infomaniak/android-kMail/pull/1958/commits/94374a410c63c6adeebf6bb71e03c0589d2b6aa4

FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
KevinBoulongne: We are saving the Draft into Realm just after.
By removing this line, it means the `body` won't be saved in Realm.

app/src/main/java/com/infomaniak/mail/ui/newMessage/BodyContentPayload.kt
```diff
@@ -0,0 +1,36 @@
...
+package com.infomaniak.mail.ui.newMessage
+
+/**
+ * @param content The string representation of the body in either html or plain text format.
+ * @param type The type of representation of [content]. Each type will lead to different processing of the content.
+ */
+data class BodyContentPayload(val content: String, val type: BodyContentType) {
+
+    companion object {
+        fun emptyBody() = BodyContentPayload("", BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
```
KevinBoulongne, line 27:
```suggestion
        fun emptyBody() = BodyContentPayload(content = "", type = BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/EditorContentManager.kt
```diff
@@ -0,0 +1,78 @@
...
+        val elements = html.children()
+
+        val canRemoveDocumentWrapping = nodeSize == 2
+                && elements.count() == 2
+                && elements[0].tagName().uppercase() == "HEAD"
+                && elements[0].childNodeSize() == 0
+                && elements[1].tagName().uppercase() == "BODY"
+
+        return if (canRemoveDocumentWrapping) body().html() else html()
+    }
```
KevinBoulongne, line 73: This could have some doc to explain what's being done with the magic values.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageRecipientFieldsManager.kt
```diff
@@ -125,11 +124,12 @@ class NewMessageRecipientFieldsManager @Inject constructor(private val snackbarM
     fun setOnFocusChangedListeners() = with(binding) {
         val listener = View.OnFocusChangeListener { _, hasFocus -> if (hasFocus) fieldGotFocus(null) }
         subjectTextField.onFocusChangeListener = listener
-        bodyTextField.onFocusChangeListener = listener
+        editorWebView.onFocusChangeListener = listener
     }
 
     fun focusBodyField() {
-        binding.bodyTextField.showKeyboard()
+        // TODO: Make it so keyboard is kept open through configuration changes whenever the editor gets focused
```
KevinBoulongne, line 131: Is this already handled in another PR ?

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -850,11 +908,10 @@ class NewMessageViewModel @Inject constructor(
         if (!isFinishing) {
             copyFromRealm()
                 .apply {
```
KevinBoulongne, line 910: Why do we do this `.apply { … }` ? 🤔

The `uiBody` assignation was used for the `saveSnapshot()` call, but the `uiSignature` & `uiQuote` ? 🤔

KevinBoulongne (changes requested):

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageRecipientFieldsManager.kt
LunarX: No. If we ever change that part, this could be an opportunity change this as well, but for the time being, we don't have the time to bother with this

KevinBoulongne (approved):

PR #1961, opened 2024-07-11
Fix a crash on API25 when using the get method on MatchResult when using AI

tevincent:


app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
```diff
@@ -88,15 +88,15 @@ class AiPropositionFragment : Fragment() {
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         setSystemBarsColors(statusBarColor = R.color.backgroundColor)
 
         handleBackDispatcher()
         setUi()
 
         if (aiViewModel.aiPropositionStatusLiveData.value == null) generateNewAiProposition()
-        observeAiProposition()
+        return observeAiProposition()
```
LunarX: Why return this here?

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiViewModel.kt
```diff
@@ -89,6 +90,30 @@ class AiViewModel @Inject constructor(
...
+                subject to content
+            }
+        } ?: (null to proposition)
+
+        return if (subject.isNullOrBlank() || content == null) {
+            null to proposition
+        } else {
+            subject to content
+        }
+    }
```
LunarX, line 115: Can't we keep the method exactly as it was by just renaming it while annotating it with the API min level and make it so the new `splitBodyAndSubject` only does an if/else to choose between calling the actual method or doing nothing and returning the value directly?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiViewModel.kt
tevincent: Can

LunarX (changes requested):

LunarX (approved):

PR #1964, opened 2024-07-15
chores(Sentry): Remove useless Sentry

FabianDevel:
This sentry is useless because it is a totally normal case. 
You can trigger it by opening a message on the webmail before the notif was fetched

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
@@ -186,7 +186,6 @@ class FetchMessagesManager @Inject constructor(
 
         // If the Message has already been seen before receiving the Notification, we don't want to display it.
         // We can leave safely.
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -183,19 +183,11 @@ class FetchMessagesManager @Inject constructor(
...
-                mailboxId = mailbox.mailboxId,
-                messageUid = sentryMessageUid,
-                mailbox = mailbox,
-            )
-            return
-        }
+
+        // If the Message has already been seen before receiving the Notification, we don't want to display it.
+        // We can leave safely.
+        // It can happened very simply if you open a message on the webmail for example
```
KevinBoulongne, line 189:
```suggestion
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #1980, opened 2024-07-24
Use single name import instead of wildcard

TommyDL-Infomaniak:


.idea/codeStyles/Project.xml
```diff
@@ -3,7 +3,6 @@
     <option name="RIGHT_MARGIN" value="130" />
     <JetCodeStyleSettings>
       <option name="LINE_BREAK_AFTER_MULTILINE_WHEN_ENTRY" value="false" />
-      <option name="NAME_COUNT_TO_USE_STAR_IMPORT" value="5" />
```
KevinBoulongne, line 6: There is still around 40 imports in the project that use wildcards, you didn't remove them all.

KevinBoulongne (changes requested):

.idea/codeStyles/Project.xml
KevinBoulongne: Done.

KevinBoulongne (approved):

PR #1982, opened 2024-07-26
Add more breadcrumbs to better understand Attachments issues

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -104,11 +104,43 @@ object SentryDebug {
...
     }
 
+    fun addAttachmentsBreadcrumb(draft: Draft) = with(draft) {
+
+        var count = 1
+        val data = mutableMapOf<String, Any>()
+
+        fun String.keyPad(): String = padStart(length = 15)
+        fun Int.countPad(): String = toString().padStart(length = 2, '0')
+        fun count(): String = "${count.countPad().also { count++ }}."
```
tevincent, line 114: `Pad` is a bit weird. I had to search what the padStart method did to understand what was the meaning of all this.

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -104,11 +104,43 @@ object SentryDebug {
...
+        fun format(index: Int): String = (index + 1).countPad()
+
+        data[count() + "email".keyPad()] = AccountUtils.currentMailboxEmail.toString()
+
+        data[count() + "draft".keyPad() + " - localUuid"] = localUuid
+        data[count() + "draft".keyPad() + " - remoteUuid"] = remoteUuid.toString()
+        data[count() + "draft".keyPad() + " - action"] = action?.name.toString()
+        data[count() + "draft".keyPad() + " - mode"] = when {
```
tevincent, line 122: You could create an intermediate variable with `"count() + "draft".keyPad() + " -"`

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -104,11 +104,43 @@ object SentryDebug {
...
+            forwardedUid != null -> "FORWARD"
+            else -> "NEW_MAIL"
+        }
+
+        data[count() + "attachments".keyPad() + " - count"] = attachments.count()
+
+        attachments.forEachIndexed { index, it ->
+            data[count() + "attachment #${format(index)}".keyPad()] =
+                "localUuid: ${it.localUuid} | uuid: ${it.uuid} | uploadLocalUri: ${it.uploadLocalUri}"
+            data[count() + "attachment #${format(index)}".keyPad()] = "uploadStatus: ${it.uploadStatus.name} | size: ${it.size}"
```
tevincent, line 133: Also here with `count() + "attachment #${format(index)}".keyPad()`

tevincent (changes requested):

tevincent:
You also have a code smell.

LunarX (changes requested):
Add a keyword so each different step where we record a breadcrumb can be identified so when analyzing the breadcrumbs we know which step introduced the issue

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
KevinBoulongne: I think it complicates the readability of the code.

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
KevinBoulongne: I think it complicates the readability of the code.

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
KevinBoulongne: What do you propose ?

KevinBoulongne: Fixed in #1988.

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
KevinBoulongne: Fixed in #1988.

PR #1992, opened 2024-08-02
Remove associated inline attachements when removing quote

LunarX:
If we want to remove cids contained in the quote, we can simply remove it like this.

This way, when forwarding an email with inline attachments, if we remove the forwarding block, we also discard its associated inline attachments

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/HtmlUtils.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/HtmlUtils.kt
@@ -1,0 +1,52 @@
+/*
+ * Infomaniak Mail - Android
+ * Copyright (C) 2024 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/utils/HtmlUtils.kt
```diff
@@ -0,0 +1,52 @@
...
+object HtmlUtils {
+
+    const val CID_PROTOCOL = "${MessageWebViewClient.CID_SCHEME}:"
+    const val SRC_ATTRIBUTE = "src"
+    private const val CID_IMAGE_CSS_QUERY = "img[$SRC_ATTRIBUTE^='$CID_PROTOCOL']"
+
+    fun <T> Document.processCids(
+        attachments: List<Attachment>,
+        associateDataToCid: (Attachment) -> T?,
+        onCidImageFound: (T, Element) -> Unit
```
KevinBoulongne, line 34:
```suggestion
        onCidImageFound: (T, Element) -> Unit,
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #2000, opened 2024-08-08
Batch API calls when there is more than 1000 UIDs

KevinBoulongne:
Depends on #2005

We can't send more than 1000 UIDs to the API, they will be refused.
So we batch them.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
@@ -439,11 +439,10 @@ object ApiRepository : ApiRepositoryCore() {
 
     /**
      * Create batches of the given values to perform the given request
-     * - Parameters:
-     *   - values: Data to batch
-     *   - limit: Chunk size
-     *   - perform: Request to perform
-     * - Returns: Array of the perform return type
+     * @param values Data to batch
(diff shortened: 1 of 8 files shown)


tevincent (commented):
Should we have something for `atLeastOneFailed` ? Because if one failed, we'll act as nothing happened and process succeeded calls right ?

app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -424,6 +437,22 @@ object ApiRepository : ApiRepositoryCore() {
...
 
+    /**
+     * Create batches of the given values to perform the given request
+     * - Parameters:
+     *   - values: Data to batch
+     *   - limit: Chunk size
+     *   - perform: Request to perform
+     * - Returns: Array of the perform return type
+     */
+    private fun <T, R> batchOver(
```
tevincent, line 448:
```kotlin
/**
* Create batches of the given values to perform the given request
* @param values Data to batch
* @param limit Chunk size
* @param perform Request to perform
* @return Array of the perform return type
*/
```

Capture d’écran 2024-08-15 à 08 00 22

tevincent (changes requested):

tevincent (approved):

PR #2025, opened 2024-08-29
Make sure we use the up-to-date version of Folder in RefreshController

KevinBoulongne:
There is a Sentry issue saying that we get Messages from the wrong Folder when we refresh Folders.
Can't see why for now, but it seems we get the wrong Folder ?? There is an issue with the folder ID.

This PR is trying to make sure we use the correct Folder so we get the correct Messages.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
@@ -388,19 +388,18 @@ class RefreshController @Inject constructor(
         apiResponse.data?.messages?.let { messages ->
 
             writeBlocking {
-                findLatest(folder)?.let { latestFolder ->
-                    val isConversationMode = localSettings.threadMode == ThreadMode.CONVERSATION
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -388,19 +388,18 @@ class RefreshController @Inject constructor(
...
+                val allImpactedThreads = createThreads(scope, upToDateFolder, messages, isConversationMode).also { threads ->
+                    val foldersIds = (if (isConversationMode) threads.map { it.folderId }.toSet() else emptySet()) + folder.id
+                    foldersIds.forEach { refreshUnreadCount(id = it, realm = this) }
                 }
+                SentryLog.d(
+                    "Realm",
+                    "Saved Messages: ${upToDateFolder.displayForSentry()} | ${upToDateFolder.messages.count()}",
+                )
+
+                impactedThreads += allImpactedThreads.filter { it.folderId == folder.id }
```
FabianDevel, line 402: Use upToDateFolder.id instead

FabianDevel (changes requested):

FabianDevel (approved):

PR #2044, opened 2024-09-12
When expanding a long Thread, display Messages by batches so it doesn't freeze the UI anymore

KevinBoulongne:
In the situation of a Thread long enough so we display the SuperCollapsedBlock :
- When clicking on the SCB, we have to display all remaining Messages.
- Instead of displaying all of them in one go, we loop them and send them 1 by 1 every few milliseconds.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
@@ -74,6 +74,7 @@ class ThreadViewModel @Inject constructor(
 
     val threadLive = MutableLiveData<Thread?>()
     val messagesLive = MutableLiveData<Pair<ThreadAdapterItems, MessagesWithoutHeavyData>>()
+    val batchedMessages = SingleLiveEvent<List<Any>>()
 
     val quickActionBarClicks = SingleLiveEvent<QuickActionBarResult>()
 
@@ -244,6 +245,23 @@ class ThreadViewModel @Inject constructor(
         }
 
         return@withContext message
+    }
+
+    fun displayBatchedMessages(items: List<Any>) = viewModelScope.launch(ioCoroutineContext) {
+
+        tailrec suspend fun sendBatchesRecursively(input: List<Any>, output: MutableList<Any>, batchSize: Int = 1) {
+
+            val batch = input.take(batchSize)
+            output.addAll(batch)
+            batchedMessages.postValue(output.toMutableList())
+            if (batch.size < batchSize) return
+
+            delay(50L)
+
+            sendBatchesRecursively(input.subList(batchSize, input.size), output)
+        }
+
+        sendBatchesRecursively(input = items, output = mutableListOf(), batchSize = 2)
     }
 
     fun openThread(threadUid: String) = liveData(ioCoroutineContext) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -73,6 +73,7 @@ class ThreadViewModel @Inject constructor(
 
     val threadLive = MutableLiveData<Thread?>()
     val messagesLive = MutableLiveData<Pair<ThreadAdapterItems, MessagesWithoutHeavyData>>()
+    val batchedMessages = SingleLiveEvent<List<Any>>()
```
LunarX, line 76: Instead of having to rely in SingleLiveEvent can't we have a live data that represents the list of all loaded messages? This way we can listen in a straight forward manner to the list of messages, when the list updates, we submit the new list. The logic to delay the loading can then be handle on its own independently and the logic is not split depending on if we have more or less messages than `ThreadViewModel.SUPER_COLLAPSED_BLOCK_MINIMUM_MESSAGES_LIMIT`

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -245,6 +246,25 @@ class ThreadViewModel @Inject constructor(
...
+            // won't trigger if we send the same list object (https://stackoverflow.com/questions/49726385).
+            batchedMessages.postValue(ArrayList(output))
+
+            if (batch.size < batchSize) return
+            delay(50L)
+            sendBatchesRecursively(input.subList(batchSize, input.size), output)
+        }
+
+        sendBatchesRecursively(input = items, output = mutableListOf(), batchSize = 2)
+    }
```
LunarX, line 266: Have you considered using a for loop?

LunarX (dismissed):

LunarX:
Why do we do all of this again though?

LunarX:
Is there no other way to avoid the freeze instead of having to introduce this unconventional delay between messages? A more conventional way.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -245,6 +246,25 @@ class ThreadViewModel @Inject constructor(
...
+
+            val batch = input.take(batchSize)
+            output.addAll(batch)
+
+            // We need to post a different list each time, because the `submitList` function in AsyncListDiffer
+            // won't trigger if we send the same list object (https://stackoverflow.com/questions/49726385).
+            batchedMessages.postValue(ArrayList(output))
+
+            if (batch.size < batchSize) return
+            delay(50L)
```
tevincent, line 261: That might be in a constant ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
KevinBoulongne: This will just complicated the code.
It means adding a new LiveData used only to `submitList()` to the Adapter, when we can directly `submitList()` now.

The logic to delay is already handled on its own independently, in `displayBatchedMessages()`.

LunarX: 1) Directly submitting the list by hand in two places makes the coupling of the code higher where as introducing a LiveData as the only place that links the logic where data is emitted from to the logic where it's consumed makes the coupling the lowest possible.

2) Also, how does this new logic interact with the print fragment?

LunarX: Right now, I feel like this PR adds more coupling and coherence issues to the already existing coupling and coherence issue. We should quickly handle this first, before this PR imo because it doesn't seem that hard nor long to fix. 

I think we should fix this before it gets uncontrollable

tevincent (approved):

PR #2055, opened 2024-09-27
Delete draft that cannot ever be sent because of corrupted attachments

LunarX:
Most of the sentry issues we receive regarding corrupted attachments come from drafts stored in DB that are retried everytime the worker is started. This bug was caused by the UploadStatus crash inside addDraftBreadcrumbs() for certain. I suspect that there might remain other reasons for the corrupted attachment bug because even with the UploadStatus' runCatching() I did reproduce the corrupted attachement bug once. Also the issue did exist before the addDraftBreadcrumbs() addition and hasn't yet been fixed.

The plan of this PR is to remove the drafts that will never be processed so they stop spamming Sentry. Once this goes live we can see the errors where the attachement is corrupted for another reason than UploadStatus and fix this as well

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
+++ b/app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
@@ -293,6 +293,10 @@ class DraftsActionsWorker @AssistedInject constructor(
                 SentryLevel.ERROR,
             )
 
+            // Remove the draft if it's corrupted instead of sending a sentry every time the worker starts again
+            draftController.deleteDraft(draft)
+            SentryLog.i("CorruptedAttachment", "Remove draft from realm due to corrupted attachment")
+
             return DraftActionResult(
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/workers/DraftsActionsWorker.kt
```diff
@@ -293,6 +293,10 @@ class DraftsActionsWorker @AssistedInject constructor(
                 SentryLevel.ERROR,
             )
 
+            // Remove the draft if it's corrupted instead of sending a sentry every time the worker starts again
```
KevinBoulongne, line 296:
```suggestion
            // Remove the Draft if it's corrupted instead of sending a Sentry every time the worker starts again
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

tevincent (approved):

PR #2058, opened 2024-09-30
Create `MailboxLocalValues` class to stop forgetting to pass values in `initLocalValues()`

KevinBoulongne:
Depends on #2059

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
@@ -21,10 +21,8 @@ import android.content.Context
 import com.infomaniak.lib.core.utils.SentryLog
 import com.infomaniak.mail.data.cache.RealmDatabase
 import com.infomaniak.mail.data.models.AppSettings
-import com.infomaniak.mail.data.models.Quotas
 import com.infomaniak.mail.data.models.mailbox.Mailbox
 import com.infomaniak.mail.data.models.mailbox.MailboxLocalValues
-import com.infomaniak.mail.data.models.mailbox.MailboxPermissions
 import com.infomaniak.mail.di.MailboxInfoRealm
 import com.infomaniak.mail.utils.AccountUtils
 import com.infomaniak.mail.utils.NotificationUtils.Companion.deleteMailNotificationChannel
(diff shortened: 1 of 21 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -377,8 +377,8 @@ class MainViewModel @Inject constructor(
             if (!isSuccess()) return@launch
             data?.let { externalMailInfo ->
                 mailboxController.updateMailbox(mailbox.objectId) {
-                    it.externalMailFlagEnabled = externalMailInfo.externalMailFlagEnabled
-                    it.trustedDomains = externalMailInfo.trustedDomains.toRealmList()
+                    it.local.externalMailFlagEnabled = externalMailInfo.externalMailFlagEnabled
+                    it.local.trustedDomains = externalMailInfo.trustedDomains.toRealmList()
```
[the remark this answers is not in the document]
LunarX: In this part of the code we have multiple spot where we modify the realm object the same way we do it for the `local.featureFlags` where we needed a `copyFromRealm()`. If we cannot remove the `copyFromRealm()` will we also need it in these places here?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/SignatureController.kt
```diff
@@ -31,7 +31,7 @@ class SignatureController @Inject constructor(
 
     //region Get data
     fun getSignaturesAsync(mailboxObjectId: String): Flow<RealmList<Signature>> {
-        return mailboxController.getMailboxAsync(mailboxObjectId).mapNotNull { it.obj?.signatures }
+        return mailboxController.getMailboxAsync(mailboxObjectId).mapNotNull { it.obj?.local?.signatures }
```
LunarX, line 34: Is it worth it to query the whole mailbox when we only need the local values' object? (This pattern appears in other controllers as well)

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
```diff
@@ -99,19 +101,10 @@ class MailboxController @Inject constructor(
             val remoteMailboxesIds = remoteMailboxes.map { remoteMailbox ->
 
                 SentryLog.d(RealmDatabase.TAG, "Mailboxes: Get current data")
-                val localMailbox = getMailbox(userId, remoteMailbox.mailboxId, realm = this)
+                val previousLocalValues = getMailbox(userId, remoteMailbox.mailboxId, realm = this)?.local?.copyFromRealm()
```
LunarX, line 104: What's the reason that makes this copyFromRealm() required? How does it fix the bimi issue? Because I'm not sure I understand correctly how copyFromRealm() should properly be used

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/SignatureController.kt
KevinBoulongne: This is an EmbeddedObject, we can't query it directly.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
KevinBoulongne: We are replacing the `mailboxLocalValues` with another one that's coming from Realm.
We need to `copyFromRealm()` it if we want it to work.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
```diff
@@ -204,7 +204,7 @@ class NewMessageAiManager @Inject constructor(
 
     fun observeAiFeatureFlagUpdates() {
         newMessageViewModel.currentMailboxLive.observeNotNull(viewLifecycleOwner) { mailbox ->
-            val isAiEnabled = mailbox.featureFlags.contains(FeatureFlag.AI)
+            val isAiEnabled = mailbox.local.featureFlags.contains(FeatureFlag.AI)
```
LunarX, line 207: This live data is only ever used to get the current mailbox to now if its local values has the ai flag, would it make sens to replace it with a live data that only returns ai's enabled status? You can still keep the currentMailboxLive as a private live data and map it in a second public `isAiEnabled` live data or something I guess. Or directly query the local values from realm without querying the whole mailbox if it makes sens

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/SignatureController.kt
LunarX: What, I asked @sirambd because I feared this exact situation and he told me we could. Maybe he did not understand what I meant

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
LunarX: Yeah I figured, but how does it work? What does it bring to the table? Why does it work? What's missing if we don't use it? How does it fix the bimi issue?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageAiManager.kt
KevinBoulongne: This is an EmbeddedObject, we can't query it directly.

app/src/main/java/com/infomaniak/mail/data/models/mailbox/Mailbox.kt
```diff
@@ -103,34 +80,25 @@ class Mailbox : RealmObject {
...
+    fun initLocalValues(userId: Int, localValues: MailboxLocalValues?) {
+        localValues?.let { mailboxLocalValues = it }
+        objectId = createObjectId(userId)
+        local.userId = userId
+    }
+
+    fun setFeatureFlags(featureFlags: List<String>) {
+        mailboxLocalValues = local.copyFromRealm().also {
+            it.featureFlags.setFeatureFlags(featureFlags)
+        }
```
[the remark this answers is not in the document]
KevinBoulongne: No we can't, it's an EmbeddedObject, we need to use its parent _(ie. the Mailbox)_.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
KevinBoulongne: You need to copy the value from Realm to be able to put it back in Realm later.
If we don't, Realm crashes, that's all.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/SignatureController.kt
KevinBoulongne: We can query it, but not directly.
There is no PrimaryKey to search for since it's an EmbeddedObject.

LunarX: As long as we can query it indirectly by checking its parent's primary key, it should be enough. But I don't know to what extent this is feasible

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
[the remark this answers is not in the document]
LunarX: If this doesn't work, it indeed introduces a new equivalent problem. It'd be even worse at that. 

But I'd be very surprised if Realm didn't create a way to listen to items in a DB and get notified when they change even in this situation. We should talk about it in person and maybe even talk about it with @sirambd

PR #2061, opened 2024-10-01
Fix using backspace to delete recipient on API 34+

NicolasBourdin88:
Delete recipient backspace wasn't working anymore from API 34, this PR fix it.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareChip.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareChip.kt
@@ -18,6 +18,7 @@
 package com.infomaniak.mail.ui.newMessage
 
 import android.content.Context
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareChip.kt
```diff
@@ -29,6 +30,19 @@ class BackspaceAwareChip @JvmOverloads constructor(
 
     private var onBackspace: () -> Unit = {}
 
+    init {
+        isFocusableInTouchMode = true
```
tevincent, line 34: You can set this in the xml:

```xml
    android:focusableInTouchMode="true"
```
I just tried and it's working.

tevincent (changes requested):

tevincent (approved):

PR #2072, opened 2024-10-08
Set avatar background color to white so transparent BIMIs are displayed correctly

LunarX:
This issue occurs with BIMIs avatar that use transparencey. Previously the background of the avatar was set to colorPriamary which clashed with some avatars' colors that where too close to the priamryColor. Setting it to white seems to be the way to go, how BIMIs where thought out

### DIFF
--- a/app/src/main/res/layout/view_avatar.xml
+++ b/app/src/main/res/layout/view_avatar.xml
@@ -30,7 +30,7 @@
         android:id="@+id/avatarImage"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:background="?android:attr/colorPrimary"
+        android:background="@color/white"
         android:contentDescription="@string/contentDescriptionUserAvatar"
         android:src="@drawable/placeholder"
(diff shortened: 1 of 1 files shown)


JorisBodin (approved):

KevinBoulongne (approved):

app/src/main/res/layout/view_avatar.xml
```diff
@@ -30,7 +30,7 @@
         android:id="@+id/avatarImage"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:background="?android:attr/colorPrimary"
+        android:background="@color/white"
```
KevinBoulongne, line 33: A -t is missing.

KevinBoulongne (changes requested):

app/src/main/res/layout/view_avatar.xml
KevinBoulongne: The -t is here now.
All is good.

KevinBoulongne (approved):

PR #2077, opened 2024-10-10
Remove `messages` realmList from `Folder` model

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -99,24 +99,34 @@ class Thread : RealmObject {
...
+                scope.setTag("issueType", reason)
+                scope.setTag("foldersId", _folders.joinToString { it.id })
+                scope.setTag("foldersCount", "${_folders.count()}")
+                scope.setExtra("folders_", "${_folders.map { "role:[${it.role?.name}] (id:[${it.id}])" }}")
+                scope.setExtra("foldersCount_", "${_folders.count()}")
+                scope.setExtra("threadUid", uid)
+                scope.setExtra("email", AccountUtils.currentMailboxEmail.toString())
+                scope.setExtra("exception", exception.message.toString())
+            }
+            _folders.first()
```
LunarX, line 129: If the reason is "no parents" won't this crash as well?

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -227,18 +227,16 @@ object SentryDebug {
             "Message is in wrong Folder (related to 'Message has multiple parent folders')",
             SentryLevel.ERROR,
         ) { scope ->
-            scope.setExtra(
-                "localMessageFolders",
-                "${localMessage?.foldersForSentry?.joinToString { "${it.role?.name} | ${it.id}" }}",
-            )
+            scope.setExtra("localMessage folderId", "${localMessage?.folderId}")
+            scope.setExtra("remoteMessage folderId", remoteMessage.folderId)
```
LunarX, line 231: Won't this be bothersome to follow in Sentry? Some of these captureMessages will have one type of extras and other will have another type of extras

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
KevinBoulongne: Yes, but there's nothing better to do here.
We want to return a non-nullable Folder, or crash.

LunarX (approved):

PR #2082, opened 2024-10-21
Truncate messages at 104_448 to prevent OOM error on SplitBody

TommyDL-Infomaniak:


app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -70,6 +72,12 @@ object MessageBodyUtils {
             block = {
                 // Do not nest jsoupParseWithLog and measureAndLogMemoryUsage so logs are independent from one another
                 val htmlDocument = jsoupParseWithLog(bodyContent)
+
+                if (htmlDocument.text().length > MESSAGE_LENGTH_LIMIT) {
+                    val content = bodyContent.substring(0, MESSAGE_LENGTH_LIMIT)
```
LunarX, line 77: Do we really need the substring? If we skip the processing of the mail, no need to substring it anymore right? The problem has already been avoided

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -38,6 +38,8 @@ object MessageBodyUtils {
 
     private const val QUOTE_DETECTION_TIMEOUT = 1_500L
 
+    private const val MESSAGE_LENGTH_LIMIT = 104_448
```
LunarX, line 41: Maybe add a comment to explain the reasoning behind this strange value so we know why it's here and what we can do with it when we will have forgotten the choice we made

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
TommyDL-Infomaniak: It is needed to prevent an Out Of Memory error, resulting in a crash

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -38,6 +38,10 @@ object MessageBodyUtils {
 
     private const val QUOTE_DETECTION_TIMEOUT = 1_500L
 
+    // Arbitrary maximum length that a message can reach without provoking a crash when opened in a thread with other equally long
+    // messages.
```
KevinBoulongne, line 42:
```suggestion
    // Arbitrary maximum length that a Message's body can reach without provoking
    // a crash when opened in a Thread with other equally long Messages.
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

LunarX (approved):

PR #2085, opened 2024-10-29
fix: Don't send a Sentry when we have an access_denied APIErrorException

tevincent:


app/src/main/java/com/infomaniak/mail/utils/ErrorCode.kt
```diff
@@ -36,6 +36,7 @@ object ErrorCode {
 
     //region Mailbox
     const val MAILBOX_LOCKED = "mailbox_locked"
+    const val ACCESS_DENIED = "access_denied"
```
KevinBoulongne, line 39: I'd like to wait a week or two before merging this.

I just merged a fix that should reduce the number of `access_denied` we receive.
I'd like to see the result of this fix before implementing this feature.

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/utils/ErrorCode.kt
KevinBoulongne: I was able to check Sentry, so we can merge now !

Thx.

KevinBoulongne (approved):

PR #2086, opened 2024-10-29
Add some sentry to better understand message display improvements that still leave some overscroll

LunarX:
When resizing the mail with js, after zooming, it can still scroll. Sometimes, the issue is caused because the target width in which we want to the fit all of the html is 0. In this case, no amount of processing will ever let us fit the content in a width of 0 obviously.

This PR adds a new clarifying sentry captureMessage to try and understand how we could possibly end up with a target width of 0 in the first place. This PR also adds a tag to the already existing captureMessage to understand in what proportion the issue is caused by zero-width webviews

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/MessageWebViewClient.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/MessageWebViewClient.kt
@@ -32,6 +32,7 @@ import com.infomaniak.mail.data.models.Attachment
 import com.infomaniak.mail.utils.LocalStorageUtils
 import com.infomaniak.mail.utils.Utils
 import com.infomaniak.mail.utils.Utils.runCatchingRealm
+import com.infomaniak.mail.utils.WebViewVersionUtils.getWebViewVersionData
 import io.sentry.Sentry
 import io.sentry.SentryLevel
 import java.io.ByteArrayInputStream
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/MessageWebViewClient.kt
```diff
@@ -101,7 +103,22 @@ class MessageWebViewClient(
 
     override fun onPageFinished(webView: WebView, url: String?) {
         runCatchingRealm {
-            webView.loadUrl("javascript:removeAllProperties(); normalizeMessageWidth(${webView.width.toDp()}, '$messageUid')")
+            val widthInDp = webView.width.toDp()
+            if (widthInDp <= 0) {
+                Sentry.withScope { scope ->
```
FabianDevel, line 108: I think having the version of the webview as a tag could be useful

app/src/main/java/com/infomaniak/mail/ui/main/thread/MessageWebViewClient.kt
```diff
@@ -101,7 +103,22 @@ class MessageWebViewClient(
...
+                    scope.level = SentryLevel.WARNING
+                    scope.setExtra("width", webView.width.toString())
+                    scope.setExtra("measuredWidth", webView.measuredWidth.toString())
+                    scope.setExtra("height", webView.height.toString())
+                    scope.setExtra("measuredHeight", webView.measuredHeight.toString())
+                    scope.setTag("visibility", webView.visibility.toString())
+                    scope.setTag("messageUid", messageUid)
```
FabianDevel, line 115: I'm not sure this is a useful info for us, the only use I thought of is seeing if for a given user, it's always the same message that bug
Is that what you want ?

FabianDevel (changes requested):
You have a 132 characters long line

app/src/main/java/com/infomaniak/mail/ui/main/thread/MessageWebViewClient.kt
LunarX: very good idea

app/src/main/java/com/infomaniak/mail/ui/main/thread/MessageWebViewClient.kt
LunarX: Kind of, maybe also seeing if a given message will always trigger the issue, this kind of thing. Not that this tells us straight forward the answer, but could give us a hint

FabianDevel (approved):

PR #2089, opened 2024-11-01
fix: 'changeToolbarColorOnScroll' crash

NicolasBourdin88:


app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
```diff
@@ -489,7 +492,9 @@ fun Fragment.changeToolbarColorOnScroll(
         valueAnimator = animateColorChange(oldColor, newColor, animate = true) { color ->
             oldColor = color
             toolbar.setBackgroundColor(color)
-            if (shouldUpdateStatusBar()) requireActivity().window.statusBarColor = color
+            if (shouldUpdateStatusBar()) {
+                viewLifecycleOwner.lifecycleScope.launch(Dispatchers.Main) { requireActivity().window.statusBarColor = color }
+            }
```
LunarX, line 497: But, this should have already been fixed, I don't understand. 

There's a already a Dispatchers.Main inside `animateColorChange()`, there should be no point in adding a second one here, why is it not enough?

Also, there should be other lines of code that will crash if not on the main Thread such as `toolbar.setBackgroundColor(color)` and `otherUpdates?.invoke(color)`. Do they not crash?

LunarX (changes requested):

NicolasBourdin88:
Already fixed

PR #2092, opened 2024-11-04
fix: Add missing Search folder check in a Sentry

KevinBoulongne:


FabianDevel (changes requested):
This won't work because the single will still crash if there is 2 threads

FabianDevel (approved):

PR #2094, opened 2024-11-05
fix: Refresh User & Mailboxes before fetching Notifications

KevinBoulongne:
When a Mailbox isn't attached to a User anymore, we need to update Realm to reflect this.
It's already done when starting the app, but it wasn't done when fetching Notifications in background.

### DIFF
--- a/app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
+++ b/app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
@@ -21,10 +21,12 @@ import android.content.Context
 import androidx.hilt.work.HiltWorker
 import androidx.work.*
 import com.infomaniak.lib.core.utils.SentryLog
+import com.infomaniak.mail.data.api.ApiRepository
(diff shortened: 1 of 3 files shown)


app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
```diff
@@ -63,21 +65,22 @@ class ProcessMessageNotificationsWorker @AssistedInject constructor(
...
             return@withContext Result.success()
         }
+
+        // Refresh User
+        AccountUtils.updateCurrentUser()
+
+        // Refresh Mailboxes
+        SentryLog.d(TAG, "Refresh mailboxes from remote")
+        with(ApiRepository.getMailboxes()) {
+            if (isSuccess()) mailboxController.updateMailboxes(data!!)
```
FabianDevel, line 75: If the call fails, we will have the same comportment as before, do we want to send a sentry if it failed for another reason than Network ?

FabianDevel (changes requested):

FabianDevel (approved):

PR #2101, opened 2024-11-18
Fix coral light color token

LunarX:
The color token was based on a wrong spec. We just noticed the issue, so here is the update color token with the correct value


valentinperignon (approved):
Awesome guy and awesome PR

PR #2104, opened 2024-11-19
fix: Handle `is_valid` API field for Mailboxes

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
```diff
@@ -196,11 +197,11 @@ class MailboxController @Inject constructor(
         }
 
         private fun getInvalidPasswordMailboxesQuery(userId: Int, realm: TypedRealm): RealmQuery<Mailbox> {
-            return realm.query("${checkHasUserId(userId)} AND NOT ($hasValidPassword OR $isMailboxLocked)")
+            return realm.query("${checkHasUserId(userId)} AND NOT ($hasValidPassword) AND NOT ($isLocked) AND $isValidInLDAP")
```
FabianDevel, line 200:
```suggestion
            return realm.query("${checkHasUserId(userId)} AND (NOT $hasValidPassword) AND (NOT $isLocked) AND $isValidInLDAP")
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
```diff
@@ -196,11 +197,11 @@ class MailboxController @Inject constructor(
...
         }
 
         private fun getLockedMailboxesQuery(userId: Int, realm: TypedRealm): RealmQuery<Mailbox> {
-            return realm.query("${checkHasUserId(userId)} AND $isMailboxLocked")
+            return realm.query("${checkHasUserId(userId)} AND ($isLocked OR NOT ($isValidInLDAP))")
```
FabianDevel, line 204:
```suggestion
            return realm.query("${checkHasUserId(userId)} AND ($isLocked OR (NOT $isValidInLDAP))")
```

app/src/main/java/com/infomaniak/mail/data/models/mailbox/Mailbox.kt
```diff
@@ -99,7 +101,7 @@ class Mailbox : RealmObject {
     inline val notificationGroupId get() = uuid.hashCode()
     inline val notificationGroupKey get() = uuid
 
-    inline val isValid get() = isPasswordValid && !isLocked
+    inline val isValid get() = isValidInLDAP && !isLocked && hasValidPassword
```
FabianDevel, line 104: We could rename to something this `isAccessible` like iOS to avoid any confusion with the api field

app/src/main/java/com/infomaniak/mail/ui/main/menuDrawer/MailboxesAdapter.kt
```diff
@@ -127,8 +127,8 @@ class MailboxesAdapter(
 
         itemStyle = if (isInMenuDrawer) SelectionStyle.MENU_DRAWER else SelectionStyle.OTHER
 
-        isPasswordOutdated = !mailbox.isPasswordValid
-        isMailboxLocked = mailbox.isLocked
+        isPasswordOutdated = !mailbox.hasValidPassword
+        isMailboxLocked = mailbox.isLocked || !mailbox.isValidInLDAP
```
FabianDevel, line 131: We could compute a new value isMailboxLocked to be sure that these values are always kept together as you use it twice

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
tevincent, line 200: Do you have to have parenthesis for `AND NOT ($isLocked)` ? `AND NOT $isLocked` is not enough ? Same question for `getValidMailboxesQuery`.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxInfo/MailboxController.kt
KevinBoulongne: True.
I removed most of them.

tevincent (approved):

FabianDevel (approved):

PR #2117, opened 2024-12-02
fix: Updated Italian language file

ghywyu:


KevinBoulongne:
Hello @ghywyu, thanks for your contribution !

We use conventional commits, so before we can accept your PR, you'll need to rename your commits messages.
Something like `fix: Update strings.xml` should do the trick.

KevinBoulongne (approved):

ghywyu:
@KevinBoulongne what else are you asking to do and how?

KevinBoulongne:
@ghywyu We've done it, so everything is fine now.
The correct version is here : #2122.

The updated texts will be available in Beta next week, and probably the week after that in Production.

Thanks a lot !

ghywyu:
I'm very disappointed by what Kevin (and whoever is behind him, or should I say they/they, because "we are...") has done, stealing credit for a contribution that I made, and that Kevin hasn't done at all, except to write "do this useless thing that's already been done". The previous AI translation without context was embarrassing, but this thing you did is even worse.

JorisBodin:
Hello @ghywyu,

Thank you very much for your valuable suggestions! We would like to point out, however, that our translations are not generated by artificial intelligence, but rather are reviewed and approved by real people. Mistakes do happen, so it's possible that some translations may not be entirely accurate.

Translation strings for our applications are shared between Android and iOS. These strings are added to an internal platform to which both Android and iOS teams have access, to avoid any differences in wording between the two platforms. Consequently, these translations must first be added to our internal platform before creating a PR on the project, as we have done. We then automatically retrieve these translations from this platform, as they will be correctly formatted.

For example, you've changed `avete` to `c'è`, but you've used a quote that isn't properly formatted and therefore wouldn't compile correctly. This is why we replace all quotes (`'`) with the typographic apostrophe (`’`).

In addition, we need [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), which your PR did not respect. Since you weren't aware of this convention and didn't understand our recommendations, for reasons of efficiency we've decided to redo the PR correctly, giving you the appropriate credit. This way, the changes will be available as of next week!

Thank you once again for your valuable suggestions, which we greatly appreciate. We hope you understand our situation.

PR #2118, opened 2024-12-05
feat: Lock the app when the screen turns off

LouisCAD:
Depends on https://github.com/Infomaniak/android-core/pull/249


FabianDevel (changes requested):
There is a regression, now coming back from newMessageActivtity after writing a mail for more than a minute triggers the appLock

LouisCAD:
I have been able to reproduce the issue, thanks.
This should be addressed in the last commit, I will test right after Christmas.

sirambd (approved):

PR #2133, opened 2025-01-14
chore: Remove logging NotAuthorized exception in RefreshController

tevincent:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -714,17 +715,19 @@ class RefreshController @Inject constructor(
     //region API calls
     private fun getDateOrderedMessagesUids(folderId: String): NewMessagesResult? {
         return with(ApiRepository.getDateOrderedMessagesUids(mailbox.uuid, folderId, okHttpClient)) {
-            if (!isSuccess()) throwErrorAsException()
+            if (canThrowException()) throwErrorAsException()
```
KevinBoulongne, line 718: I don't think we want to do that.

KevinBoulongne (changes requested):

PR #2134, opened 2025-01-15
fix: Don't log in Sentry AccessDenied & NotAuthorized exceptions

KevinBoulongne:


LunarX (commented):
Should we apply this to other apps? Like SwissTransfer?

KevinBoulongne:
> Should we apply this to other apps? Like SwissTransfer?

Probably only if we have Sentry issues with these exceptions.

tevincent (approved):

PR #2170, opened 2025-02-18
fix: Select date time dialog wrong values and wrong error messages

LunarX:
The error message sometimes not block the user when it should've or it would would block the user when it should not have. The values inside the time picker would also sometime be wrong and have a different one than the one displayed inside the dialog

### DIFF
--- a/app/src/main/res/layout/dialog_select_date_and_time_for_scheduled_draft.xml
+++ b/app/src/main/res/layout/dialog_select_date_and_time_for_scheduled_draft.xml
@@ -20,9 +20,11 @@
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
     android:layout_height="wrap_content"
+    android:divider="@drawable/spacer_standard_medium"
     android:orientation="vertical"
     android:paddingHorizontal="@dimen/marginStandard"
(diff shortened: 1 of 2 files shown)


TommyDL-Infomaniak (approved):

app/src/main/res/layout/dialog_select_date_and_time_for_scheduled_draft.xml
```diff
@@ -32,48 +34,44 @@
...
         tools:text="@string/datePickerTitle" />
 
-    <LinearLayout
+    <com.infomaniak.lib.core.views.EndIconTextInputLayout
         android:layout_width="match_parent"
-        android:layout_height="match_parent"
-        android:layout_marginVertical="@dimen/marginStandard"
-        android:baselineAligned="false">
+        android:layout_height="wrap_content"
+        app:endIconDrawable="@drawable/ic_chevron_down"
```
tevincent, line 40: Maybe we want to add a ripple to the endIcon ? Like Google in Gmail ?

tevincent (changes requested):

LunarX:
The sonar issues are actually false positive, this is some custom code of ours not a deprecated api. But yeah, I handled it thanks for notifying me

tevincent (approved):

PR #2172, opened 2025-02-19
chore: Remove logs when webview can still be scrolled after resize

tevincent:


app/src/main/res/raw/munge_email.js
```diff
@@ -76,11 +76,6 @@ function normalizeElementWidths(elements, webViewWidth, messageUid) {
             logInfo(`Zoom updated: documentWidth / element.scrollWidth -> ${documentWidth} / ${element.scrollWidth} = ${newZoom}.`);
             element.style.zoom = newZoom;
         }
-
-        if (document.documentElement.scrollWidth > document.documentElement.clientWidth) {
-            logInfo(`After zooming the mail it can still scroll: found clientWidth / scrollWidth -> ${document.documentElement.clientWidth} / ${document.documentElement.scrollWidth}`);
-            reportOverScroll(document.documentElement.clientWidth, document.documentElement.scrollWidth, messageUid);
-        }
```
LunarX, line 83: We could just keep the javascript logs to the console and simply not report the overscroll instead

LunarX (changes requested):

LunarX:
After further inspection, keeping this sentry could be interesting to fix the biggest cause we identified back in november of 2024. We might want to keep it if it's not that heavy on sentry or only report a certain percentage of it otherwise but not remove it entierly

PR #2200, opened 2025-02-26
fix: Fix notification worker

FabianDevel:


LouisCAD (approved):
I didn't test locally, but everything looks good to me.

LouisCAD (approved):
Even better. Hopefully we bring our CI runners back to life.

PR #2210, opened 2025-02-28
feat: Matomo my kSuite

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/303
Depends on #2213

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
+++ b/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
@@ -23,6 +23,7 @@ import android.content.Context
 import android.view.View
 import androidx.fragment.app.Fragment
 import androidx.navigation.NavDestination
+import com.infomaniak.core.myksuite.ui.utils.MatomoMyKSuite
 import com.infomaniak.lib.core.MatomoCore
 import com.infomaniak.lib.core.MatomoCore.TrackerAction
 import com.infomaniak.lib.core.utils.capitalizeFirstChar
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/infomaniak/mail/MatomoMail.kt
```diff
@@ -274,6 +275,14 @@ object MatomoMail : MatomoCore {
         trackEvent("scheduleSend", name)
     }
 
+    fun Fragment.trackMyKSuiteEvent(name: String) {
+        trackEvent(MatomoMyKSuite.CATEGORY_MY_KSUITE, name)
+    }
+
+    fun Context.trackMyKSuiteUpgradeBottomSheetEvent(name: String) {
+        trackEvent(MatomoMyKSuite.CATEGORY_MY_KSUITE_UPGRADE_BOTTOMSHEET, name)
+    }
```
tevincent, line 284: Do we need to different extensions ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
KevinBoulongne: Same answer as here: https://github.com/Infomaniak/android-kDrive/pull/1546?w=1#discussion_r1978768711

FabianDevel: Yes we decided to simplify the tracker to the fullest extent when we first added Matomo. 
All the event have followed this logic since, so we'll keep it like this

tevincent (approved):

PR #2211, opened 2025-02-28
feat(Matomo): Add tracker for block user dialog events

FabianDevel:
Depends on #2213

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
+++ b/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
@@ -132,6 +132,14 @@ object MatomoMail : MatomoCore {
         trackEvent("messageActions", name)
     }
 
+    fun Fragment.trackBlockUserAction(name: String) {
+        requireContext().trackBlockUserAction(name)
+    }
+
+    fun Context.trackBlockUserAction(name: String) {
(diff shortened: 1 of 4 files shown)


KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
```diff
@@ -132,6 +132,14 @@ object MatomoMail : MatomoCore {
         trackEvent("messageActions", name)
     }
 
+    fun Fragment.trackBlockUserAction(name: String) {
+        requireContext().trackBlockUserAction(name)
+    }
+
+    fun Context.trackBlockUserAction(name: String) {
+        trackEvent("blockUserAction", name)
+    }
```
tevincent, line 141: Do we really need two extensions ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
KevinBoulongne: Same answer as here: https://github.com/Infomaniak/android-kDrive/pull/1546#discussion_r1978768711

tevincent (approved):

PR #2214, opened 2025-03-04
fix: Realm doesn't fail silently anymore when trying to update uploaded Attachments

KevinBoulongne:


LunarX:
What even is the goal here? Why does writeBlocking fail differently than write?

app/src/main/java/com/infomaniak/mail/utils/DraftUtils.kt
```diff
@@ -35,13 +35,14 @@ import kotlinx.coroutines.sync.withLock
...
     mailbox: Mailbox,
     draftController: DraftController,
     realm: Realm,
 ): Draft = attachmentsUploadMutex.withLock {
+    val draft = DraftController.getDraft(localUuid, realm)!!
     draft.uploadAttachments(mailbox, draftController, realm)
-    val updatedDraft = DraftController.getDraft(draft.localUuid, realm)!!
+    val updatedDraft = DraftController.getDraft(localUuid, realm)!!
     return@withLock updatedDraft
 }
```
LunarX, line 47: Why do we need a `realm` and a `draftController`? Can't we use a single one?

app/src/main/java/com/infomaniak/mail/utils/DraftUtils.kt
LunarX, line 1: Inside `Draft.uploadAttachments()` there's an unused `setUploadStatus()` inner method

LunarX (changes requested):

LunarX (approved):

PR #2218, opened 2025-03-05
feat: Compute snooze state in thread

LunarX:
This PR simply computes the snooze state at the thread level and can directly be merged inside master as it's not even visible to the user.

Depends on #2232

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
@@ -202,7 +202,7 @@ class Thread : RealmObject {
 
     private fun updateThread() {
 
-        fun Thread.updateSnoozeSatesBasedOn(message: Message) {
+        fun Thread.updateSnoozeStatesBasedOn(message: Message) {
             message.snoozeState?.let {
                 _snoozeState = it.apiValue
                 snoozeEndDate = message.snoozeEndDate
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -208,6 +229,12 @@ class Thread : RealmObject {
             }
             if (message.hasAttachable) hasAttachable = true
             if (message.isScheduledDraft) numberOfScheduledDrafts++
+
+            updateSnoozeSatesBasedOn(message)
+        }
+
+        duplicates.forEach { duplicate ->
+            updateSnoozeSatesBasedOn(duplicate)
         }
```
KevinBoulongne, line 238:
```suggestion
        duplicates.forEach(::updateSnoozeSatesBasedOn)
```

KevinBoulongne, line 238: Do we really need/want to update the Thread based on the duplicated Messages ? 🤔

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -90,6 +93,18 @@ class Thread : RealmObject {
...
+    var snoozeEndDate: RealmInstant? = null
+    @Transient
+    var snoozeAction: String? = null
+    //endregion
+
+    //region UI data (Ignore)
+    @Ignore
+    var snoozeState: SnoozeState? by apiEnum(::_snoozeState)
+        private set
     //endregion
```
KevinBoulongne, line 108:
```suggestion
    @Ignore
    var snoozeState: SnoozeState? by apiEnum(::_snoozeState)
        private set
```

KevinBoulongne: The _"UI data"_ region is supposed to be _"@.Transient & @.Ignore"_.
Here, it's different, so I removed the region.

Does it even need to be @.Ignore here ? 🤔

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #2221, opened 2025-03-05
feat: Filter snooze in folders

LunarX:
Queries different threads in realm to choose what threads will be part of the snooze folder and the inbox folder. To do so, I introduce a RefreshStrategy so multiple folders can define different logics for the same step of the refresh algorithm.

This will also help us easily keep track of what bit of logic is different for the snooze folder if the API ever sends us more normal data for the snooze folder

Depends on #2220

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -248,8 +248,8 @@ class ThreadController @Inject constructor(
             withSnooze: Boolean,
             realm: TypedRealm,
         ): RealmQuery<Thread> {
-            // Checking for snoozeEndDate and snoozeAction on top of _snoozeState mimics the behavior on the web and helps avoid
-            // displaying threads that are in an incoherent state on the API
+            // Checking for snoozeEndDate and snoozeAction on top of _snoozeState mimics the webmail's behavior
+            // and helps to avoid displaying threads that are in an incoherent state on the API
             val isSnoozedState = "_snoozeState == $1 AND snoozeEndDate != null AND snoozeAction != null"
             val snoozeQuery = if (withSnooze) isSnoozedState else "NOT($isSnoozedState)"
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -242,6 +243,25 @@ class ThreadController @Inject constructor(
             return realm.query<Thread>("${Thread::folderId.name} == $0", folderId)
         }
 
+        private fun getThreadsWithSnoozeFilterQuery(
+            folderId: String,
+            withSnooze: Boolean,
+            realm: TypedRealm,
+        ): RealmQuery<Thread> {
+            val snoozeState = SnoozeState.Snoozed.apiValue
```
KevinBoulongne, line 251:
```suggestion
            val snoozedState = SnoozeState.Snoozed.apiValue
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -242,6 +243,25 @@ class ThreadController @Inject constructor(
...
+            // Checking for snoozeEndDate and snoozeAction on top of _snoozeState mimics the behavior on the web and helps avoid
+            // displaying threads that are in an incoherent state on the API
+            val isSnoozedState = "_snoozeState == $1 AND snoozeEndDate != null AND snoozeAction != null"
+            val snoozeQuery = if (withSnooze) isSnoozedState else "NOT($isSnoozedState)"
+
+            return realm.query<Thread>(
+                "${Thread::folderId.name} == $0 AND $snoozeQuery",
+                folderId,
+                snoozeState,
+            )
```
KevinBoulongne, line 262:
```suggestion
            return realm.query<Thread>("${Thread::folderId.name} == $0 AND $snoozeQuery", folderId, snoozeState)
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
LunarX: The name should stay "snooze" and not "snoozed" imo. Not only is it the same name as the api and the other field names, but this way it refers to the "snooze" concept which can be enabled, disabled or whatever, where as "snoozed" would refer more to the "enabled snooze" concept

LunarX: Ok it's used only to hold the "snoozed" value. Understandable. Actually I dont' even need the intermediate variable

PR #2222, opened 2025-03-05
feat: Introduce custom logic to handle deleted uids in the case of the snooze folder

LunarX:
Defines the custom logic to handle deleted uids for the snooze folder. It leverages the RefreshStrategy to define the custom logic.

Inside this PR there's an issue where the snooze folder is not refreshed after an action on a message. This is fixed in the very next PR.

Depends on #2234

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ImpactedFolders.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ImpactedFolders.kt
@@ -42,7 +42,7 @@ data class ImpactedFolders(
      */
     fun getFolderIds(realm: TypedRealm): Set<String> {
         folderRoles.forEach { folderRole ->
-            FolderController.getFolder(folderRole, realm)?.let { folderIds.add(it.id) }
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ImpactedFolders.kt
```diff
@@ -0,0 +1,50 @@
...
+        folderIds += otherTarget.folderIds
+        folderRoles += otherTarget.folderRoles
+    }
+
+    /**
+     * This method makes sure we have no duplicated folders by merging both lists into a single one
+     */
+    fun getFolderIds(realm: TypedRealm): Set<String> {
+        folderRoles.forEach { folderRole ->
+            FolderController.getFolder(folderRole, realm)?.let { folderIds.add(it.id) }
```
KevinBoulongne, line 45:
```suggestion
            FolderController.getFolder(folderRole, realm)?.id?.let(folderIds::add)
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #2223, opened 2025-03-05
feat: Refresh snooze after action

LunarX:
Don't forget to refresh the snooze folder after an action has been triggered on an inbox message. By default, this wasn't working because it wasn't ever expected that message with a uid of a folder might want to refresh other folders.

Also, do we want to rename variables of the new ImpactedFolder type?

Depends on #2236

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
@@ -421,15 +421,11 @@ fun List<Folder>.addDividerBeforeFirstCustomFolder(dividerType: Any): List<Any> 
 fun List<Message>.getFoldersIds(exception: String? = null): ImpactedFolders {
     val impactedFolders = ImpactedFolders()
 
-    forEach { message ->
-        when {
-            message.folderId == exception -> Unit
-            message.snoozeState == SnoozeState.Snoozed -> {
-                impactedFolders += message.folderId
-                impactedFolders += FolderRole.SNOOZED
-            }
-            else -> impactedFolders += message.folderId
-        }
+    for (message in this) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
```diff
@@ -416,7 +418,22 @@ fun List<Folder>.addDividerBeforeFirstCustomFolder(dividerType: Any): List<Any>
...
+    forEach { message ->
+        when {
+            message.folderId == exception -> Unit
+            message.snoozeState == SnoozeState.Snoozed -> {
+                impactedFolders += message.folderId
+                impactedFolders += FolderRole.SNOOZED
+            }
+            else -> impactedFolders += message.folderId
+        }
+    }
```
KevinBoulongne, line 433:
```suggestion
    for (message in this) {
        if (message.folderId == exception) continue
        impactedFolders += message.folderId
        if (message.snoozeState == SnoozeState.Snoozed) impactedFolders += FolderRole.SNOOZED
    }
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
KevinBoulongne: `impactedFolders += message.folderId` was duplicated, and I felt it was kinda hidden that it was supposed to be done every time.

LunarX: Yeah I agree, it's nice to show that it needs to be done *every* time with the way you wrote it

KevinBoulongne (approved):

PR #2232, opened 2025-03-07
refactor: Sort code in RefreshController

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
LunarX, line 1: Maybe we could keep the method name `addAllMessagesToAllThreads()` as well?

LunarX (approved):

PR #2233, opened 2025-03-07
refactor: Tidy up `createNewThread()`

KevinBoulongne:
Depends on #2230

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/DefaultRefreshStrategy.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/DefaultRefreshStrategy.kt
@@ -138,6 +138,8 @@ interface DefaultRefreshStrategy : RefreshStrategy {
         // Because we are missing the links between Messages, it will create multiple Threads for the same Folder.
         // Hence, we need to find these duplicates, and remove them.
         val duplicatedThreads = identifyExtraDuplicatedThreads(remoteMessage.messageIds)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/DefaultRefreshStrategy.kt
```diff
@@ -138,50 +124,68 @@ interface DefaultRefreshStrategy : RefreshStrategy {
...
     ) {
-        if (existingThreads.isEmpty()) return
+        // Update already existing Threads (i.e. in other Folders, or specific cases like Snoozed)
+        impactedThreadsManaged += addAllMessagesToAllThreads(scope, remoteMessage, existingThreads, existingMessages)
+
+        // Some Messages don't have references to all previous Messages of the Thread (ex: these from the iOS Mail app).
+        // Because we are missing the links between Messages, it will create multiple Threads for the same Folder.
+        // Hence, we need to find these duplicates, and remove them.
+        val duplicatedThreads = identifyExtraDuplicatedThreads(remoteMessage.messageIds)
+        impactedThreadsManaged -= duplicatedThreads
```
LunarX, line 141:
```suggestion
       // We need to make sure to remove the duplicated threads even if they were previously added to `impactedThreadsManaged`. Later on, this set will be used to recompute threads and threads that are deleted from realm will crash realm if we try to access them
        impactedThreadsManaged -= duplicatedThreads
```

LunarX (approved):

LunarX (changes requested):
wrong approval

PR #2235, opened 2025-03-07
refactor: Clean code

KevinBoulongne:
Depends on #2233

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
@@ -97,7 +97,7 @@ class ThreadController @Inject constructor(
         filterFolder: Folder?,
     ): List<Thread> = withContext(ioDispatcher) {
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -97,7 +97,7 @@ class ThreadController @Inject constructor(
         filterFolder: Folder?,
     ): List<Thread> = withContext(ioDispatcher) {
 
-        fun MutableRealm.keepOldMessagesAndAddToSearchFolder(remoteThread: Thread, searchFolder: Folder) {
+        fun MutableRealm.keepOldMessagesAndAddToSearchFolder(remoteThread: Thread) {
```
LunarX, line 100: I don't think this method adds anything to the search folder, is the name still apropriate?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
KevinBoulongne: True, the `add` part was removed in Octobre 2024.

PR #2236, opened 2025-03-07
fix: Correctly update Message fields

KevinBoulongne:
Depends on #2240

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/CustomRefreshStrategies.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/CustomRefreshStrategies.kt
@@ -88,23 +88,8 @@ val snoozeRefreshStrategy = object : DefaultRefreshStrategy {
         impactedThreadsManaged: MutableSet<Thread>,
         realm: MutableRealm,
     ) {
-
-        MessageController.getMessage(remoteMessage.uid, realm)?.let { localMessage ->
-            remoteMessage.initLocalValues(
-                areHeavyDataFetched = localMessage.areHeavyDataFetched,
-                isTrashed = localMessage.isTrashed,
-                messageIds = localMessage.messageIds,
-                draftLocalUuid = localMessage.draftLocalUuid,
-                isFromSearch = localMessage.isFromSearch,
-                isDeletedOnApi = localMessage.isDeletedOnApi,
-                latestCalendarEventResponse = localMessage.latestCalendarEventResponse,
-                swissTransferFiles = localMessage.swissTransferFiles,
-            )
-            remoteMessage.keepHeavyData(localMessage)
-        }
-
+        MessageController.getMessage(remoteMessage.uid, realm)?.let(remoteMessage::keepLocalValues)
         val updatedMessage = MessageController.upsertMessage(remoteMessage, realm)
-
         impactedThreadsManaged += updatedMessage.threads
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/CustomRefreshStrategies.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/CustomRefreshStrategies.kt
@@ -21,7 +21,6 @@ import android.content.Context
 import com.infomaniak.mail.data.models.Folder
 import com.infomaniak.mail.data.models.mailbox.Mailbox
 import com.infomaniak.mail.data.models.message.Message
-import com.infomaniak.mail.data.models.message.Message.MessageInitialState
 import com.infomaniak.mail.data.models.thread.Thread
 import io.realm.kotlin.MutableRealm
 import io.realm.kotlin.TypedRealm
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/CustomRefreshStrategies.kt
```diff
@@ -92,13 +91,14 @@ val snoozeRefreshStrategy = object : DefaultRefreshStrategy {
 
         MessageController.getMessage(remoteMessage.uid, realm)?.let { localMessage ->
             remoteMessage.initLocalValues(
-                messageInitialState = MessageInitialState(
-                    isFullyDownloaded = localMessage.isFullyDownloaded(),
-                    isTrashed = localMessage.isTrashed,
-                    isFromSearch = localMessage.isFromSearch,
-                    draftLocalUuid = localMessage.draftLocalUuid,
-                ),
+                isFullyDownloaded = localMessage.isFullyDownloaded(),
```
LunarX, line 94: Why can't we simply take the localMessage._isFullyDownloaded directly in this case? We don't need to compute anything right? We only need to take the existing localMessage as is, nothing more, nothing less

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -108,13 +107,14 @@ class ThreadController @Inject constructor(
...
                 // which is the reason why we can compute the `isTrashed` value so loosely.
                 remoteMessage.initLocalValues(
-                    MessageInitialState(
-                        isFullyDownloaded = localMessage?.isFullyDownloaded() ?: false,
-                        isTrashed = filterFolder?.role == FolderRole.TRASH,
-                        isFromSearch = localMessage == null,
-                        draftLocalUuid = localMessage?.draftLocalUuid,
-                    ),
-                    latestCalendarEventResponse = null,
+                    isFullyDownloaded = localMessage?.isFullyDownloaded() ?: false,
```
LunarX, line 110: Same question here about `isFullyDownloaded()` when we have a localMessage to keep local values from

LunarX (changes requested):
Also we should be able to a `fun keepLocalValues(localMessage: Message)` as follows and only call it when adding new message in the snooze strategy and when we already have a local message in the search. The reste of the time we can call the already existing `initLocalValues()`

```kt
fun keepLocalValues(localMessage: Message) {
    initLocalValues(
        isFullyDownloaded = localMessage._isFullyDownloaded,
        isTrashed = localMessage.isTrashed,
        messageIds = localMessage.messageIds,
        draftLocalUuid = localMessage.draftLocalUuid,
        isFromSearch = localMessage.isFromSearch,
        isDeletedOnApi = localMessage.isDeletedOnApi,
        latestCalendarEventResponse = localMessage.latestCalendarEventResponse,
        swissTransferFiles = localMessage.swissTransferFiles,
    )
    keepHeavyData(localMessage)
}
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -346,14 +346,13 @@ class ThreadController @Inject constructor(
...
-                                    isFromSearch = localMessage.isFromSearch,
-                                    draftLocalUuid = remoteMessage.getDraftLocalUuid(realm),
-                                ),
-                                latestCalendarEventResponse = localMessage.latestCalendarEventResponse,
+                                isFullyDownloaded = true,
+                                isTrashed = localMessage.isTrashed,
                                 messageIds = localMessage.messageIds,
+                                draftLocalUuid = remoteMessage.getDraftLocalUuid(realm),
+                                isFromSearch = localMessage.isFromSearch,
+                                isDeletedOnApi = false,
```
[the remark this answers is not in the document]
KevinBoulongne: It's a local value, and we should initialize them all in `initLocalValues()`.
We forgot to do it for this one.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/CustomRefreshStrategies.kt
KevinBoulongne: Actually, we can.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -346,7 +348,7 @@ class ThreadController @Inject constructor(
                             } ?: realmListOf()
 
                             remoteMessage.initLocalValues(
-                                isFullyDownloaded = true,
+                                areHeavyDataFetched = true,
                                 isTrashed = localMessage.isTrashed,
                                 messageIds = localMessage.messageIds,
                                 draftLocalUuid = remoteMessage.getDraftLocalUuid(realm),
```
LunarX, line 354: Can we spare a realm call if we simply read the localMessage.draftLocalUuid or does it matter here to recompute it?

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -284,7 +299,7 @@ class Message : RealmObject {
         ) {
             false
         } else {
-            _isFullyDownloaded
+            areHeavyDataFetched
```
LunarX, line 302: Would we want to rename the method here to match the new `areHeavyDataFetched` naming as well?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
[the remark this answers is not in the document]
LunarX: If we forget to add values to `initLocalValues()` we should create an embedded object with all local values of the Message or something similar so it will prevent us from compiling if we forget to define it inside `initLocalValues()`. Because I think that it's not the first time I've heard this issue of forgetting local values.

Nonetheless this is too big for this PR, but we should at least create a card to evaluate what we want to do about it or something

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
KevinBoulongne: No, the `areHeavyDataFetched` is just a part of the `isFullyDownloaded` logic.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
KevinBoulongne: No, because we've got no assurance that the localMessage will have the `draftLocalUuid`.
Since a Draft can have multiple Messages, we need to use the remote version to be sure we don't lose the data.

LunarX (approved):

PR #2243, opened 2025-03-11
feat: First snoozed thread UI

LunarX:
Only displays the little clock and correct date in the ThreadListFragment for now, as a first step. I stuck with the existing code but it's strange that the compact mode and the scheduled or snoozed folder all want to avoid displaying sections but they don't do it the same way


app/src/main/res/values-night/colors.xml
```diff
@@ -88,6 +88,9 @@
     <!-- Calendar events -->
     <color name="calendarHeaderBackground">@color/orca</color>
 
+    <!-- Snooze -->
+    <color name="snoozeColor">@color/coral_dark</color>
```
KevinBoulongne, line 92: It looks like it's unused.

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #2270, opened 2025-03-19
fix: Don't send a Sentry when a Thread is both in its correct Folder and in the Search folder

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -132,7 +132,13 @@ class Thread : RealmObject, Snoozable {
     // TODO: Remove this `runCatching / getOrElse` when the Threads parental issues are fixed
     val folder
         get() = runCatching {
-            _folders.single()
+            // The only situation were we can have more than 1 parent folder is if the parent folders
+            // are 2 with this exact situation : 1x any real folder and 1x the Search folder.
+            if (_folders.count() == 2) {
```
sirambd, line 137: Are you sure it's not possible to have more than two?

sirambd (changes requested):

PR #2272, opened 2025-03-20
fix: Directly use `folderRoleName` in `computeAvatarRecipient()`

KevinBoulongne:


LunarX (commented):
Why do we need to add redundancy to the folder role inside a Message? Reading it from folder.role isn't enough? 

Also, your modification literally replaces folder.role with folder.role, it's just read sooner, does it really changes anything anyway?

KevinBoulongne:
> Why do we need to add redundancy to the folder role inside a Message? Reading it from folder.role isn't enough?
> 
> Also, your modification literally replaces folder.role with folder.role, it's just read sooner, does it really changes anything anyway?

We've got issues when accessing to backlinks in the UI, specifically when trying to compute avatars _(probably avatars because it's one of the 1st things we do, so it's here that it crashes 1st)_.
Directly having this data in the model will enable us to not use backlinks when accessing data from the UI.

tevincent (approved):

PR #2278, opened 2025-03-24
feat: Deserialize snoozeAction into snoozeUuid directly

LunarX:
The api will always return the snooze uuid in the form of a resource because the web needs it this way. Instead of sending it a second time in another field, mobile will simply deserialize it from the resource field.

This PR does just this, it adds an intermediary serializer that will extract the snooze uuid out of the snooze action resource. This way we can easily interface with snooze uuid inside the app


app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
```diff
@@ -156,3 +158,28 @@ private fun MigrationContext.replaceOriginalDateWithDisplayDateAfterTwentyFourth
     }
 }
 //endregion
+
+// Migrate from version #25
+private fun MigrationContext.deserializeSnoozeUuidDirectlyAfterTwentyFifthsMigration() {
+
+    if (oldRealm.schemaVersion() <= 25L) {
+        enumerate(className = "Message") { oldObject: DynamicRealmObject, newObject: DynamicMutableRealmObject? ->
+            newObject?.apply {
```
tevincent, line 167: Is that apply really necessary ?

app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
```diff
@@ -156,3 +158,28 @@ private fun MigrationContext.replaceOriginalDateWithDisplayDateAfterTwentyFourth
...
+                // Initialize new property with old property value
+                val snoozeAction = oldObject.getNullableValue<String>(fieldName = "snoozeAction")
+                val snoozeUuid = snoozeAction?.lastUuidOrNull()
+                set(propertyName = "snoozeUuid", value = snoozeUuid)
+            }
+        }
+
+        enumerate(className = "Thread") { oldObject: DynamicRealmObject, newObject: DynamicMutableRealmObject? ->
+            newObject?.apply {
```
tevincent, line 176: Is that apply really necessary ?

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
LunarX: I was about to say "no, it's not necessary" but I just realized that it prevents us from computing anything if newObject is `null`. So yeah it has a slight purpose I guess

tevincent (approved):

PR #2279, opened 2025-03-24
feat: Enable cache of WebViews

tevincent:
With cache: 

https://github.com/user-attachments/assets/f6549d3b-1145-459c-9cba-8fa8fcae3a4a

Without cache: 

https://github.com/user-attachments/assets/ce185564-cf1d-4df2-bdf4-a90275cd6680

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
@@ -21,6 +21,7 @@ import android.annotation.SuppressLint
 import android.content.Context
 import android.webkit.JavascriptInterface
 import android.webkit.WebSettings
+import android.webkit.WebSettings.LOAD_CACHE_ELSE_NETWORK
 import android.webkit.WebView
 import com.infomaniak.mail.R
 import com.infomaniak.mail.utils.HtmlFormatter.Companion.getCustomDarkMode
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -130,21 +131,26 @@ class WebViewUtils(context: Context) {
             jsBridge = JavascriptBridge(onWebViewFinishedLoading = onWebViewFinishedLoading)
         }
 
-        fun WebSettings.setupThreadWebViewSettings() {
+        private fun WebSettings.setupDefaultWebViewSettings() {
```
LunarX, line 134: Maybe we should use the term "common" over "default". These settings are not "default" per say. If we create a 3rd webview somewhere but don't need some of these settings we should not default to them without thinking just because they're the "default" ones

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -130,21 +131,26 @@ class WebViewUtils(context: Context) {
...
             displayZoomControls = false
         }
 
         fun WebSettings.setupNewMessageWebViewSettings() {
-            @SuppressLint("SetJavaScriptEnabled")
-            javaScriptEnabled = true
+            setupDefaultWebViewSettings()
 
             loadWithOverviewMode = true
             useWideViewPort = true
```
LunarX, line 156: These two are already defined inside `setupDefaultWebViewSettings()`. No need to define them again

LunarX (changes requested):

LunarX (changes requested):
Doesn't work for all images :/. Example shown directly to @tevincent

LunarX (approved):
The images that would not be cached had a cache duration of 0 when retrieving it from the api call. So this is just the expected behavior. Everything works correctly then

PR #2280, opened 2025-03-25
feat: Unsnooze messages and threads on the api when they receive a new message

LunarX:
This PR reproduces the hack that the webmail has to do to let snoozed threads with new messages be unsnoozed and reappear in the inbox.

When refreshing folders, we need to detect snoozed threads that have their last message belonging to inbox not snoozed. This state is the state we need to fix, by making an "unsnooze" api call so the thread will come back into inbox.

If the api doesn't lets us unsnooze this specific uuid, the thread could be detected on every refresh and try to be unsnoozed forever, so we take the party to manually unsnooze it and its snoozed messages in the database in this strange case.

Also, the migration for `isLastInboxMessageSnoozed` might be overkill, initializing to `false` would only mean that existing threads in this very rare state won't be automatically unsnoozed by the hack. But they would later be unsnoozed if the thread is recomputed

Depends on #2293

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
@@ -345,6 +345,10 @@ object ApiRepository : ApiRepositoryCore() {
         return callApi(ApiRoutes.blockUser(mailboxUuid, folderId, shortUid), POST)
     }
 
+    fun unsnoozeThread(mailboxUuid: String, snoozeUuid: String): ApiResponse<Boolean> {
+        return callApi(ApiRoutes.snoozeAction(mailboxUuid, snoozeUuid), DELETE)
+    }
+
     fun unsnoozeThreads(mailboxUuid: String, snoozeUuids: List<String>): List<ApiResponse<Unit>> {
         return batchOver(snoozeUuids, limit = Utils.MAX_UUIDS_PER_CALL_SNOOZE_DELETE) {
             callApi(ApiRoutes.snooze(mailboxUuid), DELETE, mapOf("uuids" to it))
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/infomaniak/mail/data/api/ApiRoutes.kt
```diff
@@ -277,6 +277,16 @@ object ApiRoutes {
     }
     //endregion
 
+    //region Snooze
+    fun snooze(mailboxUuid: String): String {
+        return "${mailMailbox(mailboxUuid)}/snoozes"
+    }
+
+    fun snoozeAction(mailboxUuid: String, snoozeUuid: String): String {
```
KevinBoulongne, line 285:
```suggestion
    fun unsnoozeThread(mailboxUuid: String, snoozeUuid: String): String {
```

app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -345,6 +345,16 @@ object ApiRepository : ApiRepositoryCore() {
         return callApi(ApiRoutes.blockUser(mailboxUuid, folderId, shortUid), POST)
     }
 
+    fun unsnoozeThread(mailboxUuid: String, snoozeUuid: String): ApiResponse<Boolean> {
+        return callApi(ApiRoutes.snoozeAction(mailboxUuid, snoozeUuid), DELETE)
```
KevinBoulongne, line 349:
```suggestion
        return callApi(ApiRoutes.unsnoozeThread(mailboxUuid, snoozeUuid), DELETE)
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/data/api/ApiRoutes.kt
LunarX: Technically speaking, when only considering the "route" it has nothing to do with unsnoozing. The route truly is what the API calls a "snooze action". This route can be used with a DELETE to unsnooze but it can be used with a PUT to update the snooze date

KevinBoulongne (approved):

PR #2282, opened 2025-03-26
feat: Hide currently scheduled date when prompt the user to choose another scheduled date

LunarX:
When rescheduling a draft, the already selected date is displayed and can be selected again. This PR hides it. It's also going to be useful for snoozed messages when the scheduling feature will come

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -65,6 +65,7 @@ import dagger.hilt.android.lifecycle.HiltViewModel
 import io.realm.kotlin.Realm
 import io.realm.kotlin.ext.toRealmList
 import io.realm.kotlin.notifications.ResultsChange
+import io.realm.kotlin.types.RealmInstant
 import io.sentry.Attachment
 import io.sentry.Sentry
 import io.sentry.SentryLevel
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -65,6 +65,7 @@ import dagger.hilt.android.lifecycle.HiltViewModel
 import io.realm.kotlin.Realm
 import io.realm.kotlin.ext.toRealmList
 import io.realm.kotlin.notifications.ResultsChange
+import io.realm.kotlin.types.RealmInstant
```
tevincent, line 68:
```suggestion
```

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -90,6 +90,9 @@ class ThreadViewModel @Inject constructor(
         AccountUtils.currentMailboxId,
     ).map { it.obj }.asLiveData(ioCoroutineContext)
 
+    // Save the current scheduled date of the draft we're rescheduling to be able to pass it to the schedule bottom sheet
+    var reschedulingCurrentlyScheduledEpoch: Long? = null
```
tevincent, line 94: Should we name this like `...Millis` ?

tevincent (changes requested):

PR #2288, opened 2025-03-28
refactor: Don't make api call outside of the repository

NicolasBourdin88:


app/src/main/java/com/infomaniak/mail/utils/extensions/AttachmentExt.kt
```diff
@@ -129,42 +123,32 @@ object AttachmentExt {
...
-            .build()
-
-        val response = AccountUtils.getHttpClient(mailbox.userId).newCall(request).execute()
-
-        val apiResponse = ApiController.json.decodeFromString<ApiResponse<Attachment>>(response.body?.string() ?: "")
-        if (apiResponse.isSuccess() && apiResponse.data != null) {
+        val okHttpClient = AccountUtils.getHttpClient(mailbox.userId)
+        val apiResponse = ApiRepository.createAttachments(attachmentFile, mailbox, name, mimeType, userApiToken, okHttpClient)
+
+        if (apiResponse != null && apiResponse.isSuccess() && apiResponse.data != null) {
```
FabianDevel, line 138: apiResponse?.data != true should be enough
And either way you can merge the first two clauses like this : `apiResponse?.isSuccess() == true`

FabianDevel (changes requested):

FabianDevel (approved):

PR #2300, opened 2025-04-02
feat: Force sync calendar

tevincent:


app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -207,6 +211,9 @@ class MainActivity : BaseActivity() {
         observeDeletedMessages()
         observeDeleteThreadTrigger()
         observeDraftWorkerResults()
+
+        observeCanShowSyncDiscovery()
```
FabianDevel, line 215: Why the line feed ?

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -282,6 +290,14 @@ class MainActivity : BaseActivity() {
         }
     }
 
+    private fun observeCanShowSyncDiscovery() {
+        syncDiscoveryManager.syncDiscoveryViewModel.canShowSyncDiscovery.observe(this) { canShowSyncDiscovery ->
```
FabianDevel, line 294: I find it odd to expose the manager's viewModel
Maybe we should wrap the value inside the manager ?

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -282,6 +290,14 @@ class MainActivity : BaseActivity() {
...
+            if (canShowSyncDiscovery) {
+                showSyncDiscovery()
+            }
```
FabianDevel, line 297:
```suggestion
            if (canShowSyncDiscovery) showSyncDiscovery()
```

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
FabianDevel: Also, as the manager is a DefaultLifecycleObserver, we could directly put this observe inside it

FabianDevel (changes requested):

PR #2303, opened 2025-04-04
refactor: Random Snooze fixes

KevinBoulongne:
Depends on #2302


app/src/main/java/com/infomaniak/mail/ui/alertDialogs/SelectDateAndTimeForSnoozeDialog.kt
```diff
@@ -50,6 +50,6 @@ open class SelectDateAndTimeForSnoozeDialog @Inject constructor(
     )
 
     companion object {
-        const val MAX_SCHEDULE_DELAY_YEARS = 1
+        const val MAX_SNOOZE_DELAY_YEARS = 1
```
LunarX, line 53: That value still represents the maximum delay at which you can schedule a snooze, so the "schedule" term is nice to have but it can become `MAX_SNOOZE_SCHEDULE_DELAY_YEARS` if you find it clearer that way

LunarX (changes requested):

LunarX (approved):

PR #2315, opened 2025-04-09
feat: When fetching new Messages, display newest first

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -401,10 +401,14 @@ class RefreshController @Inject constructor(
         return impactedThreads
     }
 
+    private fun List<Int>.getAtStart(size: Int) = subList(0, size) to subList(size, count())
+
+    private fun List<Int>.getAtEnd(size: Int) = subList(count() - size, count()) to subList(0, count() - size)
+
```
LunarX, line 407: Maybe `amount` or `quantity` over `size`? `size` feels a bit vague but more importantly it's already a reserved name in the context of a list, so it can be more misleading than one of the alternatives? What do you think?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -401,10 +401,14 @@ class RefreshController @Inject constructor(
...
     private fun computeUids(folder: Folder, direction: Direction): Pair<List<Int>, List<Int>> {
         val allUids = if (direction == Direction.TO_THE_FUTURE) folder.newMessagesUidsToFetch else folder.oldMessagesUidsToFetch
         return if (allUids.count() > Utils.PAGE_SIZE) {
-            allUids.subList(0, Utils.PAGE_SIZE) to allUids.subList(Utils.PAGE_SIZE, allUids.count())
+            if (direction == Direction.TO_THE_FUTURE) allUids.getAtEnd(Utils.PAGE_SIZE) else allUids.getAtStart(Utils.PAGE_SIZE)
```
LunarX, line 411: We're certain that `allUids` field will always be filled in this very exact order? With oldest messages at the start and newest message at the end?

LunarX (changes requested):

LunarX (approved):

PR #2316, opened 2025-04-14
fix: Avoid realm migration crash when setting originalDate or reading snoozeAction

LunarX:
When migrating with a schema version inferior or equal to 22, during the 22th migration, the field `originalDate` of the new object is set but this field doesn't exist in the latest version of the model's schema anymore as of today. Trying to `set()` this non existing field crashes. Another crash occurs when trying to read `snoozeAction` that never was initialized.

There is no efficient and generic solution with a low memory footprint that can provide a permanent safeguard against this issue, while remaining transparent during migration writing.

Instead, the fix safeguards this very specific case of the error that's been raised and observed in production.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
@@ -119,7 +119,7 @@ private fun MigrationContext.initializeInternalDateAsDateAfterTwentySecondMigrat
                 set(propertyName = "internalDate", value = oldObject.getValue<RealmInstant>(fieldName = "date"))
 
                 // Initialize new property with old property value
-                setSafe(propertyName = "originalDate", value = oldObject.getValue<RealmInstant>(fieldName = "date"))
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
```diff
@@ -219,3 +238,31 @@ private fun MigrationContext.initIsLastInboxMessageSnoozedAfterTwentySeventhMigr
     }
 }
 //endregion
+
+/**
+ * If the property we're trying to set doesn't exist anymore in our model at the latest schema version, instead of crashing skip
+ * this property value.
+ */
+private fun DynamicMutableRealmObject.setSafe(propertyName: String, value: Any?) {
```
NicolasBourdin88: As saw together, you should change the name for 'setIfExist'

NicolasBourdin88 (changes requested):

NicolasBourdin88 (approved):

NicolasBourdin88 (approved):

sirambd (approved):

PR #2319, opened 2025-04-15
fix: Add some more Sentry logs to try to understand why a simple string check doesn't succeed when it should

KevinBoulongne:
The Sentry says that the Message has a Thread in the correct Folder.
The data is correct on Sentry.
Everything should be fine.

But the `message.folderId == thread.folderId` check doesn't succeed. Why ??? 😭


app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -228,20 +229,18 @@ class Message : RealmObject, Snoozable {
                     scope.setExtra("email", AccountUtils.currentMailboxEmail.toString())
                 }
 
-                correctFolder = (threads.firstOrNull { it.folderId == folderId } ?: threads.first()).folder
+                correctFolder = (threads.firstOrNull { it.folder.id == folderId } ?: threads.first()).folder
```
LunarX, line 232: Also, between `Thread.folder.id` and `Thread.folderId`, which one is the correct one that should be used and why?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
KevinBoulongne: Both are supposed to be ok, but things don't seem to really work here, so I'm trying stuff. 😭

LunarX: Ok but then you won't understand the reason why `it.folderId == folderId` wasn't working then, because the code will have changed and might not capture a message anymore.

I don't know how much better you expect this new solution to work but if you really want this version instead, maybe try to compute the old solution in parallel and send sentry about the old version even if the code fallsback to the new solution. Or/and, maybe send to sentry differences when you detect that `thread.folderId != thread.folder.id` for any thread

LunarX: Actually I reproduce the issue on master very easily when navigating the search. You should probably start looking though this first instead to understand the issue. No need to add these sentries, for now, to understand your issue.

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
LunarX, line 1: Also, I'm wondering but why do we search for the search folder id? This could never find anything because the `Thread.folderId` can never be set to the search folder id according to the code I think.

```kt
if (correctFolder == null) {
    computeFolderAndReason(FolderController.SEARCH_FOLDER_ID).let {
        correctFolder = it.first
        reason = it.second
    }
}
```

LunarX, line 1: On the same kind of topic, what exact folder are we looking for with this `folder` property? The folder inside of which this message instance can be found in realm or the folder that this message is inside on the api?

Because these two folders can differ, for example, during the search.

During the search, if a Message was never downloaded by the app beforehand, it would still be part of an existing folder on the API, but locally it would only be contained inside of the fake search folder doesn't it?

LunarX (changes requested):

PR #2333, opened 2025-04-24
refactor: Extract swipe action performing logic to its own file to reduce cognitive complexity

LunarX:
Sonar notified us that this method had too much cognitive complexity, but to not make the other PR noisy, I fixed the cognitive complexity here instead.

It also removes 100+ lines from the already very big ThreadListFragment, so it's an even better refactor.

Depends on #2332

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/folder/PerformSwipeActionManager.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/folder/PerformSwipeActionManager.kt
@@ -1,0 +1,160 @@
+/*
+ * Infomaniak Mail - Android
+ * Copyright (C) 2025 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/folder/PerformSwipeActionManager.kt
```diff
@@ -0,0 +1,160 @@
...
+     * The boolean return value is used to know if we should keep the Thread in
+     * the RecyclerView (true), or remove it when the swipe is done (false).
+     */
+    fun ThreadListFragment.performSwipeActionOnThread(
+        swipeAction: SwipeAction,
+        thread: Thread,
+        position: Int,
+        isPermanentDeleteFolder: Boolean,
+        mainViewModel: MainViewModel,
+        localSettings: LocalSettings,
```
tevincent, line 48: No need to have the `MainViewModel` and `LocalSettings` as parameters here since it's an extension of `ThreadListFragment`.

tevincent (changes requested):

tevincent (approved):

PR #2334, opened 2025-04-24
feat: Drop Folder when there is too much new Activities

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -309,19 +330,27 @@ class RefreshController @Inject constructor(
         }
     }
 
-    private suspend fun Realm.fetchActivities(scope: CoroutineScope, folder: Folder, previousCursor: String) {
+    private suspend fun Realm.fetchActivities(scope: CoroutineScope, folder: Folder, previousCursor: String): Int {
```
LunarX, line 333: We can't be returning meaningless integers like that, use an enum it's much more appropriate

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
KevinBoulongne: It's been updated.

tevincent (approved):

LunarX (approved):
I approve of the updated fetchActivities returned value even if an enum representing the result of the operation would have still been clearer than a boolean imo

PR #2335, opened 2025-04-24
fix: Block swipe actions during multi select

LunarX:
We can't simplify this code much to only compute enabled state in a single place because we need to know when local settings swipe actions have changed and if we want to do so, we should observer shared preferences which would require a few more lines of code. All in all, I went with this approach instead to avoid introducing too much complexity in the code


tevincent (approved):

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -277,11 +277,15 @@ class ThreadListFragment : TwoPaneFragment() {
     }
 
     private fun unlockSwipeActionsIfSet() = with(binding.threadsList) {
+        val isMultiSelectClosed = mainViewModel.isMultiSelectOn.not()
+
         val leftIsSet = localSettings.swipeLeft != SwipeAction.NONE
```
KevinBoulongne, line 282:
```suggestion
        val isLeftSet = localSettings.swipeLeft != SwipeAction.NONE
```

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -277,11 +277,15 @@ class ThreadListFragment : TwoPaneFragment() {
...
-        if (leftIsSet) enableSwipeDirection(DirectionFlag.LEFT) else disableSwipeDirection(DirectionFlag.LEFT)
+        val isLeftEnabled = leftIsSet && isMultiSelectClosed
```
KevinBoulongne, line 283:
```suggestion
        val isLeftEnabled = isLeftSet && isMultiSelectClosed
```

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -277,11 +277,15 @@ class ThreadListFragment : TwoPaneFragment() {
...
+        if (isLeftEnabled) enableSwipeDirection(DirectionFlag.LEFT) else disableSwipeDirection(DirectionFlag.LEFT)
 
         val rightIsSet = localSettings.swipeRight != SwipeAction.NONE
```
KevinBoulongne, line 286:
```suggestion
        val isRightSet = localSettings.swipeRight != SwipeAction.NONE
```

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -277,11 +277,15 @@ class ThreadListFragment : TwoPaneFragment() {
...
-        if (rightIsSet) enableSwipeDirection(DirectionFlag.RIGHT) else disableSwipeDirection(DirectionFlag.RIGHT)
+        val isRightEnabled = rightIsSet && isMultiSelectClosed
```
KevinBoulongne, line 287:
```suggestion
        val isRightEnabled = isRightSet && isMultiSelectClosed
```

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #2340, opened 2025-04-28
feat: Migrate to SDK 35

sirambd:


app/src/main/java/com/infomaniak/mail/utils/extensions/WindowInsetsExt.kt
```diff
@@ -0,0 +1,49 @@
...
+        WindowInsetsCompat.CONSUMED
+    }
+}
+
+fun View.updatePaddingWithStatusBar(insets: WindowInsetsCompat) {
+    val statusBar = insets.statusBar()
+    updatePadding(top = statusBar.top, left = statusBar.left, right = statusBar.right, bottom = statusBar.bottom)
+}
+
+fun View.updatePaddingWithSystemBars(insets: WindowInsetsCompat) {
```
FabianDevel, line 38: It's a bit confusing that the "withSystemBars" doesn't  add top padding, maybe we should find another name

FabianDevel (changes requested):

FabianDevel (approved):

PR #2346, opened 2025-04-30
feat: Sdk 35 secondary activities

FabianDevel:
Depends on https://github.com/Infomaniak/android-core/pull/342


app/src/main/java/com/infomaniak/mail/ui/sync/SyncOnboardingFragment.kt
sirambd, line 1: Do the same here as for the intro to handle the edge-to-edge

sirambd (changes requested):

sirambd (approved):

PR #2347, opened 2025-04-30
fix: Add missing folder name tags in search results

LunarX:
Some folder name tags were missing from the search when fetching API threads on some occasions and they were always missing from the local realm search. This PR fixes this issue by introducing some shared Thread processing between the API and Realm searches. The shared processing is only comprised of the folder name computation for now.

To put this processing in common, I had to move around some of the current processing. I managed to regroup all search threads processing in the same SearchUtils file. I also decoupled the place where threads state is processed from the place where threads are saved to Realm as "search results"

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/search/SearchViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/search/SearchViewModel.kt
@@ -218,8 +218,8 @@ class SearchViewModel @Inject constructor(
         if (isFirstPage && isLastPage) searchUtils.deleteRealmSearchData()
 
         if (apiResponse.isSuccess()) {
-            createSearchThreadsFromRemote(apiResponse.data?.threads, folder)
             with(apiResponse) {
+                data?.let { createSearchThreadsFromRemote(it.threads, folder) }
                 resourceNext = data?.resourceNext
                 isFirstPage = data?.resourcePrevious == null
             }
@@ -240,12 +240,10 @@ class SearchViewModel @Inject constructor(
         visibilityMode.postValue(resultsVisibilityMode)
     }
 
-    private suspend fun createSearchThreadsFromRemote(apiThreads: List<Thread>?, folder: Folder?) {
+    private suspend fun createSearchThreadsFromRemote(remoteThreads: List<Thread>, folder: Folder?) {
         runCatching {
-            apiThreads?.let { remoteThreads ->
-                val threads = searchUtils.convertRemoteThreadsToSearchThreads(remoteThreads, folder)
-                threadController.saveSearchThreads(threads)
-            }
+            val searchThreads = searchUtils.convertRemoteThreadsToSearchThreads(remoteThreads, folder)
+            threadController.saveSearchThreads(searchThreads)
         }.getOrElse { exception ->
             exception.printStackTrace()
             Sentry.captureException(exception)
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/search/SearchViewModel.kt
```diff
@@ -226,18 +213,18 @@ class SearchViewModel @Inject constructor(
         val searchFilters = searchUtils.searchFilters(query, newFilters, resource)
         val apiResponse = ApiRepository.searchThreads(currentMailbox.uuid, folderId, searchFilters, resource)
 
-        searchJob?.ensureActive()
+        currentCoroutineContext().ensureActive()
 
         if (isFirstPage && isLastPage) searchUtils.deleteRealmSearchData()
 
         if (apiResponse.isSuccess()) {
+            createSearchThreadsFromRemote(apiResponse.data?.threads, folder)
```
tevincent, line 221: instead of having a `apiThreads: List<Thread>?` in `createSearchThreadsFromRemote`, we can check in the `if` if the apiResponse.data is not null ? And in that case, we can fallback to the local search ?

LunarX: If we receive a response from the api we are commited to the doing a remote search and not a local one so we should never fallback to local in this case imo.

I'm not exactly sure if we can really receive a null list by the way, when nothing matches the search query for exemple, we receive an empty list but not a null one.

LunarX: Which means we can always default to an empty list instead of defaulting to a null list. Also this is not even a realm object, which makes the modification even easier. I checked with @sirambd and he agrees with changing the `threads` typing

app/src/main/java/com/infomaniak/mail/utils/SearchUtils.kt
```diff
@@ -80,17 +89,91 @@ class SearchUtils @Inject constructor(
...
+                isFromSearch = true
+                setFolderId(filterFolder)
+                keepOldMessagesData(filterFolder, mailboxContentRealm())
+                sharedThreadProcessing(appContext, cachedFolderNames, realm = mailboxContentRealm())
             }
         }
     }
+
+    private fun Thread.setFolderId(filterFolder: Folder?) {
+        this.folderId = if (messages.count() == 1) messages.single().folderId else filterFolder!!.id
```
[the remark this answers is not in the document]
LunarX: I only moved this code around without modifying the logic but yeah this is not the prettiest of code.

When reading it carefully I see why this works though. The api will return threads of multiple messages only in the case were we've provided a filterFolder and nothing else to the search api call.

This strange way of writing this bit of logic could certainly be avoided by using sealed classes though

app/src/main/java/com/infomaniak/mail/utils/SearchUtils.kt
```diff
@@ -80,17 +89,91 @@ class SearchUtils @Inject constructor(
...
+
+/**
+ * Thread processing that applies to both search threads from the api and from realm. Be careful, it relies on
+ * [Thread.folderId] being set correctly.
+ */
+private fun Thread.sharedThreadProcessing(context: Context, cachedFolderNames: MutableMap<String, String>, realm: Realm) {
+    setFolderName(cachedFolderNames, realm, context)
+}
+
+private fun Thread.setFolderName(cachedFolderNames: MutableMap<String, String>, realm: Realm, context: Context) {
```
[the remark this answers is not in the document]
LunarX: They do not depend on the SearchUtils instance's state, as a matter of fact they are entirely "stateless" or lowly coupled if it makes sense, and they are also private. 

Defining them outside of SearchUtils highlights their level of low coupling and statelessness. This only works as long as they're private otherwise it would make no sens to access them without any form of namespace.

tevincent (approved):

PR #2364, opened 2025-05-13
fix(Api35): Fix sync activity for api 25

FabianDevel:


app/src/main/java/com/infomaniak/mail/ui/login/LoginFragment.kt
```diff
@@ -181,6 +188,7 @@ class LoginFragment : Fragment() {
 
         animateColorChange(oldSecondaryBackground, newSecondaryBackground) { color ->
             requireActivity().window.statusBarColor = color
```
LunarX, line 190: Should you not remove the `statusBarColor` right above? It's not used anymore now right?

LunarX (approved):

PR #2385, opened 2025-05-22
feat: Display computed emoji reactions per message

LunarX:
Depends on https://github.com/Infomaniak/android-kMail/pull/2384

PR #2386, opened 2025-05-22
feat: Hide reaction messages

LunarX:
When a reaction message targets an existing message, it needs to not display and not be taken into account in some places in the code.

To hide the messages we don't want I replaced accesses to `Thread.messages` that should only consider a subset of messages with the method `getDisplayMessages()` that will adapt and return the correct list of messages everytime it's called.

Depends on https://github.com/Infomaniak/android-kMail/pull/2385

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
@@ -18,6 +18,7 @@
 package com.infomaniak.mail.data.cache.mailboxContent
 
 import android.content.Context
+import com.infomaniak.mail.data.LocalSettings
 import com.infomaniak.mail.data.cache.RealmDatabase
 import com.infomaniak.mail.data.models.correspondent.Recipient
 import com.infomaniak.mail.data.models.mailbox.Mailbox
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -57,7 +65,7 @@ class MessageController @Inject constructor(private val mailboxContentRealm: Rea
         return getMessage(uid, mailboxContentRealm())
     }
 
-    fun getLastMessageToExecuteAction(thread: Thread): Message = with(thread) {
+    fun getLastMessageToExecuteAction(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message = with(thread) {
```
KevinBoulongne:
```suggestion
    fun getLastMessageToExecuteAction(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message {
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -72,16 +80,17 @@ class MessageController @Inject constructor(private val mailboxContentRealm: Rea
                 " AND \$recipient.${Recipient::email.name} ENDSWITH '${end}'" +
                 ").@count < 1"
 
+        val messages = getDisplayedMessages(featureFlags, localSettings)
```
KevinBoulongne:
```suggestion
        val messages = thread.getDisplayedMessages(featureFlags, localSettings)
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/thread/Thread.kt
```diff
@@ -203,13 +210,14 @@ class Thread : RealmObject, Snoozable {
     fun recomputeThread(realm: MutableRealm? = null) {
 
         messages.sortBy { it.internalDate }
+        val allMessages = messages
```
sirambd: Explain it with a comment

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -136,26 +146,20 @@ class ThreadViewModel @Inject constructor(
...
-        AccountUtils.currentUserId,
-        AccountUtils.currentMailboxId,
-    ).map { it.obj }.asLiveData(ioCoroutineContext)
-
     // Save the current scheduled date of the draft we're rescheduling to be able to pass it to the schedule bottom sheet
     var reschedulingCurrentlyScheduledEpochMillis: Long? = null
 
-    val isThreadSnoozeHeaderVisible: LiveData<ThreadHeaderVisibility> = Utils.waitInitMediator(currentMailboxLive, threadLive)
+    val isThreadSnoozeHeaderVisible: LiveData<ThreadHeaderVisibility> = Utils
+        .waitInitMediator(currentMailboxFlow.asLiveData(), threadLive)
```
sirambd: Keep the livedata instance on a shared local value

app/src/main/java/com/infomaniak/mail/utils/SearchUtils.kt
```diff
@@ -127,7 +127,8 @@ class SearchUtils @Inject constructor(
     }
 
     private fun Thread.setFolderId(filterFolder: Folder?) {
-        this.folderId = if (messages.count() == 1) messages.single().folderId else filterFolder!!.id
+        val allMessages = messages
```
sirambd: I think that it's useless here

PR #2395, opened 2025-06-03
feat: Add emoji picker inside of a bottom sheet

LunarX:
Uses emoji2 to add an emoji picker when the "add reaction" button is clicked. For now, the emoji only notifies of the selected emoji but it's not yet sent via an api call


app/src/main/res/layout/fragment_thread.xml
```diff
@@ -174,4 +174,11 @@
         tools:title="@string/noConversationSelected"
         tools:visibility="gone" />
 
+    <androidx.emoji2.emojipicker.EmojiPickerView
+        android:layout_gravity="bottom"
+        android:id="@+id/emojiPicker"
+        android:visibility="gone"
+        android:layout_width="match_parent"
+        android:layout_height="300dp" />
+
```
sirambd: Remove it because it's useless

app/src/main/java/com/infomaniak/mail/utils/extensions/NavigationExt.kt
```diff
@@ -67,6 +76,10 @@ fun NavController.animatedNavigation(@IdRes resId: Int, args: Bundle? = null, cu
     if (canNavigate(currentClassName, currentClassName)) navigate(resId, args, getAnimatedNavOptions())
 }
 
+fun Fragment.animatedNavigationSheet(@IdRes resId: Int, substituteClassName: String) {
+    safelyNavigate(resId = resId, substituteClassName = substituteClassName, navOptions = getSheetAnimatedNavOptions())
+}
+
```
sirambd: Remove it if useless because it's not called

sirambd (changes requested):

PR #2410, opened 2025-06-13
refactor: Remove unreachable condition and avoid main view model dependence

LunarX:
Inside the main view model's `getActionFolderRole()` method, the condition `threads?.firstOrNull()?.folderId == FolderController.SEARCH_FOLDER_ID` can never be met because `folderId` comes from the API and therefore cannot ever be equal to our local `FolderController.SEARCH_FOLDER_ID`. I removed the check, which renders the logic of the method pretty straight forward.

The new logic is so straight forward that, in a lot of cases, there's no need to depend on the view model for this logic. It even introduces unnecessary cognitive complexity to be able to make sense of what is really computed. I kept all the calls centralized in a unique point by introducing the `Thread.folderRole` property.

I had to rename the existing `folderRole` inside the ThreadListAdapter to not clash with the `Thread.folderRole` when we have a Thread as receiver in a scope

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
@@ -1322,14 +1322,6 @@ class MainViewModel @Inject constructor(
         isDownloadingChanges.postValue(false)
     }
 
-    fun getActionFolderRole(thread: Thread?): FolderRole? {
-        return getActionFolderRole(thread?.let(::listOf))
-    }
-
-    fun getActionFolderRole(message: Message): FolderRole? {
-        return getActionFolderRole(message.threads, message)
-    }
-
 // TODO: Handle this correctly if MultiSelect feature is added in the Search.
     /**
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -1339,15 +1331,8 @@ class MainViewModel @Inject constructor(
...
-            }
-            threads?.firstOrNull()?.folderId == FolderController.SEARCH_FOLDER_ID -> {
-                folderController.getFolder(threads.first().folderId)?.role
-            }
-            else -> {
-                threads?.firstOrNull()?.folder?.role
-            }
+            message != null -> message.folder.role
+            else -> threads?.firstOrNull()?.folderRole
         }
```
KevinBoulongne, line 1336: It looks like it won't get the correct folder role when the thread is from the Search.

KevinBoulongne (changes requested):

LunarX:
We kept the dependance on realm and made the change on another PR

PR #2413, opened 2025-06-17
fix: Make `getActionFolderRole()` use the correct Folder in all cases

KevinBoulongne:
Depends on #2417

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
@@ -117,17 +117,6 @@ object RealmDatabase {
             }
         }
     }
-
-    // Used for tests only.
-    class TestMailboxContent : MailboxContent() {
-        override operator fun invoke() = runBlocking(Dispatchers.IO) {
-            mailboxContentMutex.withLock {
-                _mailboxContent ?: newMailboxContentInstance(
-                    userId = -42, mailboxId = -1337, loadDataInMemory = true,
-                ).also { _mailboxContent = it }
-            }
-        }
-    }
     //endregion
 
     //region Close Realms
--- a/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
@@ -110,10 +110,21 @@ object RealmDatabase {
         }
     }.getOrThrow()
 
-    class MailboxContent {
-        operator fun invoke() = runBlocking(Dispatchers.IO) {
+    open class MailboxContent {
+        open operator fun invoke() = runBlocking(Dispatchers.IO) {
             mailboxContentMutex.withLock {
                 _mailboxContent ?: newMailboxContentInstance.also { _mailboxContent = it }
+            }
+        }
+    }
+
+    // Used for tests only.
+    class TestMailboxContent : MailboxContent() {
+        override operator fun invoke() = runBlocking(Dispatchers.IO) {
+            mailboxContentMutex.withLock {
+                _mailboxContent ?: newMailboxContentInstance(
+                    userId = -42, mailboxId = -1337, loadDataInMemory = true,
+                ).also { _mailboxContent = it }
             }
         }
     }
--- a/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
@@ -87,8 +87,8 @@ object RealmDatabase {
     val mailboxInfo get() = openRealmOrDropDbAndReboot(RealmConfig.mailboxInfo)
 
     val newMailboxContentInstance get() = newMailboxContentInstance(AccountUtils.currentUserId, AccountUtils.currentMailboxId)
-    fun newMailboxContentInstance(userId: Int, mailboxId: Int): Realm {
-        return openRealmOrDropDbAndReboot(RealmConfig.mailboxContent(userId, mailboxId))
+    fun newMailboxContentInstance(userId: Int, mailboxId: Int, loadDataInMemory: Boolean = false): Realm {
+        return openRealmOrDropDbAndReboot(RealmConfig.mailboxContent(userId, mailboxId, loadDataInMemory))
     }
 
     /**
@@ -254,7 +254,7 @@ object RealmDatabase {
             .migration(MAILBOX_INFO_MIGRATION)
             .build()
 
-        fun mailboxContent(userId: Int, mailboxId: Int): RealmConfiguration {
+        fun mailboxContent(userId: Int, mailboxId: Int, loadDataInMemory: Boolean = false): RealmConfiguration {
 
             if (mailboxId == DEFAULT_ID) {
                 Sentry.captureMessage("RealmConfiguration problem with mailbox content, default ID is used.", SentryLevel.ERROR)
@@ -263,8 +263,13 @@ object RealmDatabase {
             return RealmConfiguration.Builder(mailboxContentSet)
                 .name(mailboxContentDbName(userId, mailboxId))
                 .schemaVersion(MAILBOX_CONTENT_SCHEMA_VERSION)
+                .loadDataInMemoryIfNeeded(loadDataInMemory)
                 .migration(MAILBOX_CONTENT_MIGRATION)
                 .build()
+        }
+
+        private fun RealmConfiguration.Builder.loadDataInMemoryIfNeeded(loadDataInMemory: Boolean): RealmConfiguration.Builder {
+            return apply { if (loadDataInMemory) inMemory() }
         }
         //endregion
     }
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -44,6 +44,8 @@ class FolderController @Inject constructor(
     private val mailboxContentRealm: RealmDatabase.MailboxContent,
 ) {
 
+    var testRealm: Realm? = null
+
     //region Get data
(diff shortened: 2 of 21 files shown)


app/src/main/java/com/infomaniak/mail/utils/FolderRoleUtils.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    private fun getActionFolderRole(
+        folderId: String,
+        snoozable: Snoozable,
+        folderController: FolderController,
+    ): FolderRole? {
+        val folderRole = folderController.getFolder(folderId)?.role
+        return if (folderRole == FolderRole.INBOX && snoozable.isSnoozed()) FolderRole.SNOOZED else folderRole
+    }
```
LunarX: I see you ended up changing your approach and you passed down the folderController through every method in the end instead of injecting the FolderRoleUtils through DI. What was the motivation behind this change?

app/src/main/java/com/infomaniak/mail/utils/FolderRoleUtils.kt
LunarX, line 1: We should definitely test the methods of this file because it's easier and more robust than trying to find a way to test everything inside of the app by hand in this case

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -87,8 +88,8 @@ import com.infomaniak.lib.core.R as RCore
 class MainViewModel @Inject constructor(
     application: Application,
     avatarMergedContactData: AvatarMergedContactData,
+    val folderController: FolderController,
```
LunarX: I wonder to what extent we want to expose a controller outside of a viewmodel and into views

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
```diff
@@ -44,6 +44,8 @@ class FolderController @Inject constructor(
     private val mailboxContentRealm: RealmDatabase.MailboxContent,
 ) {
 
+    var testRealm: Realm? = null
```
LunarX: We can check tomorrow but this should not be needed because you can pass the realm through the constructor in mailboxContentRealm, no? And I think that you already do it.

LunarX (changes requested):
I only quickly glanced at it but I'll look into it tomorrow

app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
```diff
@@ -110,13 +110,24 @@ object RealmDatabase {
...
+    // Used for tests only.
+    class TestMailboxContent : MailboxContent() {
+        override operator fun invoke() = runBlocking(Dispatchers.IO) {
+            mailboxContentMutex.withLock {
+                _mailboxContent ?: newMailboxContentInstance(
+                    userId = -42, mailboxId = -1337, loadDataInMemory = true,
+                ).also { _mailboxContent = it }
+            }
+        }
+    }
```
LunarX: We can extract it outside of the "real files" and into the test package

LunarX (changes requested):
first comment

app/src/main/java/com/infomaniak/mail/data/cache/RealmDatabase.kt
```diff
@@ -110,13 +110,24 @@ object RealmDatabase {
...
         }
     }
+
+    // Used for tests only.
+    class TestMailboxContent : MailboxContent() {
+        override operator fun invoke() = runBlocking(Dispatchers.IO) {
+            mailboxContentMutex.withLock {
+                _mailboxContent ?: newMailboxContentInstance(
+                    userId = -42, mailboxId = -1337, loadDataInMemory = true,
+                ).also { _mailboxContent = it }
```
LunarX: We should be able to call `RealmConfig.mailboxContent` directly and leave `newMailboxContentInstance` intact

app/src/main/java/com/infomaniak/mail/data/models/Folder.kt
```diff
@@ -56,7 +56,7 @@ class Folder : RealmObject, Cloneable {
     var path: String = ""
     var name: String = ""
     @SerialName("role")
-    private var _role: String? = null
+    var _role: String? = null // Don't edit this value, it's only public for tests.
```
LunarX: We could create a "annotation" package with the following annotation to use here

```kt
@RequiresOptIn(
    level = RequiresOptIn.Level.ERROR,
    message = "Don't access it outside of testing"
)
@Retention(AnnotationRetention.BINARY)
@Target(AnnotationTarget.PROPERTY)
annotation class TestOnly
```

app/src/test/java/com/infomaniak/mail/FolderRoleUtilsTest.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.junit.runner.RunWith
+import org.mockito.Mock
+import org.mockito.Mockito.mock
+import org.mockito.junit.MockitoJUnitRunner
+
+@RunWith(MockitoJUnitRunner::class)
+class FolderRoleUtilsTest {
+
+    @Mock
+    private val mockContext = mock<Context>()
```
LunarX: Also we should add a comment explaining that we don't need the context because it's only used in methods that we don't call during this test

app/src/test/java/com/infomaniak/mail/FolderRoleUtilsTest.kt
```diff
@@ -0,0 +1,95 @@
...
+    private val mailboxContentRealm = DummyMailboxContent()
+    private val folderController = FolderController(mockContext, mailboxContentRealm)
+    private val folderRoleUtils = FolderRoleUtils(folderController)
+
+    private fun setup() {
+        mailboxContentRealm().writeBlocking {
+            copyToRealm(folderInbox, UpdatePolicy.ALL)
+            copyToRealm(folderDraft, UpdatePolicy.ALL)
+        }
+    }
```
LunarX: There's usually a way for test to run a specific bit of code before each test, can we use it here?

app/src/test/java/com/infomaniak/mail/dataset/DummyThreads.kt
```diff
@@ -0,0 +1,41 @@
...
+import com.infomaniak.mail.dataset.DummyMessages.messageDraft
+import com.infomaniak.mail.dataset.DummyMessages.messageInbox
+import com.infomaniak.mail.dataset.DummyMessages.messageInboxSnoozed
+
+object DummyThreads {
+
+    private val mailboxContentRealm = DummyMailboxContent()
+
+    val threadInboxSnoozed = threadOf(messageInbox, messageInboxSnoozed)
+    val threadDraft = threadOf(messageDraft)
```
LunarX: Maybe we should also test how this works when dealing with "search" threads and their different edge-cases like a single message of any folder vs a single message of the snooze folder or a single message and the case were we can get threads of multiple messages at once

app/src/test/java/com/infomaniak/mail/dataset/DummyFolders.kt
```diff
@@ -0,0 +1,45 @@
...
+object DummyFolders {
+
+    const val FOLDER_INBOX_ID = "FOLDER_INBOX_ID"
+    const val FOLDER_DRAFT_ID = "FOLDER_DRAFT_ID"
+
+    val folderInbox = Folder().apply {
+        id = FOLDER_INBOX_ID
+        _role = FolderRole.INBOX.name
+
+        threadInboxSnoozed.folderId = id
```
LunarX: Does it make sense to set this thread property inside of the folder initialization?

app/src/test/java/com/infomaniak/mail/dataset/DummyThreads.kt
```diff
@@ -0,0 +1,40 @@
...
+package com.infomaniak.mail.dataset
+
+import com.infomaniak.mail.data.models.message.Message
+import com.infomaniak.mail.data.models.thread.Thread
+import com.infomaniak.mail.dataset.DummyMessages.messageDraft
+import com.infomaniak.mail.dataset.DummyMessages.messageInbox
+import com.infomaniak.mail.dataset.DummyMessages.messageInboxSnoozed
+
+object DummyThreads {
+    val threadInboxSnoozed = threadOf(messageInbox, messageInboxSnoozed)
```
LunarX: This `threadOf(messageInbox, messageInboxSnoozed)` situation is actually impossible in practice. A thread in inbox will always have either all of its inbox message be snoozed or all of them be not snoozed.

Non-snoozed messages in a snoozed thread will always be from another folder

app/src/test/java/com/infomaniak/mail/FolderRoleUtilsTest.kt
LunarX: Also, if we need a setup then don't we need an associated reciprocate tear down?

app/src/test/java/com/infomaniak/mail/FolderRoleUtilsTest.kt
```diff
@@ -0,0 +1,95 @@
...
+        assertTrue(folderRole == FolderRole.DRAFT)
+    }
+
+    @Test
+    fun threadInboxSnoozed_should_be_SNOOZED() {
+        setup()
+
+        val folderRole = folderRoleUtils.getActionFolderRole(threadInboxSnoozed)
+        assertTrue(folderRole == FolderRole.SNOOZED)
+    }
```
LunarX: We need a test of a snoozed thread that's inside of the snooze folder to correctly return that it's folder role is snooze

app/src/test/java/com/infomaniak/mail/dataset/DummyFolders.kt
LunarX: A snoozed thread cannot be inside the inbox folder, it will be inside the snooze folder

LunarX (approved):

PR #2415, opened 2025-06-17
fix: Logout if not user

sirambd:


app/src/main/java/com/infomaniak/mail/MainApplication.kt
```diff
@@ -269,7 +272,6 @@ open class MainApplication : Application(), ImageLoaderFactory, DefaultLifecycle
         @OptIn(ExperimentalCoroutinesApi::class)
         val userTokenFlow = AppSettingsController.getCurrentUserIdFlow()
             .mapLatest { id -> id?.let { AccountUtils.getUserById(it)?.apiToken } }
-            .filterNotNull()
```
tevincent, line 291: If we delete that line, it'll fail silently in `TokenAuthenticator`, meaning `TokenAuthenticator` will only say we cannot authenticate that specific call by returning `null`. Do we really want that? I don't think so?

We already discussed that issue with @LunarX

tevincent:
We also improved the `authenticate` method in this [PR](https://github.com/Infomaniak/android-core/pull/401).

app/src/main/java/com/infomaniak/mail/MainApplication.kt
sirambd: Already discussed with @LunarX and @tevincent, and it's all good

LunarX (approved):

PR #2424, opened 2025-06-23
fix: Blink mail

NicolasBourdin88:
**What was happening before:** 

Thread disappear  →`threadController.updateIsLocallyMovedOutStatus(true)`
Mail reappear  →`threadController.updateIsLocallyMovedOutStatus(false)`
Mail disappear  → `refreshFoldersAsync()`

**Now:** 

Thread disappear  →`threadController.updateIsLocallyMovedOutStatus(true)`
Mail reappear only if one the moving API call fail →`threadController.updateIsLocallyMovedOutStatus(false)`
Mail disappears only if it was affected by one of the failing calls, even though its own call succeeded  → `refreshFoldersAsync()`


TommyDL-Infomaniak (approved):

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -626,12 +623,9 @@ class MainViewModel @Inject constructor(
                 destinationFolderId = trashId,
                 callbacks = RefreshCallbacks(onStart = ::onDownloadStart, onStop = { onDownloadStop(threadsUids) }),
             )
-        } else if (isSwipe) {
-            // We need to make the swiped Thread come back, so we reassign the LiveData with Realm values
-            reassignCurrentThreadsLive()
```
KevinBoulongne, line 672: As this PR is already approved, I'm commenting just to block the merge, because I'd like to test it beforehand.

KevinBoulongne (approved):

PR #2431, opened 2025-06-26
fix: Magine in scroll

Muuushuuu:


tevincent (changes requested):
What is ... Magine ? Hard to understand what this PR is about.

Edit: Now I get it. I think you should name your PR a bit more precisely. That way, we don't need to ask you for more details.

tevincent (approved):

PR #2407, opened 2025-06-12
feat: Translate emoji reaction errors

LunarX:
Add the different strings translated by iOS to the emoji reactions errors and define them inside `InfomaniakCore.apiErrorCodes`

Depends on #2395

PR #2432, opened 2025-06-26
feat: Send emoji email

LunarX:
Adapts the existing code that lets us init a Draft instance and make it reusable so we can use it inside of the MainActivity when reacting with an emoji.

This PR already makes the api call and observe the updated state of the emoji reactions to update the UI but for now we have to wait on the api call returning to update the UI. Better UX comes in later PRs.

Depends on #2407


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -264,7 +257,10 @@ class NewMessageViewModel @Inject constructor(
             } else {
                 getExistingDraft(draftLocalUuid) ?: return@runCatching
             }
-        }.cancellable().onFailure(Sentry::captureException)
+        }.cancellable().onFailure {
+            it.printStackTrace()
+            Sentry.captureException(it)
```
sirambd: Replace it with a `SentryLog.e`

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -1276,7 +1285,36 @@ class MainViewModel @Inject constructor(
 
     //region Emoji reaction
     fun sendEmojiReply(emoji: String, messageUid: String) {
-        // TODO
+        viewModelScope.launch {
+            val previousMessage = messageController.getMessage(messageUid) ?: return@launch
```
sirambd: Rename to `targetMessage`

app/src/main/java/com/infomaniak/mail/utils/DraftInitManager.kt
sirambd, line 1: This file need a reformat

sirambd (approved):

sirambd (changes requested):
Fix sonar issues

sirambd (approved):

PR #2433, opened 2025-06-26
feat: Fake sent emojis locally

LunarX:
Fake emojis locally when they are clicked so the UX is quick and responsive. In the meanwhile the code will try to send the emoji for real and only undo the faked reaction if the api call has failed.

This required to introduce a MessageUi model because relying on the database model made it impossible to fake reactions at the UI level. MessageUi mainly wraps the existing Message instance so I don't have to refactor the whole app.

For now reactions are only faked locally but some buttons can be clicked even though they should not. This is handled in later PRs.

Depends on #2432

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
@@ -641,10 +641,10 @@ class ThreadFragment : Fragment() {
             runningWorkInfoLiveData.observe(viewLifecycleOwner) {
                 it.forEach { workInfo ->
                     workInfo.progress.let {
-                        val emojiSendResult = it.getSerializable<DraftsActionsWorker.EmojiSendResult>(EMOJI_SENT_STATUS) ?: return@forEach
+                        val emojiSendResult =
+                            it.getSerializable<DraftsActionsWorker.EmojiSendResult>(EMOJI_SENT_STATUS) ?: return@forEach
                         if (emojiSendResult.isSuccess.not()) {
-                            // TODO: Remove the local status of this emoji
-                            threadViewModel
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -615,10 +631,51 @@ class ThreadFragment : Fragment() {
...
+
+                    emojiSendResults.results.forEach { emojiSendResult ->
+                        undoFakeEmojiReply(emojiSendResult)
+                    }
+                }
+            }
+        }
+    }
+
+    private fun undoFakeEmojiReply(emojiSendResult: DraftsActionsWorker.EmojiSendResult) {
```
sirambd: Rename into `undoFakeEmojiReplyIfNeeded`

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -615,10 +631,51 @@ class ThreadFragment : Fragment() {
...
+                        .getSerializable<DraftsActionsWorker.EmojiSendResult>(EMOJI_SENT_STATUS) ?: return@forEach
+
+                    undoFakeEmojiReply(emojiSendResult)
+                }
+            }
+
+            // Listening to the draft results ensures we will get all of the possible emoji results and not miss any unlike when
+            // we listen to the worker's progress.
+            val treatedWorkInfoUuids = mutableSetOf<UUID>()
+            draftsActionsWorkerScheduler.getCompletedAndFailedInfoLiveData().observe(viewLifecycleOwner) {
```
sirambd: Rename into `getCompletedInfoLiveData` and rename the function with a success return to `succeess` instead of completed

PR #2449, opened 2025-07-02
feat: Add Sentry on malformed snooze or unsnooze

NicolasBourdin88:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -541,6 +545,20 @@ class RefreshController @Inject constructor(
...
     }
+
+    private fun reportMalformedSnoozeMessages(messages: List<Message>) {
+        messages.forEach { message ->
+            if (message.isSnoozeOrUnsnoozeMalformed()) {
+                Sentry.withScope { scope ->
+                    scope.level = SentryLevel.WARNING
+                    scope.setExtra("snoozeState", message.snoozeState?.apiValue)
+                    scope.setExtra("snoozeUuid", message.snoozeUuid)
+                    scope.setExtra("snoozeEndDate", message.snoozeEndDate?.toDate().toString())
```
LunarX: If you want to format the date in an easily readable format, use a format where the date's timezone is explicit

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -541,6 +545,20 @@ class RefreshController @Inject constructor(
...
+                    Sentry.captureMessage("Message contains malformed snooze or unsnooze information")
```
LunarX: It's either "Message contains malformed snooze information" or "Message contains malformed snoozed or unsnoozed information" but not a mix of the two. Same goes for the condition method's name

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -541,6 +545,20 @@ class RefreshController @Inject constructor(
...
+                }
```
LunarX: We should add a tag to easily see the proportion of errors on the `snoozed` state vs. errors on the `unsnoozed` state

LunarX (changes requested):

PR #2453, opened 2025-07-04
fix: Add margin for informationBlockView EdgeToEdge

Muuushuuu:
When you go to Account Management (in Settings), the small block at the bottom has no margin when you rotate the phone.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/settings/privacy/AccountManagementSettingsFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/settings/privacy/AccountManagementSettingsFragment.kt
@@ -23,14 +23,19 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import androidx.activity.result.contract.ActivityResultContracts.StartActivityForResult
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/settings/privacy/AccountManagementSettingsFragment.kt
```diff
@@ -78,6 +84,17 @@ class AccountManagementSettingsFragment : Fragment() {
         }
     }
 
+    private fun handleEdgeToEdge() = with(binding) {
+        ViewCompat.setOnApplyWindowInsetsListener(informationBlockView) { view, insets ->
+            with(insets.safeArea()) {
+                view.setMargins(
+                    bottom = context?.resources?.getDimension(R.dimen.marginStandard)?.toInt()
```
tevincent: We have methods to handle insets in the project like `applyWindowInsetsListener`. And also, you never use the insets returned by the listener and you apply a margin at the bottom every time ? I don't understand.

tevincent (approved):

PR #2459, opened 2025-07-09
feat(Encryption-08): Add encryption chips

FabianDevel:
Create a new style for contact chips to make them indicates the current encryption status of their recipient


app/src/main/java/com/infomaniak/mail/ui/newMessage/encryption/EncryptionMessageManager.kt
```diff
@@ -223,4 +223,26 @@ class EncryptionMessageManager @Inject constructor(
...
+            toField.unencryptableRecipients = it
+            ccField.unencryptableRecipients = it
+            bccField.unencryptableRecipients = it
+        }
+        encryptionPassword?.let {
+            toField.encryptionPassword = it
+            ccField.encryptionPassword = it
+            bccField.encryptionPassword = it
+        }
+    }
```
tevincent: What about this?


```suggestion
    private fun applyEncryptionStyleOnRecipientFields(
        isEncryptionActivated: Boolean? = null,
        unencryptableRecipients: Set<String>? = null,
        encryptionPassword: String? = null,
    ) = with(binding) {
        listOf(toField, ccField, bccField).forEach { field ->
            isEncryptionActivated?.let { field.isEncryptionActivated = it }
            unencryptableRecipients?.let { field.unencryptableRecipients = it }
            encryptionPassword?.let { field.encryptionPassword = it }
        }
    }
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/encryption/EncryptionPasswordFragment.kt
```diff
@@ -76,7 +92,21 @@ class EncryptionPasswordFragment : Fragment() {
         }
     }
 
+    private fun generatePassword(): String {
```
tevincent: Could be in the `EncryptionViewModel` ?

tevincent (changes requested):

tevincent (changes requested):
Image

If you have one partially encrypted contact and you click on "Change password" and go back, every contacts are in `fully encrypted` mode.

It happens only in "Don't keep activities". I don't know if we want to handle that.

tevincent (approved):

PR #2460, opened 2025-07-09
feat(Encryption-09): Clean encryption lock button view and add delay to loader

FabianDevel:
Depends on #2459 

Clean the code that was aiming to instanciate this view as chips, but is finally not used
Add a delay before displaying the loader to avoid having it blinking when we open a draft

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/encryption/EncryptionLockButtonView.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/encryption/EncryptionLockButtonView.kt
@@ -26,14 +26,9 @@ import androidx.annotation.ColorRes
 import androidx.annotation.DrawableRes
 import androidx.core.view.isGone
 import androidx.core.view.isVisible
+import com.infomaniak.lib.core.utils.Utils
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/encryption/EncryptionLockButtonView.kt
```diff
@@ -134,7 +108,21 @@ class EncryptionLockButtonView @JvmOverloads constructor(
...
+     */
+    private fun loadWithDelay() {
+        loadingJob = CoroutineScope(Dispatchers.Default).launch {
+            delay(200)
+            Dispatchers.Main {
+                binding.unencryptedRecipientLoader.isVisible = true
+                binding.unencryptableGroup.isGone = true
+            }
+        }
+    }
```
KevinBoulongne: Usually, to achieve that, we use a `Utils.createRefreshTimer()`.
But this way is fine with me too.

KevinBoulongne (changes requested):

KevinBoulongne (approved):

PR #2463, opened 2025-07-10
chore: Add a Sentry when the messageUid from a notification is not found

TommyDL-Infomaniak:


app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
```diff
@@ -122,6 +123,16 @@ class ProcessMessageNotificationsWorker @AssistedInject constructor(
...
+
+        checkIfNotificationMessageUidIsInRealm(messageUid, mailboxContentRealm)
+
+        return@withContext workerResult
+    }
+
+    private fun checkIfNotificationMessageUidIsInRealm(messageUid: String, mailboxContentRealm: Realm) {
+        if (MessageController.getMessage(messageUid, mailboxContentRealm) == null) {
+            Sentry.captureMessage("Unable to find the message with the corresponding messageUid from the notification.")
+        }
```
LunarX: This is to be expected if we did not successfully manage to download the updates. In this case there's no point in checking and there is no point in sending a sentry event

LunarX (changes requested):

app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
```diff
@@ -117,11 +122,29 @@ class ProcessMessageNotificationsWorker @AssistedInject constructor(
...
+            workerFailureReason = it
             Result.failure()
         }.also {
             if (!hasShownNotification) displayGenericNewMailsNotification()
             mailboxContentRealm.close()
         }
+
+        // We don't want to log to Sentry if the error is a common outcome.
+        when (workerFailureReason) {
+            is CancellationException,
```
LunarX: You can't check for CancellationException like that here because they are caught and thrown immediately. Is this an issue for you if the code returns a cancellation exception?

app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
```diff
@@ -103,8 +107,9 @@ class ProcessMessageNotificationsWorker @AssistedInject constructor(
         val mailboxContentRealm = RealmDatabase.newMailboxContentInstance(userId, mailbox.mailboxId)
         var hasShownNotification = false
 
-        return@withContext runCatching {
+        var workerFailureReason: Throwable? = null
 
+        val workerResult = runCatching {
```
LunarX: I just saw that there's no need to extract workerResult as an independent variable. There's already other processing that happen after the `runCatching` and before returning inside the `.also`. All these processing should be defined at the same place in the code

app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
```diff
@@ -103,8 +107,9 @@ class ProcessMessageNotificationsWorker @AssistedInject constructor(
         val mailboxContentRealm = RealmDatabase.newMailboxContentInstance(userId, mailbox.mailboxId)
         var hasShownNotification = false
 
-        return@withContext runCatching {
+        var workerFailureReason: Throwable? = null
```
LunarX: If we keep this variable it doesn't hold a "reason" but rather a "throwable". I'd say a name along the lines of `workerFailureThrowable` would be more appropriate maybe

app/src/standard/java/com/infomaniak/mail/firebase/ProcessMessageNotificationsWorker.kt
```diff
@@ -117,11 +122,29 @@ class ProcessMessageNotificationsWorker @AssistedInject constructor(
...
             if (!hasShownNotification) displayGenericNewMailsNotification()
             mailboxContentRealm.close()
         }
+
+        // We don't want to log to Sentry if the error is a common outcome.
+        when (workerFailureReason) {
+            is CancellationException,
+            is NetworkException -> Unit
+            else -> {
+                checkIfNotificationMessageUidIsInRealm(messageUid, mailboxContentRealm)
```
LunarX: The way you wrote it, this will also be checked if the notification is successfully displayed. Is this behavior required?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -175,6 +179,29 @@ class FetchMessagesManager @Inject constructor(
...
+        }
+
+        if (hasSentryMessageBeenFetched.not()) {
+            SentryDebug.sendFailedNotification(
+                "Message uid from notification has not been fetched",
+                userId,
+                mailbox.mailboxId,
+                notificationMessageUidToLog,
+                mailbox
+            )
```
LunarX: Is it really a good idea to use `sendFailedNotification()`? I feel like this loggin method's purpose is to log other types of errors. Errors for which we could not display a notification to the user. But in this new situation, you check something entirely different. Should it not be its own log?

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -175,6 +179,28 @@ class FetchMessagesManager @Inject constructor(
...
+        threadsWithNewMessages: List<Thread>,
+        notificationMessageUidToLog: String?,
+        userId: Int,
+        mailbox: Mailbox
+    ) {
+        val hasSentryMessageBeenFetched = threadsWithNewMessages.any { thread ->
+            thread.messages.any { message ->
+                message.uid == notificationMessageUidToLog
+            }
+        }
```
LunarX: You can't search for the notificationMessageUidToLog inside of threadsWithNewMessages. You're not gonna find it on every subsequent firebase notification inside of a group of notifications that arrive in parallel.

You have not other choice but to look for them inside of realm, which is the way you did pitch this PR's idea to me in the first place

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -154,6 +154,10 @@ class FetchMessagesManager @Inject constructor(
             return true
         }
 
+        notificationMessageUidToLog?.let { notificationMessageUidToLog ->
+            logNotFetchedMessageUid(threadsWithNewMessages, notificationMessageUidToLog, userId, mailbox)
```
LunarX: Also your code doesn't trigger at all when threadsWithNewMessages is empty because you're handling a subsequent firebase notification in a parallel group

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -177,6 +181,22 @@ class FetchMessagesManager @Inject constructor(
...
     }
 
+    private fun logNotFetchedMessageUid(
+        notificationMessageUidToLog: String,
+        userId: Int,
+        mailbox: Mailbox,
+        realm: Realm,
+    ) {
+        if (MessageController.doesMessageExist(notificationMessageUidToLog, realm)) {
+            SentryDebug.sendMissingMessageUidInRealm(
```
LunarX: Why do we send a "missing message" sentry when the message exists?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -190,6 +191,10 @@ class MessageController @Inject constructor(
         private fun getMessagesByFolderIdQuery(folderId: String, realm: TypedRealm): RealmQuery<Message> {
             return realm.query<Message>("${Message::folderId.name} == '$folderId'")
         }
+
+        private fun doesMessageExistQuery(uid: String, realm: TypedRealm): RealmScalarQuery<Long> {
+            return realm.query<Message>("${Message::uid.name} == $0", uid).count()
+        }
```
LunarX: The method starts with `does` but doesn't return a boolean. Maybe we can find a better naming

TommyDL-Infomaniak: It is named this way because it is the query of the doesMessageExist method.

LunarX: We don't name methods based on what other method uses them but rather on what each method does. If this method's naming bothers you, another alternative would be to not create a new query and only call `getMessagesQuery()` and inside of `doesMessageExistQuery()` you would call `.count().find() > 0`

app/src/main/java/com/infomaniak/mail/utils/SentryDebug.kt
```diff
@@ -208,6 +208,23 @@ object SentryDebug {
         }
     }
 
+    fun sendMissingMessageUidInRealm(
+        userId: Int? = null,
+        mailboxId: Int? = null,
+        messageUid: String? = null,
+        mailbox: Mailbox? = null,
+        throwable: Throwable? = null,
```
LunarX: `throwable: Throwable? = null` is unused

LunarX (approved):

PR #2465, opened 2025-07-11
feat(Encryption-11): Various fixes

FabianDevel:
- Hide the button in the encryptionAlert as long as the back has not put in production the route update 
- Fix encyptionButton not being diplayed at init
- Disable the chip from EncryptionFragment and make them scrollable
- Update copy password button to make it share instead of copy
- Update the password description traduction
- Fix the unencryptable recipients chips being wrongly considered as encryptable after removing  it from one field when it's still in others


app/src/main/java/com/infomaniak/mail/ui/newMessage/encryption/EncryptionLockButtonView.kt
```diff
@@ -60,6 +60,10 @@ class EncryptionLockButtonView @JvmOverloads constructor(
             setToolbarButtonUi()
         }
 
+    init {
+        setToolbarButtonUi()
```
tevincent, line 61: What do you think about that ?

```kotlin
private fun setToolbarButtonUi() {
        loadingDelayTimer.cancel()

        with(encryptionStatus) {
            setIconUi(
                iconRes =  iconRes,
                iconTintRes = iconTintRes,
                shouldDisplayPastille = shouldDisplayPastille,
                isLoading = isLoading,
            )
        }
    }

enum class EncryptionStatus(
    @DrawableRes val iconRes: Int,
    @ColorRes val iconTintRes: Int,
    val shouldDisplayPastille: Boolean,
    val isLoading: Boolean = false,
) {
    Unencrypted(
        iconRes = R.drawable.ic_lock_open_filled,
        iconTintRes = R.color.iconColor,
        shouldDisplayPastille = false,
    ),
    Loading(
        iconRes = R.drawable.ic_lock_filled,
        iconTintRes = R.color.encryptionIconColor,
        shouldDisplayPastille = false,
        isLoading = true,
    ),
    PartiallyEncrypted(
        iconRes = R.drawable.ic_lock_filled,
        iconTintRes = R.color.encryptionIconColor,
        shouldDisplayPastille = true,
    ),
    Encrypted(
        iconRes = R.drawable.ic_lock_filled,
        iconTintRes = R.color.encryptionIconColor,
        shouldDisplayPastille = false,
    )
}
```

tevincent (changes requested):

tevincent (approved):

PR #2475, opened 2025-07-15
fix: Search chips go under the navbar and cutout

MorganeFar:


TommyDL-Infomaniak:
This fix also causes a spacing problem:

Screenshot of a spacing problem with the search feature

app/src/main/java/com/infomaniak/mail/ui/main/search/SearchFragment.kt
```diff
@@ -131,8 +132,12 @@ class SearchFragment : TwoPaneFragment() {
 
     private fun handleEdgeToEdge(): Unit = with(binding) {
         applyWindowInsetsListener(shouldConsume = false) { _, insets ->
-            appBar.applyStatusBarInsets(insets)
+            toolbar.applySideAndBottomSystemInsets(insets, withBottom = false)
             swipeRefreshLayout.applySideAndBottomSystemInsets(insets, withBottom = false)
+            chipsList.applySideAndBottomSystemInsets(insets, withBottom = false)
+            with(insets.safeArea()) {
+                toolbar.setMargins(top = top)
+            }
```
LunarX: No need to use this `with()` when it's only used to access a single value

LunarX: Regroup the lines of code that apply to similar items together so it's easier to read and info on a specific view is not spread through the code

app/src/main/java/com/infomaniak/mail/ui/main/search/SearchFragment.kt
```diff
@@ -131,8 +132,12 @@ class SearchFragment : TwoPaneFragment() {
 
     private fun handleEdgeToEdge(): Unit = with(binding) {
         applyWindowInsetsListener(shouldConsume = false) { _, insets ->
-            appBar.applyStatusBarInsets(insets)
+            toolbar.applySideAndBottomSystemInsets(insets, withBottom = false)
             swipeRefreshLayout.applySideAndBottomSystemInsets(insets, withBottom = false)
+            chipsList.applySideAndBottomSystemInsets(insets, withBottom = false)
+            with(insets.safeArea()) {
+                toolbar.setMargins(top = top)
```
LunarX: Add a comment explaining the implications of this setMargins and why we need it

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/search/SearchFragment.kt
```diff
@@ -131,8 +132,10 @@ class SearchFragment : TwoPaneFragment() {
 
     private fun handleEdgeToEdge(): Unit = with(binding) {
         applyWindowInsetsListener(shouldConsume = false) { _, insets ->
-            appBar.applyStatusBarInsets(insets)
+            toolbar.applySideAndBottomSystemInsets(insets, withBottom = false)
+            toolbar.setMargins(top = insets.safeArea().top) // Apply margins to keep the icon and the search bar aligned
```
LunarX: You can use a `toolbar.apply` to group these two lines even more.

Also, group `chipsList` next to `toolbar` as they are both views contained inside of the app bar layout unlike `swipeRefreshLayout`

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/search/SearchFragment.kt
LunarX: The comment needs to explain more clearly why suddenly we can't just use paddings like everywhere else and need to use margin and split it from the horizontal padding from the line above it.

We can easily see that the line that calls `setMargins()` is used to ... set margin, don't worry 😂

LunarX (changes requested):

LunarX (approved):

PR #2485, opened 2025-07-21
fix: Align the icon and title in the new message screen

MorganeFar:


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -395,7 +396,9 @@ class NewMessageFragment : Fragment() {
         addressListPopupWindow = ListPopupWindow(binding.root.context)
 
         toolbar.setNavigationOnClickListener { activity?.onBackPressedDispatcher?.onBackPressed() }
-        changeToolbarColorOnScroll(toolbar, compositionNestedScrollView)
+        changeToolbarColorOnScroll(toolbar, compositionNestedScrollView, otherUpdates = { color ->
+            appBarLayout.backgroundTintList = ColorStateList.valueOf(color)
+        })
```
LunarX: Now that it's always called with an appBarLayout, we could change the signature of the method to take in an AppBarLayout instead of a MaterialToolbar so it does things automatically for everyone. Also this will force people to correctly handled insets if they want to use `changeToolbarColorOnScroll()` by defining a parent app bar layout

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -28,19 +28,25 @@
...
+                android:id="@+id/toolbar"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:minHeight="?attr/actionBarSize"
+                android:touchscreenBlocksFocus="false"
+                app:navigationContentDescription="@string/buttonClose"
+                app:navigationIcon="@drawable/ic_close_big"
+                app:title="@string/buttonNewMessage"
+                app:titleCentered="true" />
+        </com.google.android.material.appbar.AppBarLayout>
```
LunarX: Somehow, we can't navigate to the toolbar close button anymore using the keyboard now. It must have to do with `touchscreenBlocksFocus` maybe. Can you check how to avoid this regression? Other toolbars in the app should already work correctly in this regard

PR #2497, opened 2025-07-25
feat: Update emoji reaction model

LunarX:
Update the data we store in realm to include from what author and message uid each emoji reaction comes from. This data is needed to display details of who used what emoji in a bottom sheet later on and also to be able to delete all messages that were emoji reactions alongside the target message deletion.

I also changed the data we store in realm and use in the compose code from unordered dictionaries/sets to ordered lists so the emojis are always ordered the way they were sent in the thread.

The method containsTheSameEmojiValuesAs() now checks for nullability for no reason even though we don't need it anymore because we're not using RealmDictionary anymore. The problem is fixed in the next PR because it make a lot of noise to fix it here. The tests associated with this method are also fixed there. 

Depends on #2437

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/message/EmojiReactionState.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/message/EmojiReactionState.kt
@@ -17,19 +17,29 @@
  */
 package com.infomaniak.mail.data.models.message
 
-import com.infomaniak.emojicomponents.data.ReactionState
+import com.infomaniak.emojicomponents.data.Reaction
 import com.infomaniak.mail.data.models.correspondent.Recipient
 import io.realm.kotlin.ext.realmListOf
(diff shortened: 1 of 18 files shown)


app/src/main/java/com/infomaniak/mail/data/models/message/EmojiReactionState.kt
```diff
@@ -17,28 +17,26 @@
...
+        this.emoji = emoji
     }
 
-    override fun hashCode(): Int {
-        var result = count
-        result = 31 * result + hasReacted.hashCode()
-        return result
+    fun addAuthor(newAuthor: EmojiReactionAuthor) {
+        authors.add(newAuthor)
+        hasReacted = hasReacted || newAuthor.recipient?.isMe() == true
```
sirambd: We can simplify with a `if` condition for better readability

sirambd (changes requested):

PR #2499, opened 2025-07-25
feat: Delete reactions alongside message

LunarX:
Add query param so the backend will delete emoji reactions messages alongside target messages when they are moved/deleted. This query param needs to only be used when the feature is available.

This caused emoji reaction messages from other folders to quickly pop into the thread as the folders get refreshed. To solve this issue, we also delete the emoji reaction messages of a target message locally from realm when the delete called succeeded before even refreshing folders


app/src/main/java/com/infomaniak/mail/data/api/ApiRepository.kt
```diff
@@ -297,11 +298,16 @@ object ApiRepository : ApiRepositoryCore() {
         mailboxUuid: String,
         messagesUids: List<String>,
         destinationId: String,
+        alsoMoveReactionMessages: Boolean,
         okHttpClient: OkHttpClient = HttpClient.okHttpClient,
     ): List<ApiResponse<MoveResult>> {
+        fun createApiRoute(): String {
+            return ApiRoutes.moveMessages(mailboxUuid) + if (alsoMoveReactionMessages) "?move_reactions=1" else ""
+        }
```
sirambd: A `val` is suffisant

sirambd (changes requested):

PR #2510, opened 2025-07-29
fix: Add logs when an error occured on Login and Account Creation webview

tevincent:
depends on https://github.com/Infomaniak/android-core/pull/457


app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
```diff
@@ -433,6 +434,11 @@ fun Context.getInfomaniakLogin() = InfomaniakLogin(
     appUID = BuildConfig.APPLICATION_ID,
     clientID = BuildConfig.CLIENT_ID,
     accessType = null,
+    sentryCallback = { error ->
+        SentryLog.e(tag = "WebViewLogin", "An error occurred on the login/Account creation webview", scopeCallback = {
+            it.setTag("error", error)
+        })
```
sirambd: Use directly error as message for WebViewLogin tag

sirambd (changes requested):

sirambd (approved):

PR #2535, opened 2025-08-14
chore: Bump libs

tevincent:


LouisCAD (changes requested):
@tevincent Could you make a separate PR, or at least a separate commit that is only about the Groovy DSL to Kotlin DSL migration?

tevincent:
> @tevincent Could you make a separate PR, or at least a separate commit that is only about the Groovy DSL to Kotlin DSL migration?

Here it is: https://github.com/Infomaniak/android-kMail/pull/2537

LouisCAD (approved):

PR #2538, opened 2025-08-14
feat: Use factorized logic in cross app login view model

LunarX:
I used the newly shared logic of handle cross app login to replace the duplicated one

Depends on https://github.com/Infomaniak/android-core/pull/485

PR #2541, opened 2025-08-19
refactor: Use BuildConfig URL instead of hardcoded one in Extensions.kt

KevinBoulongne:
Depends on #2538


tevincent (changes requested):
Not building

PR #2557, opened 2025-08-26
feat: GitHub workflow concurrency and actions versions bump

LunarX:
Adds missing concurrency group block to github workflows, this let us remove the github action from the tests that was serving the same purpose. Also bumps version of github actions.

Depends on https://github.com/Infomaniak/android-core/pull/500


sirambd (approved):

sirambd (changes requested):
Fix CI

sirambd (approved):

PR #2568, opened 2025-08-29
refactor: Junk multiselect refactor to fix some bugs and clean code

FabianDevel:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/UserToBlockBottomSheetDialog.kt
```diff
@@ -17,45 +17,37 @@
...
-    override fun onViewCreated(view: View, savedInstanceState: Bundle?) = with(navigationArgs) {
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
-        mainViewModel.getMessagesFromUniqueExpeditors(threadUids.asList()).observe(viewLifecycleOwner) { messages ->
-            if (messages.isNotEmpty()) {
-                binding.recipients.adapter = UserToBlockAdapter(messages) { message ->
-                    messagesOfUserToBlock.add(message)
+
+        junkMessagesViewModel.potentialBlockedUsers.value?.let { potentialBlockedUsers ->
+            if (potentialBlockedUsers.values.isNotEmpty()) {
```
tevincent, line 45: If we arrive here with no users or a null users, what to we do ? Maybe we can, at least, dismiss that BottomSheet?

tevincent (changes requested):

tevincent (approved):

PR #2540, opened 2025-08-18
perf: Avoid blocking I/O by using asFlow() Realm API

LouisCAD:
Some blocking functions were also renamed to have the "Blocking" suffix.

I did it all thoroughly, but we need to test all the potentially affected app features to ensure we have no regression.

PR #2575, opened 2025-09-03
perf: Avoid blocking I/O

LouisCAD:
This time with more thorough self-review, and fix of the regressions from #2540 that were reverted in #2560

Found one "random" issue that might or might not be a regression so far:
- Sending a draft showing the sending in progress snackbar, but never completing, staying as a draft. (Couldn't reproduce it yet)

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
@@ -41,7 +41,7 @@ import splitties.experimental.ExperimentalSplittiesApi
 class DownloadMessagesProgressDialog : DownloadProgressDialog() {
     private val downloadThreadsViewModel: DownloadMessagesViewModel by viewModels()
 
-    override val dialogTitle: String = "-" // Placeholder value to ensure the dialog's title view gets created.
+    override val dialogTitle: String = " " // Non-empty placeholder value to ensure the dialog's title view gets created.
 
     private val dialogTitleLazy = lifecycleScope.suspendLazy { downloadThreadsViewModel.getDialogName() }
 


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -289,7 +315,7 @@ class NewMessageViewModel @Inject constructor(
         return getLocalOrRemoteDraft(localUuid)?.also { draft ->
             saveNavArgsToSavedState(draft.localUuid)
             if (draft.identityId.isNullOrBlank()) {
-                draft.identityId = currentMailbox.getDefaultSignatureWithFallback().id.toString()
+                draft.identityId = currentMailbox().getDefaultSignatureWithFallback().id.toString()
```
tevincent, line 318: Do you really need to call `currentMailbox()` everytime ? You can, like you did in other files, store in a `val` `currentMailbox()` ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
```diff
@@ -15,32 +15,47 @@
...
+
+    private val dialogTitleLazy = lifecycleScope.suspendLazy { downloadThreadsViewModel.getDialogName() }
 
     override fun onCreate(savedInstanceState: Bundle?) {
         observeDownload()
         super.onCreate(savedInstanceState)
     }
 
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog = super.onCreateDialog(savedInstanceState).also { dialog ->
+        lifecycleScope.launch {
```
tevincent, line 54: Shouldn't be `viewLifecycleOwner.lifecycleScope.launch` here ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
```diff
@@ -15,32 +15,47 @@
...
 import com.infomaniak.mail.utils.SaveOnKDriveUtils.canSaveOnKDrive
+import kotlinx.coroutines.launch
+import splitties.coroutines.suspendLazy
+import splitties.experimental.ExperimentalSplittiesApi
 
 class DownloadMessagesProgressDialog : DownloadProgressDialog() {
     private val downloadThreadsViewModel: DownloadMessagesViewModel by viewModels()
 
-    override val dialogTitle: String? by lazy { downloadThreadsViewModel.getDialogName() }
+    override val dialogTitle: String = "-" // Placeholder value to ensure the dialog's title view gets created.
```
tevincent: Might be better to have an empty string here instead of `-` ? Just in case it can be visible ?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
```diff
@@ -206,10 +204,12 @@ class ThreadListMultiSelection {
             changeIcon(FAVORITE_INDEX, favoriteIcon)
 
             val isSelectionEmpty = selectedThreads.isEmpty()
-            val isFromArchive = folderRoleUtils.getActionFolderRole(selectedThreads) == FolderRole.ARCHIVE
-            for (index in 0 until getButtonCount()) {
-                val shouldDisable = isSelectionEmpty || (isFromArchive && index == ARCHIVE_INDEX)
-                if (shouldDisable) disable(index) else enable(index)
+            threadListFragment.lifecycleScope.launch {
```
tevincent: `threadListFragment.viewLifecycleOwner.lifecycleScope.launch {` in this file ? Since we're doing UI stuff.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
LouisCAD: Nope, not for `DialogFragment`s using `onCreateDialog`. Did that initially, and it was a nightmare.

From the [docs](https://developer.android.com/guide/fragments/dialogs#lifecycle):
> Note: When subscribing to lifecycle-aware components such as LiveData, never use [viewLifecycleOwner](https://developer.android.com/reference/androidx/fragment/app/Fragment#getviewlifecycleowner) as the [LifecycleOwner](https://developer.android.com/reference/androidx/lifecycle/LifecycleOwner) in a DialogFragment that uses Dialog objects. Instead, use the DialogFragment itself, or, if you're using [Jetpack Navigation](https://developer.android.com/guide/navigation), use the [NavBackStackEntry](https://developer.android.com/reference/androidx/navigation/NavBackStackEntry).

Very bad API design, yeah…

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
LouisCAD: The problem here is that empty string is equal to no string for the dialog, and the underlying `TextView` doesn't get created or added to the layout, and updating the text later on has no visible effect, so we need at least one character.

Using a space character works though. Do you think an empty line is better of a placeholder than a short dash?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
LouisCAD: Nice catch. I initially thought this was setting some state that was kept through view recreation cycles, but it's actually touching the UI directly, and is only reflecting observed state, so yes, should be using `viewLifecycleOwner`. I will fix this.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
LouisCAD: It's used only once in this function, so I don't see an alternative that wouldn't mean having blocking code.

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListMultiSelection.kt
LouisCAD: Fixed, thanks

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
tevincent: My bad, github collapsed the code so I thought it was in the same method.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
tevincent: Yeah just to be extra sure, I prefer a space instead of a -. 🙏

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadMessagesProgressDialog.kt
LouisCAD: Updated

PR #2583, opened 2025-09-15
refactor: Remove unused variables in Message class

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -148,7 +148,8 @@ class Message : RealmObject, Snoozable {
 
     // TODO: Those are unused for now, but if we ever want to use them, we need to save them in `Message.keepHeavyData()`.
     //  If we don't do it now, we'll probably forget to do it in the future.
-    //  When we use them, ce can remove all comments here and in `Message.keepHeavyData()`.
+    //  When we use them, we can remove all comments here and in `Message.keepHeavyData()`.
+    @SerialName("acknowledge")
```
LunarX: We can't use these `acknowledge` in the future anymore because they were badly deserilialized and therefore not saved to the database right?

PR #2586, opened 2025-09-16
feat: Format reaction notification

baillyjamy:
Clean emoji reaction preview and add api to reuse it for ThreadList

|  Before  |  After | 
|---|---|
|  Screenshot_20250916_093330  |  Screenshot_20250916_152208  |

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/message/FormatedPreview.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/message/FormatedPreview.kt
@@ -1,0 +1,31 @@
+/*
+ * Infomaniak Mail - Android
+ * Copyright (C) 2025 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.mail.data.models.message
+
+sealed interface FormatedPreview {
+    val content: String
+
+    @JvmInline
+    value class Encryption(override val content: String): FormatedPreview
+    @JvmInline
+    value class Reaction(override val content: String): FormatedPreview
+    @JvmInline
+    value class Empty(override val content: String): FormatedPreview
+    @JvmInline
+    value class Body(override val content: String): FormatedPreview
+}
--- a/app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
@@ -21,7 +21,6 @@ package com.infomaniak.mail.data.models.message
 
 import android.content.Context
 import com.infomaniak.core.utils.apiEnum
-import com.infomaniak.html.cleaner.HtmlSanitizer
 import com.infomaniak.lib.core.utils.Utils.enumValueOfOrNull
 import com.infomaniak.mail.R
 import com.infomaniak.mail.data.api.RealmInstantSerializer
@@ -40,9 +39,7 @@ import com.infomaniak.mail.data.models.getMessages.DefaultMessageFlags
 import com.infomaniak.mail.data.models.getMessages.SnoozeMessageFlags
 import com.infomaniak.mail.data.models.thread.Thread
 import com.infomaniak.mail.utils.AccountUtils
-import com.infomaniak.mail.utils.MessageBodyUtils
 import com.infomaniak.mail.utils.MessageBodyUtils.SplitBody
-import com.infomaniak.mail.utils.extensions.removeLineBreaksFromHtml
 import com.infomaniak.mail.utils.extensions.replaceContent
 import com.infomaniak.mail.utils.extensions.toRealmInstant
 import io.realm.kotlin.ext.backlinks
--- a/app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
@@ -417,11 +417,12 @@ class Message : RealmObject, Snoozable {
     }
 
     fun getFormattedPreview(context: Context): String {
-        return when {
-            isEncrypted -> "\n" + context.getString(R.string.encryptedMessageHeader)
-            isReaction -> "\n" + context.getString(R.string.previewReaction, from.first().name, emojiReaction)
-            else -> preview.ifBlank { null }?.let { "\n${it.trim()}" } ?: ""
+        val content = when {
+            isEncrypted -> context.getString(R.string.encryptedMessageHeader)
+            isReaction -> context.getString(R.string.previewReaction, from.first().name, emojiReaction)
+            else -> preview.ifBlank { null }?.trim()
         }
+        return content?.let { "\n$it" } ?: ""
     }
 
     fun shouldBeExpanded(index: Int, lastIndex: Int) = !isDraft && (!isSeen || index == lastIndex)
--- a/app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
@@ -416,11 +416,18 @@ class Message : RealmObject, Snoozable {
         snoozeEndDate = flags.snoozeEndDate.toRealmInstant()
     }
 
-    fun getFormattedPreview(context: Context): String = "\n" + when {
-        isEncrypted -> context.getString(R.string.encryptedMessageHeader)
-        isReaction -> context.getString(R.string.previewReaction, from.first().name, emojiReaction)
-        preview.isBlank() -> context.getString(R.string.noBodyDescription)
-        else -> preview.trim()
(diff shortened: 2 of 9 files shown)


app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -414,6 +416,14 @@ class Message : RealmObject, Snoozable {
         snoozeEndDate = flags.snoozeEndDate.toRealmInstant()
     }
 
+    fun getFormattedPreview(context: Context): String {
+        return when {
+            isEncrypted -> "\n" + context.getString(R.string.encryptedMessageHeader)
+            isReaction -> "\n" + context.getString(R.string.previewReaction, from.first().name, emojiReaction)
+            else -> preview.ifBlank { null }?.let { "\n${it.trim()}" } ?: ""
+        }
```
LunarX: Maybe it could make sense to factorize the common leading `\n`

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -211,7 +211,7 @@ class FetchMessagesManager @Inject constructor(
         // We can leave safely.
         if (message.isSeen) return true
 
-        val formattedPreview = message.preview.ifBlank { null }?.let { "\n${it.trim()}" } ?: ""
+        val formattedPreview = message.getFormattedPreview(appContext)
         val body = if (message.body?.value.isNullOrBlank()) {
             formattedPreview
         } else {
```
LunarX: We should have had a non blank body, how come the notification did not go into the `else` case?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
baillyjamy: I don't have enough history in the project, so I don't know if this is normal, but here the body is null.

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -414,6 +416,14 @@ class Message : RealmObject, Snoozable {
         snoozeEndDate = flags.snoozeEndDate.toRealmInstant()
     }
 
+    fun getFormattedPreview(context: Context): String {
+        return when {
```
FabianDevel: We like to put the when as expression body when the only thing the function does is returning a when. 
It saves a level of indentation and stays easily readable.

```suggestion
    fun getFormattedPreview(context: Context) = when {
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
KevinBoulongne: It looks like there is no leading `\n` in the `else` case if the preview is blank ?

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -414,6 +416,15 @@ class Message : RealmObject, Snoozable {
...
 
+    fun getFormattedPreview(context: Context): String {
+        val content = when {
+            isEncrypted -> context.getString(R.string.encryptedMessageHeader)
+            isReaction -> context.getString(R.string.previewReaction, from.first().name, emojiReaction)
+            else -> preview.ifBlank { null }?.trim()
+        }
+        return content?.let { "\n$it" } ?: ""
+    }
+
```
baillyjamy: For readability, which do you prefer? @FabianDevel 


```suggestion
    fun getFormattedPreview(context: Context) = when {
            isEncrypted -> context.getString(R.string.encryptedMessageHeader)
            isReaction -> context.getString(R.string.previewReaction, from.first().name, emojiReaction)
            else -> preview.ifBlank { null }?.trim()
    }?.let { "\n$it" } ?: ""

```

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
baillyjamy: That's right, but I proposed another refactored version that doesn't add \n if the preview is blank.

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
FabianDevel: The second is good. 

But it just came to my mind that when the preview is blank, we normally put the R.string.noBodyTitle string

baillyjamy: But it's not for a title. Should I still use the resource `noBodyTitle` ?

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
LunarX: Do we even make api calls to download it during this step or not at all?

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
FabianDevel: Yes it's true that the resource is not named the best way, as it's technically never used for a title. 
But yes use it, and you could maybe rename it in loco with "noBodyDescription"

baillyjamy: Change done

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
KevinBoulongne: There is now a hardcoded `\n` in every returns of `getFormattedPreview()`, this is contradictory with what you just said.

baillyjamy: Because in the meantime, I was told that there should be no empty previews, but if that happens, then display the message from the resource `R.string.noBodyDescription`, so ultimately the `\n` can apply to all of them.

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
baillyjamy: After investigation with @sirambd. The body is not loaded in the case of a reaction. So yes, it is normal to use the fallback to handle the case of the reaction or the encrypted message.

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -414,6 +416,20 @@ class Message : RealmObject, Snoozable {
...
+        PreviewType.Body -> preview.trim()
+    }
+
+    fun getPreviewType(): PreviewType = when {
+        isEncrypted -> PreviewType.Encryption
+        isReaction -> PreviewType.Reaction
+        preview.isBlank() -> PreviewType.Empty
+        else -> PreviewType.Body
+    }
+
```
sirambd: Use `value class` to simplify it

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -433,6 +449,13 @@ class Message : RealmObject, Snoozable {
 
     override fun hashCode(): Int = uid.hashCode()
 
+    enum class PreviewType {
+        Encryption,
+        Reaction,
+        Empty,
+        Body
```
sirambd: Extract to a new file

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
baillyjamy: Done

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
baillyjamy: Done

app/src/main/java/com/infomaniak/mail/data/models/message/FormatedPreview.kt
```diff
@@ -0,0 +1,31 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.infomaniak.mail.data.models.message
+
+sealed interface FormatedPreview {
+    val content: String
```
sirambd, line 21: you can use directly a `sealed class` instead

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
```diff
@@ -213,21 +214,21 @@ class FetchMessagesManager @Inject constructor(
...
-            formattedPreview
-        } else {
+        val notificationBody = "\n" + if (formattedPreview is FormatedPreview.Body) {
             message.body
                 ?.let {
                     val content = MessageBodyUtils.splitContentAndQuote(it).content
                     val dirtyDocument = content.removeLineBreaksFromHtml()
                     val cleanedDocument = HtmlSanitizer.getInstance().sanitize(dirtyDocument)
-                    return@let "\n${cleanedDocument.wholeText().trim()}"
+                    return@let cleanedDocument.wholeText().trim()
```
sirambd, line 223: Why do you remove the `\n`

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/utils/FetchMessagesManager.kt
baillyjamy: Because I move it here. We add `\n` in every case
```kotlin
val notificationBody = "\n" + if (formattedPreview is FormatedPreview.Body) {
```

sirambd (approved):

PR #2587, opened 2025-09-17
feat: Stop reduce message if alone

baillyjamy:
If the thread contains only one message, disable “handleHeaderClick()” to prevent the message from being reduced.

Example of what we don't want in video

https://github.com/user-attachments/assets/ab0c8025-dc60-43c2-b5b8-c94a588e5127

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
@@ -916,7 +916,8 @@ class ThreadAdapter(
         data object FailedMessage : NotifyType
         data object OnlyRebindCalendarAttendance : NotifyType
         data object OnlyRebindEmojiReactions : NotifyType
-        data class MessagesCollapseStateChanged(val isCollapsable: Boolean) : NotifyType
+        @JvmInline
+        value class MessagesCollapseStateChanged(val isCollapsable: Boolean) : NotifyType
     }
 
     enum class ContextMenuType {
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
@@ -501,7 +501,7 @@ class ThreadFragment : Fragment() {
             threadViewModel.updateCurrentThreadUid(threadViewModel.AllMessages(threadUid))
         }
 
-        lifecycleScope.launch {
+        viewLifecycleOwner.lifecycleScope.launch {
             threadViewModel.threadFlow.collect { thread ->
                 if (thread == null) {
                     twoPaneViewModel.closeThread()
@@ -580,7 +580,7 @@ class ThreadFragment : Fragment() {
     }
 
     private fun observeMessagesIsCollapsable() {
-        lifecycleScope.launch {
+        viewLifecycleOwner.lifecycleScope.launch {
             threadViewModel.messagesIsCollapsableFlow.collect(threadAdapter::messagesCollapseStateChange)
         }
     }


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -423,7 +423,9 @@ class ThreadAdapter(
 
         setDetailedFieldsVisibility(message)
 
-        handleHeaderClick(message)
+        if (isReducible) {
+            handleHeaderClick(message)
+        }
```
FabianDevel: We oneline the expression when they're easy to read

baillyjamy: Change done

FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -235,11 +235,11 @@ class ThreadAdapter(
         }
     }
 
-    private fun MessageViewHolder.bindMail(messageUi: MessageUi, position: Int) {
+    private fun MessageViewHolder.bindMail(messageUi: MessageUi, position: Int, isReducible: Boolean = false) {
```
KevinBoulongne: There's no point in adding a default value here.
The default value implies we could forget to set the value, and that could introduce bugs.
```suggestion
    private fun MessageViewHolder.bindMail(messageUi: MessageUi, position: Int, isReducible: Boolean) {
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -219,7 +219,7 @@ class ThreadAdapter(
         }
 
         if (item is MessageUi) {
-            (holder as MessageViewHolder).bindMail(item, position)
+            (holder as MessageViewHolder).bindMail(item, position, isReducible = items.size > 1)
```
KevinBoulongne: I know it's technically exactly the same, but we tend to prefer the Kotlin Extension here.
```suggestion
            (holder as MessageViewHolder).bindMail(item, position, isReducible = items.count() > 1)
```

KevinBoulongne (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -235,11 +235,11 @@ class ThreadAdapter(
         }
     }
 
-    private fun MessageViewHolder.bindMail(messageUi: MessageUi, position: Int) {
+    private fun MessageViewHolder.bindMail(messageUi: MessageUi, position: Int, isReducible: Boolean) {
```
LunarX: Maybe `areHeadersCollapsable` or `isCollapsable` or similar would better match the existing terms as we often use the collapse/expand terms for this behavior through the code

KevinBoulongne (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
baillyjamy: I choose `isCollapsable`

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -423,7 +423,7 @@ class ThreadAdapter(
 
         setDetailedFieldsVisibility(message)
 
-        handleHeaderClick(message)
+        if (isCollapsable) handleHeaderClick(message)
```
LunarX: Because we're still in an adapter nonetheless, I'd make sure the code is clean by setting the click listener to null when we want the mail to not be collapsable

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -219,7 +219,7 @@ class ThreadAdapter(
         }
 
         if (item is MessageUi) {
-            (holder as MessageViewHolder).bindMail(item, position)
+            (holder as MessageViewHolder).bindMail(item, position, isCollapsable = items.count() > 1)
```
LunarX: I don't know which one is best, but I see we do strange things with itemCount, maybe we need to understand why it's computed differently to know which one to use?

```suggestion
            (holder as MessageViewHolder).bindMail(item, position, isCollapsable = itemCount > 1)
```

LunarX: And actually, on top of that, when computing the count, the logic right there takes into account all items present in the adapter which might not be the amount of messages present in the thread if we have other types of item present in the adapter's list. 

Right now, by chance, this would work, but when we will add a new type in the list, we will never think about fixing the code here and the code will definitely break. I'd rather check for the exact amount of  actual mail items instead

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
LunarX: If we have a thread with a single message and receive a 2nd message, the 2nd message will compute isCollapsable to `true` while the 1st item will remain with isCollapsable to `false` and won't be bound again even though we now have enough messages in the list to be able to collapse them messages

baillyjamy: I added a commit with a new NotifyType that triggers when messageCount change. If this notification is throw, all header recompute headerClick binding

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
baillyjamy: Now I use messageCount that filter itemsCount by `MessageUi`

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -199,6 +201,14 @@ class ThreadAdapter(
         emojiReactions.bindEmojiReactions(message)
     }
 
+    private fun MessageViewHolder.handleItemsCountChangedPayload(message: Message) {
+        handleHeaderClick(message, messageCount > 1)
```
LunarX: Same reasoning as before, I'd name the boolean parameter here

```suggestion
        handleHeaderClick(message, isCollapsable = messageCount > 1)
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -435,19 +445,24 @@ class ThreadAdapter(
...
-            if (isExpandedMap[message.uid] == true) {
-                isExpandedMap[message.uid] = false
-                onExpandOrCollapseMessage(message)
-            } else {
-                if (message.isDraft) {
-                    threadAdapterCallbacks?.onDraftClicked?.invoke(message)
-                } else {
-                    isExpandedMap[message.uid] = true
+    private fun MessageViewHolder.handleHeaderClick(message: Message, isCollapsable: Boolean) = with(threadAdapterState) {
+        binding.messageHeader.isEnabled = isCollapsable
```
LunarX, line 452: What's the goal of adding this `isEnabled` here?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -570,7 +569,12 @@ class ThreadFragment : Fragment() {
                 return@observe
             }
 
-            threadAdapter.submitList(items)
+            val messageCountBeforeUpdate = threadAdapter.messageCount
+            threadAdapter.submitList(items) {
+                if (messageCountBeforeUpdate > 0 && messageCountBeforeUpdate != threadAdapter.messageCount) {
+                    threadAdapter.messageCountChange()
+                }
```
LunarX: You notify the items even when going from 3 messages to 4 messages for exemple. But in this case there's no need to notify them as the `isCollapsed` is still the same

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -570,7 +569,12 @@ class ThreadFragment : Fragment() {
...
+            }
```
LunarX: I had in mind to compute a flow of the `isCollapsed` boolean at the view model level and listen to its changes. This could also help cache the last computed value and avoid having to compute `items.count { ... }` everytime especially during the classic `onBind` of the adapter. What do you think?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
baillyjamy: The goal is to disable ripple animation of `messageHeader` if `isCollapsable` is false

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
baillyjamy: Good idea. Since I use the `messagesLive` livedata in the `threadViewModel`, I exposed another livedata rather than a flow to avoid transformation.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -170,11 +170,14 @@ class ThreadAdapter(
...
-                NotifyType.FAILED_MESSAGE -> handleFailedMessagePayload(item.message.uid)
-                NotifyType.ONLY_REBIND_CALENDAR_ATTENDANCE -> handleCalendarAttendancePayload(item.message)
-                NotifyType.ONLY_REBIND_EMOJI_REACTIONS -> handleEmojiReactionPayload(item)
+                NotifyType.ToggleLightMode -> holder.handleToggleLightModePayload(item.message.uid)
+                NotifyType.ReRender -> reloadVisibleWebView()
+                NotifyType.FailedMessage -> handleFailedMessagePayload(item.message.uid)
+                NotifyType.OnlyRebindCalendarAttendance -> handleCalendarAttendancePayload(item.message)
+                NotifyType.OnlyRebindEmojiReactions -> handleEmojiReactionPayload(item)
+                is NotifyType.MessagesCollapseStateChanged -> {
+                    holder.handleItemsCountChangedPayload(item.message, isCollapsable = payload.isCollapsable)
```
LunarX: I'd rename the `handleItemsCountChangedPayload` method name for something more specific

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -219,7 +230,8 @@ class ThreadAdapter(
         }
 
         if (item is MessageUi) {
-            (holder as MessageViewHolder).bindMail(item, position)
+            val messageCount = runCatchingRealm { items.count { it is MessageUi } }.getOrDefault(0)
```
LunarX: Here we can avoid looping over the list to count the items if we access the cached messageCount in the view model

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -884,16 +902,21 @@ class ThreadAdapter(
...
     }
 
     // Only public because it's accessed inside of a test file
-    enum class NotifyType {
-        TOGGLE_LIGHT_MODE,
-        RE_RENDER,
-        FAILED_MESSAGE,
-        ONLY_REBIND_CALENDAR_ATTENDANCE,
-        ONLY_REBIND_EMOJI_REACTIONS,
+    sealed class NotifyType {
```
LunarX: we could turn NotifyType into a sealed interface

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -884,16 +902,21 @@ class ThreadAdapter(
...
+        object ToggleLightMode : NotifyType()
+        object ReRender : NotifyType()
+        object FailedMessage : NotifyType()
+        object OnlyRebindCalendarAttendance : NotifyType()
+        object OnlyRebindEmojiReactions : NotifyType()
+        class MessagesCollapseStateChanged(val isCollapsable: Boolean) : NotifyType()
```
LunarX: Do you have a preference between objects vs data objects and class vs data class here?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -578,6 +578,10 @@ class ThreadFragment : Fragment() {
         }
     }
 
+    private fun observeMessagesIsCollapsable() = with(threadViewModel) {
+        isCollapsable.observe(viewLifecycleOwner, threadAdapter::messagesCollapseStateChange)
+    }
```
LunarX: We usually bother with `with(threadViewModel)` when it's accessed more than once in the method but not otherwise

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -206,6 +207,15 @@ class ThreadViewModel @Inject constructor(
...
         }
 
+    val isCollapsable: LiveData<Boolean> = MediatorLiveData<Boolean>()
+        .apply {
+            addSource(messagesLive) { (items, _) ->
+                val messageCount = runCatchingRealm { items.count { it is MessageUi } }.getOrDefault(0)
+                value = messageCount > 1
+            }
+        }
+        .distinctUntilChanged()
```
LunarX: The idea of keeping a LiveData instead of turning everything into a Flow is appreciable but here it's trivial to turn the messages into a flow and keep exposing a LiveData of messages nonetheless. This will help us get rid of live datas little by little

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
baillyjamy: Currently, we don't use `toString(), equals(), hashCode()`, so I thought we wouldn't need them. But I don't see any problem with using the keyword `data` for future use.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -102,6 +102,7 @@ class ThreadAdapter(
     private val shouldLoadDistantResources: Boolean,
     private val isForPrinting: Boolean = false,
     private val isSpamFilterActivated: () -> Boolean = { false },
+    private val messagesIsCollapsable: () -> Boolean,
```
LunarX: To be correct, it should be written with "are" instead of "is" because there are multiple messages. Also we usually name booleans that have the verb "to be" by starting the name with the verb like "isEnabled" or "areContentsTheSame"

```suggestion
    private val areMessagesCollapsable: () -> Boolean,
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -219,7 +231,7 @@ class ThreadAdapter(
         }
 
         if (item is MessageUi) {
-            (holder as MessageViewHolder).bindMail(item, position)
+            (holder as MessageViewHolder).bindMail(item, position, isCollapsable = messagesIsCollapsable())
```
LunarX: Because the callback is accessible anywhere in the class we don't really need to evaluate it here and pass it down through like 3 methods. We can directly access it at the lowest possible method I think

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -199,6 +203,14 @@ class ThreadAdapter(
         emojiReactions.bindEmojiReactions(message)
     }
 
+    private fun MessageViewHolder.handleMessagesCollapseStatePayload(message: Message, isCollapsable: Boolean) {
+        handleHeaderClick(message, isCollapsable)
+        if (!isCollapsable) {
+            threadAdapterState.isExpandedMap[message.uid] = true
+            onExpandOrCollapseMessage(message)
```
LunarX: We should not track to matomo automatic expand or collapse of threads like this

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -184,7 +187,17 @@ class ThreadViewModel @Inject constructor(
...
+        }
+
+    val messagesLive: LiveData<Pair<ThreadAdapterItems, MessagesWithoutHeavyData>> = messagesFlow.asLiveData(ioCoroutineContext)
+
+    val messagesIsCollapsableFlow: StateFlow<Boolean> = messagesFlow
+        .map { (items, _) ->
+            val messageCount = runCatchingRealm { items.count { it is MessageUi } }.getOrDefault(0)
+            return@map messageCount > 1
+        }
+        .distinctUntilChanged()
```
LunarX: `StateFlow`s naturally won't emit repetitions of the same value. `distinctUntilChanged()` is therefore not needed

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
baillyjamy: I remove the default parameter to avoid forgetting to define if we want to track this

LunarX (changes requested):
Sometimes when the capacity to collapse messages of the thread changes the header doesn't behave as intended

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -184,7 +186,16 @@ class ThreadViewModel @Inject constructor(
...
                 items.toUiMessages(fakeReactions, isReactionsAvailable) to messagesToFetch
             }
-        }.asLiveData(ioCoroutineContext)
+        }
+
+    val messagesLive: LiveData<Pair<ThreadAdapterItems, MessagesWithoutHeavyData>> = messagesFlow.asLiveData(ioCoroutineContext)
+
+    val messagesIsCollapsableFlow: StateFlow<Boolean> = messagesFlow
+        .map { (items, _) ->
+            val messageCount = runCatchingRealm { items.count { it is MessageUi } }.getOrDefault(0)
```
sirambd: - Add a comment from the list definition of Any, and simplify this condition.
- Remove the runCatchingRealm because it's useless

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -578,6 +579,12 @@ class ThreadFragment : Fragment() {
         }
     }
 
+    private fun observeMessagesIsCollapsable() {
+        lifecycleScope.launch {
```
sirambd: Prefer use `viewLifecyleOwner.lifecycleScope` instead

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -884,16 +902,21 @@ class ThreadAdapter(
...
-        FAILED_MESSAGE,
-        ONLY_REBIND_CALENDAR_ATTENDANCE,
-        ONLY_REBIND_EMOJI_REACTIONS,
+    sealed interface NotifyType {
+        data object ToggleLightMode : NotifyType
+        data object ReRender : NotifyType
+        data object FailedMessage : NotifyType
+        data object OnlyRebindCalendarAttendance : NotifyType
+        data object OnlyRebindEmojiReactions : NotifyType
+        data class MessagesCollapseStateChanged(val isCollapsable: Boolean) : NotifyType
```
sirambd: Use a value class here instead

sirambd (approved):

PR #2602, opened 2025-09-22
feat: Add reaction action bottom sheet

baillyjamy:
Add bottom sheet button to add reaction.

This pull request content a refactor of `getLastMessageToExecuteAction` to add an optional filter. This optional filter is use to find the last message can be react. 

The shared flow that exposes the last actionable message has been combined with another flow that checks whether the last message can be reacted to. An actionable message is not necessarily a message that can be reacted to.

This button is added to bottom sheets linked to messages or threads.

Screenshot_20250919_105910


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,11 +99,23 @@ class MessageController @Inject constructor(
...
+        val messages = thread.getDisplayedMessages(featureFlags, this@MessageController.localSettings)
+        return getLastMessageToExecuteActionWithExtraQuery(messages = messages) ?: messages.last()
+    }
+
+    suspend fun getLastMessageCanBeReact(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message? {
+        val isValidReactionTarget = "${Message::_emojiReactionNotAllowedReason.name} == null"
+
+        val messages = thread.getDisplayedMessages(featureFlags, this@MessageController.localSettings)
+        return getLastMessageToExecuteActionWithExtraQuery(messages, extraQuery = isValidReactionTarget)
     }
```
LunarX, line 91: What do you think about trying out this kind of naming approach?

```
messages.lastToExecuteActionOn()
messages.lastToExecuteReactionOn()
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,11 +99,23 @@ class MessageController @Inject constructor(
...
+            ?: extraQuery?.let { messages.query(extraQuery).last() }
+    }
+
+    suspend fun getLastMessageToExecuteAction(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message {
+        val messages = thread.getDisplayedMessages(featureFlags, this@MessageController.localSettings)
+        return getLastMessageToExecuteActionWithExtraQuery(messages = messages) ?: messages.last()
+    }
+
+    suspend fun getLastMessageCanBeReact(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message? {
+        val isValidReactionTarget = "${Message::_emojiReactionNotAllowedReason.name} == null"
```
LunarX: I'd probably have explained what makes a reaction target valid by specifying it in the name of the variable like this directly in the first place:

```suggestion
        val canBeReactedTo = "${Message::_emojiReactionNotAllowedReason.name} == null"
```

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -695,6 +700,17 @@ class ThreadListFragment : TwoPaneFragment() {
         }
     }
 
+    private fun observePickedEmoji() {
+        getBackNavigationResult<PickedEmojiPayload>(ThreadList.name) { (emoji, messageUid) ->
+            trackEmojiReactionsEvent(MatomoName.AddReactionFromEmojiPicker)
```
LunarX, line 697: You could not have known as it didn't even exist, I just fixed it in our shared spec, but regarding the matomo values in the PR, we will want to change the category of the event depending on where it has been executed from. We can talk about it when you see this comment, I'll show you where you can find the info to do it.

Here we would like to adapt the category of OpenEmojiPicker and AddReactionFromEmojiPicker as specified in the spec

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MailActionsBottomSheetDialog.kt
```diff
@@ -100,6 +101,7 @@ abstract class MailActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         }
         saveKDrive.setClosingOnClickListener(shouldCloseMultiSelection) { onClickListener.onSaveToKDrive() }
         reportDisplayProblem.setClosingOnClickListener(shouldCloseMultiSelection) { onClickListener.onReportDisplayProblem() }
+        addReaction.setOnClickListener { onClickListener.onAddReaction() }
```
LunarX: Keep this line ordered in the code the same was it's ordered in the UI

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MailActionsBottomSheetDialog.kt
```diff
@@ -157,6 +159,10 @@ abstract class MailActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         }
     }
 
+    fun setReactionUi(canBeReact: Boolean) = with(binding.addReaction) {
```
LunarX: I'd have been probably more verbose on the naming and call the boolean `canBeReactedTo` or similar, but I'd be curious to know your train of thought which lead to this name

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MessageActionsBottomSheetDialog.kt
```diff
@@ -195,6 +199,11 @@ class MessageActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
                     onPositiveButtonClicked = { mainViewModel.reportDisplayProblem(message.uid) },
                 )
             }
+
+            override fun onAddReaction() {
+                trackEmojiReactionsEvent(MatomoName.OpenEmojiPicker)
+                setBackNavigationResult(OPEN_REACTION_BOTTOM_SHEET, messageUid)
+            }
```
LunarX: Same for other places in the code if there are other ones but keep this part of the code ordered the same way it's ordered in the UI and everywhere else in the code relatively to other actions

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -80,19 +82,21 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        threadActionsViewModel.threadMessageToExecuteAction.observe(viewLifecycleOwner) { (thread, messageUidToExecuteAction) ->
-            folderRole = folderRoleUtils.getActionFolderRole(thread)
-            isFromArchive = folderRole == FolderRole.ARCHIVE
-            isFromSpam = folderRole == FolderRole.SPAM
+        threadActionsViewModel.threadMessageWithActionAndReact
+            .observe(viewLifecycleOwner) { (thread, messageUidToExecuteAction, messageCanBeReactUid) ->
```
LunarX: I'd probably follow the existing naming pattern by naming the new value something in this idea
```suggestion
            .observe(viewLifecycleOwner) { (thread, messageUidToExecuteAction, messageUidToReactTo) ->
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -80,19 +82,21 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        threadActionsViewModel.threadMessageToExecuteAction.observe(viewLifecycleOwner) { (thread, messageUidToExecuteAction) ->
-            folderRole = folderRoleUtils.getActionFolderRole(thread)
-            isFromArchive = folderRole == FolderRole.ARCHIVE
-            isFromSpam = folderRole == FolderRole.SPAM
+        threadActionsViewModel.threadMessageWithActionAndReact
```
LunarX: Maybe it'd make sense to name it in plural now as it holds multiple messages
```suggestion
        threadActionsViewModel.threadMessagesWithActionAndReact
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -242,6 +246,11 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
                 onPositiveButtonClicked = { mainViewModel.reportDisplayProblem(messageUidToExecuteAction) },
             )
         }
+
+        override fun onAddReaction() {
+            trackEmojiReactionsEvent(MatomoName.OpenEmojiPicker)
+            setBackNavigationResult(OPEN_REACTION_BOTTOM_SHEET, messageUidToExecuteAction)
```
LunarX: Is this truly the correct message uid to use in this situation?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -80,19 +82,21 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
-            setArchiveUi(isFromArchive)
-            setFavoriteUi(thread.isFavorite)
-            setJunkUi()
-            setSnoozeUi(thread.isSnoozed())
+                setMarkAsReadUi(thread.isSeen)
+                setArchiveUi(isFromArchive)
+                setFavoriteUi(thread.isFavorite)
+                setJunkUi()
+                setSnoozeUi(thread.isSnoozed())
+                setReactionUi(messageCanBeReactUid != null)
```
LunarX: I'd probably have named this parameter in this situation but this is a detail
```suggestion
                setReactionUi(canBeReact = messageCanBeReactUid != null)
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -53,8 +55,20 @@ class ThreadActionsViewModel @Inject constructor(
...
+    private val threadMessageCanBeReact: Flow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
+        .mapNotNull { it.obj?.let { thread -> getThreadAndMessageUidCanBeReact(thread) } }
+
+    val threadMessageWithActionAndReact: SharedFlow<ThreadMessageToExecuteAction> =
+        combine(threadMessageToExecuteAction, threadMessageCanBeReact) { messageToExecuteActions, messageCanBeReact ->
+            ThreadMessageToExecuteAction(
+                messageToExecuteActions.thread,
+                messageToExecuteActions.messageUid,
+                messageCanBeReact.messageUid
+            )
```
LunarX: The typing here in this data class is strange. It doesn't look like `messageCanBeReactUid` can ever be null but you still defined it as a nullable String

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -53,8 +55,20 @@ class ThreadActionsViewModel @Inject constructor(
 
     private val threadUid inline get() = savedStateHandle.get<String>(ThreadActionsBottomSheetDialogArgs::threadUid.name)!!
 
-    val threadMessageToExecuteAction: SharedFlow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
+    private val threadMessageToExecuteAction: Flow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
         .mapNotNull { it.obj?.let { thread -> getThreadAndMessageUidToExecuteAction(thread) } }
+
+    private val threadMessageCanBeReact: Flow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
+        .mapNotNull { it.obj?.let { thread -> getThreadAndMessageUidCanBeReact(thread) } }
```
LunarX: Here you have a mapNotNull even though the code looks like it could find no messages to react to in the thread. If that's the case, the combine later on will never trigger because `messageCanBeReactUid` will never emit any value which should in turn never notify any observer of `threadMessageWithActionAndReact`

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -66,6 +80,11 @@ class ThreadActionsViewModel @Inject constructor(
 
     private suspend fun getThreadAndMessageUidToExecuteAction(thread: Thread): ThreadMessageToExecuteAction {
         val messageUid = Dispatchers.IO { messageController.getLastMessageToExecuteAction(thread, featureFlagsLive.value).uid }
-        return ThreadMessageToExecuteAction(thread, messageUid)
+        return ThreadMessageToExecuteAction(thread, messageUid, null)
+    }
+
+    private suspend fun getThreadAndMessageUidCanBeReact(thread: Thread): ThreadMessageToExecuteAction? {
+        val messageUid = Dispatchers.IO { messageController.getLastMessageCanBeReact(thread, featureFlagsLive.value)?.uid }
+        return messageUid?.let { ThreadMessageToExecuteAction(thread, it, it) }
```
LunarX: This intermediate way of passing values through ThreadMessageToExecuteAction where some fields are not relevant feels odd. Maybe some sealed class or whatever that could only hold the relevant data in each situation would be more appropriate

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -66,6 +80,11 @@ class ThreadActionsViewModel @Inject constructor(
 
     private suspend fun getThreadAndMessageUidToExecuteAction(thread: Thread): ThreadMessageToExecuteAction {
         val messageUid = Dispatchers.IO { messageController.getLastMessageToExecuteAction(thread, featureFlagsLive.value).uid }
-        return ThreadMessageToExecuteAction(thread, messageUid)
+        return ThreadMessageToExecuteAction(thread, messageUid, null)
+    }
+
+    private suspend fun getThreadAndMessageUidCanBeReact(thread: Thread): ThreadMessageToExecuteAction? {
+        val messageUid = Dispatchers.IO { messageController.getLastMessageCanBeReact(thread, featureFlagsLive.value)?.uid }
```
LunarX: Somehow the code above was doing it wrong as well, but you can use the injected dispatcher IO instead
```suggestion
        val messageUid = ioDispatcher { messageController.getLastMessageCanBeReact(thread, featureFlagsLive.value)?.uid }
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -984,6 +989,10 @@ class ThreadFragment : Fragment() {
         twoPaneFragment.navigateToSnoozeBottomSheet(snoozeScheduleType, threadViewModel.threadLive.value?.snoozeEndDate)
     }
 
+    private fun navigateToReactionBottomSheet(messageUid: String, emojiPickerSourceKey: EmojiPickerSourceKey) {
+        twoPaneFragment.navigateToEmojiPicker(messageUid, emojiPickerSourceKey)
+    }
```
LunarX: The `twoPaneFragment` receiver is not really needed as this class (ThreadFragment) is already a Fragment and can therefore call navigateToEmojiPicker() without specifying `twoPaneFragment`.

Also, the Fragment receiver as a whole is not needed and could be removed.

On top of all that, if you removed all of this, I'm not sure wrapping navigateToEmojiPicker() inside of navigateToReactionBottomSheet() is even needed at this point

app/src/main/res/drawable/ic_face_smile.xml
LunarX, line 1: You can remove the `android:fillColor="#00000000"` as it has no effect, but more importantly you can reformat the file

app/src/main/res/layout/bottom_sheet_actions_menu.xml
```diff
@@ -47,6 +47,16 @@
             app:visibleDivider="false"
             tools:visibility="visible" />
 
+        <com.infomaniak.mail.ui.main.thread.actions.ActionItemView
+            android:id="@+id/addReaction"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content"
+            android:visibility="gone"
+            app:icon="@drawable/ic_face_smile"
+            app:title="@string/buttonAddReaction"
```
LunarX: All other strings in this screen had another naming pattern. I don't think it's mendatory to stick to it, but it could've been a better one when you were searching how to name the key

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
baillyjamy: We decided together a better name :)

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
baillyjamy: No, I fixed it

app/src/main/res/layout/bottom_sheet_actions_menu.xml
baillyjamy: key renamed to actionAddReaction

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -53,8 +56,20 @@ class ThreadActionsViewModel @Inject constructor(
 
     private val threadUid inline get() = savedStateHandle.get<String>(ThreadActionsBottomSheetDialogArgs::threadUid.name)!!
 
-    val threadMessageToExecuteAction: SharedFlow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
+    private val threadMessageToExecuteAction: Flow<ThreadMessageInteraction> = threadController.getThreadAsync(threadUid)
         .mapNotNull { it.obj?.let { thread -> getThreadAndMessageUidToExecuteAction(thread) } }
+
+    private val threadMessageToExecuteReaction: Flow<ThreadMessageInteraction?> = threadController.getThreadAsync(threadUid)
+        .map { it.obj?.let { thread -> getThreadAndMessageToExecuteReaction(thread) } }
```
LunarX: When threadController.getThreadAsync() returns null, you will emit from the flow if you remove the `mapNotNull`

app/src/main/java/com/infomaniak/mail/utils/MessageUtils.kt
```diff
@@ -50,5 +50,13 @@ object MessageUtils {
...
-data class ThreadMessageToExecuteAction(val thread: Thread, val messageUid: String)
+sealed class ThreadMessageInteraction(open val thread: Thread, open val messageUid: String) {
+    data class Action(override val thread: Thread, override val messageUid: String) : ThreadMessageInteraction(thread, messageUid)
+    data class Reaction(override val thread: Thread, override val messageUid: String): ThreadMessageInteraction(thread, messageUid)
+}
+data class ThreadMessageToExecuteInteraction(
+    val thread: Thread,
+    val messageUidToExecuteAction: String,
+    val messageUidToExecuteReaction: String?
+)
```
LunarX: We're used to leaving a new line between classes definitions
```suggestion
sealed class ThreadMessageInteraction(open val thread: Thread, open val messageUid: String) {
    data class Action(override val thread: Thread, override val messageUid: String) : ThreadMessageInteraction(thread, messageUid)
    data class Reaction(override val thread: Thread, override val messageUid: String): ThreadMessageInteraction(thread, messageUid)
}

data class ThreadMessageToExecuteInteraction(
    val thread: Thread,
    val messageUidToExecuteAction: String,
    val messageUidToExecuteReaction: String?
)

```

app/src/main/java/com/infomaniak/mail/utils/MessageUtils.kt
```diff
@@ -50,5 +50,13 @@ object MessageUtils {
     }
 }
 
-data class ThreadMessageToExecuteAction(val thread: Thread, val messageUid: String)
+sealed class ThreadMessageInteraction(open val thread: Thread, open val messageUid: String) {
+    data class Action(override val thread: Thread, override val messageUid: String) : ThreadMessageInteraction(thread, messageUid)
+    data class Reaction(override val thread: Thread, override val messageUid: String): ThreadMessageInteraction(thread, messageUid)
+}
```
LunarX: Is it really useful to differentiate between Action and Reaction?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -182,6 +189,11 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
             }
         }
 
+        override fun onAddReaction() {
+            trackBottomSheetThreadActionsEvent(MatomoName.OpenEmojiPicker)
+            setBackNavigationResult(OPEN_REACTION_BOTTOM_SHEET, messageUidToReactTo)
```
LunarX, line 193: What happens when we call getBackNavigationResult() if we passed `null` from here? getBackNavigationResult() is defined as taking non-nullable String

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -53,8 +56,20 @@ class ThreadActionsViewModel @Inject constructor(
...
+
+    val threadMessagesWithActionAndReact: SharedFlow<ThreadMessageToExecuteInteraction> =
```
LunarX:
```suggestion
    val threadMessagesWithActionAndReaction: SharedFlow<ThreadMessageToExecuteInteraction> =
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
baillyjamy: The dialog is closed and getBackNavigationResult will never trigger the callback

app/src/main/java/com/infomaniak/mail/utils/MessageUtils.kt
```diff
@@ -50,5 +50,14 @@ object MessageUtils {
...
+sealed class ThreadMessageInteraction(open val messageUid: String) {
+    data class Action(val thread: Thread, override val messageUid: String) : ThreadMessageInteraction(messageUid)
+    data class Reaction(override val messageUid: String): ThreadMessageInteraction(messageUid)
+}
+
+data class ThreadMessageToExecuteInteraction(
+    val thread: Thread,
+    val messageUidToExecuteAction: String,
+    val messageUidToExecuteReaction: String?
+)
```
LunarX: The [comment](https://github.com/Infomaniak/android-kMail/pull/2602#discussion_r2381925053) about new lines between definitions of classes is still applicable here

app/src/main/java/com/infomaniak/mail/utils/MessageUtils.kt
```diff
@@ -50,5 +50,14 @@ object MessageUtils {
     }
 }
 
-data class ThreadMessageToExecuteAction(val thread: Thread, val messageUid: String)
+sealed class ThreadMessageInteraction(open val messageUid: String) {
+    data class Action(val thread: Thread, override val messageUid: String) : ThreadMessageInteraction(messageUid)
+    data class Reaction(override val messageUid: String): ThreadMessageInteraction(messageUid)
+}
```
LunarX: If your class is only used inside of ThreadActionsViewModel, you can define it there instead

LunarX: Same comment for ThreadMessageToExecuteInteraction actually

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -746,6 +748,10 @@ class ThreadFragment : Fragment() {
             )
         }
 
+        getBackNavigationResult(OPEN_REACTION_BOTTOM_SHEET) { messageUid: String ->
+            navigateToEmojiPicker(messageUid, if (twoPaneViewModel.isThreadOpen) Thread else ThreadList)
```
LunarX: Is this condition truly correct on a tablet?

LunarX: Ok now I remember why it's correct but it's not intuitive. I feel like we should add a comment to make things clearer or rename the enum values. We can talk about it

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,11 +99,23 @@ class MessageController @Inject constructor(
...
-        val messages = thread.getDisplayedMessages(featureFlags, localSettings)
-        return messages.query("$isNotDraft AND $isNotScheduledDraft AND $isNotFromRealMe AND $isNotFromPlusMe").last()
-            ?: messages.query("$isNotDraft AND $isNotScheduledDraft").last()
-            ?: messages.query(isNotScheduledDraft).last()
-            ?: messages.last()
+        return messages.query("$isNotDraft AND $isNotScheduledDraft AND $isNotFromRealMe AND $isNotFromPlusMe")
+            .appendNullableExtraQuery(extraQuery).last()
+            ?: messages.query("$isNotDraft AND $isNotScheduledDraft").appendNullableExtraQuery(extraQuery).last()
+            ?: messages.query(isNotScheduledDraft).appendNullableExtraQuery(extraQuery).last()
+            ?: extraQuery?.let { messages.query(extraQuery).last() }
```
sirambd: Previously, we had the default case `messages.last()`. Why are you no longer taking it in consideration?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -91,11 +99,23 @@ class MessageController @Inject constructor(
...
+    }
+
+    suspend fun getLastMessageToExecuteAction(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message {
+        val messages = thread.getDisplayedMessages(featureFlags, this@MessageController.localSettings)
```
sirambd, line 82: Why duplication ?

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
baillyjamy: Because in the case of searching for a message where a response is possible, I don't want to receive an fallback message if I can't find one (in the case of noreply), I prefer to return null. And this behavior may or may not be necessary, regardless of whether there is an extraQuery.
Also I need to keep nullable return for getLastMessageToExecuteReaction and non null return for getLastMessageToExecuteAction.
If I use default case I lost nullable return and if I get messages in `getLastMessageToExecuteActionWithExtraQuery`, I can't have a fallback

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -78,24 +79,15 @@ class MessageController @Inject constructor(
...
+        return getLastMessageToExecuteActionWithExtraQuery(messages = messages) ?: messages.last()
+    }
 
-        val (start, end) = AccountUtils.currentMailboxEmail.getStartAndEndOfPlusEmail()
-        val isNotFromPlusMe = "SUBQUERY(${Message::from.name}, \$recipient," +
-                " \$recipient.${Recipient::email.name} BEGINSWITH '${start}'" +
-                " AND \$recipient.${Recipient::email.name} ENDSWITH '${end}'" +
-                ").@count < 1"
+    suspend fun getLastMessageToExecuteReaction(thread: Thread, featureFlags: Mailbox.FeatureFlagSet?): Message? {
+        val canBeReactedTo = "${Message::_emojiReactionNotAllowedReason.name} == null"
```
FabianDevel, line 87: Just use Message.isValidReactionTarget.

You also have the real field to use "emojiReactionNotAllowedReason" that's the enum version of the api string, and the one that should be used in the code

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -216,6 +208,34 @@ class MessageController @Inject constructor(
...
+
+        private suspend fun getLastMessageToExecuteActionWithExtraQuery(
+            messages: RealmList<Message>,
+            extraQuery: String? = null
+        ): Message? {
+            suspend fun RealmQuery<Message>.last(): Message? = sort(Message::internalDate.name, Sort.DESCENDING).first().findSuspend()
+
+            fun RealmQuery<Message>.appendNullableExtraQuery(extra: String?): RealmQuery<Message> {
+                return if (extra.isNullOrEmpty()) this else this.query(extra)
+            }
```
FabianDevel: Why not direclty append the extra query to the queryString instead of creating this ?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -695,6 +700,17 @@ class ThreadListFragment : TwoPaneFragment() {
         }
     }
 
+    private fun observePickedEmoji() {
```
FabianDevel: Maybe we could move this in an interface or fun interface to make sure all place where we want emoji to be reacted implements this observe ?

app/src/main/res/drawable/ic_face_smile.xml
```diff
@@ -0,0 +1,34 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
FabianDevel, line 18: You left out the file's header

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -53,8 +54,21 @@ class ThreadActionsViewModel @Inject constructor(
 
     private val threadUid inline get() = savedStateHandle.get<String>(ThreadActionsBottomSheetDialogArgs::threadUid.name)!!
 
-    val threadMessageToExecuteAction: SharedFlow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
+    private val threadMessageToExecuteAction: Flow<ThreadMessageInteraction.Action> = threadController.getThreadAsync(threadUid)
         .mapNotNull { it.obj?.let { thread -> getThreadAndMessageUidToExecuteAction(thread) } }
+
+    private val messageToExecuteReaction: Flow<ThreadMessageInteraction.Reaction?> = threadController.getThreadAsync(threadUid)
+        .mapNotNull { it.obj }
+        .map { getMessageToExecuteReaction(it) }
```
FabianDevel: Is there a reason you didn't write it the same way as "threadMessageToExecuteAction" ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -53,8 +54,21 @@ class ThreadActionsViewModel @Inject constructor(
 
     private val threadUid inline get() = savedStateHandle.get<String>(ThreadActionsBottomSheetDialogArgs::threadUid.name)!!
 
-    val threadMessageToExecuteAction: SharedFlow<ThreadMessageToExecuteAction> = threadController.getThreadAsync(threadUid)
+    private val threadMessageToExecuteAction: Flow<ThreadMessageInteraction.Action> = threadController.getThreadAsync(threadUid)
         .mapNotNull { it.obj?.let { thread -> getThreadAndMessageUidToExecuteAction(thread) } }
+
+    private val messageToExecuteReaction: Flow<ThreadMessageInteraction.Reaction?> = threadController.getThreadAsync(threadUid)
```
FabianDevel: Maybe rename it to threadMessageToExecuteAction to keep the same name logic as before, and the other flow

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
```diff
@@ -64,8 +78,24 @@ class ThreadActionsViewModel @Inject constructor(
...
-        val messageUid = Dispatchers.IO { messageController.getLastMessageToExecuteAction(thread, featureFlagsLive.value).uid }
-        return ThreadMessageToExecuteAction(thread, messageUid)
+    private suspend fun getThreadAndMessageUidToExecuteAction(thread: Thread): ThreadMessageInteraction.Action {
+        val messageUid = ioDispatcher { messageController.getLastMessageToExecuteAction(thread, featureFlagsLive.value).uid }
+        return ThreadMessageInteraction.Action(thread, messageUid)
+    }
+
+    private suspend fun getMessageToExecuteReaction(thread: Thread): ThreadMessageInteraction.Reaction? {
+        val messageUid = ioDispatcher { messageController.getLastMessageToExecuteReaction(thread, featureFlagsLive.value)?.uid }
+        return messageUid?.let { ThreadMessageInteraction.Reaction(it) }
```
FabianDevel: What about :
```suggestion
        return messageUid?.let(ThreadMessageInteraction::Reaction)
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -407,7 +409,7 @@ class ThreadFragment : Fragment() {
                 onEncryptionSeeConcernedRecipients = ::navigateToUnencryptableRecipients,
                 onAddReaction = {
                     trackEmojiReactionsEvent(MatomoName.OpenEmojiPicker)
-                    navigateToEmojiPicker(it.uid)
+                    navigateToEmojiPicker(it.uid, Thread)
```
FabianDevel: As Thread is a really common term in this app to talk about a lot of thing, be sure to specify the Enum class name before when you use it here and at the other usages (same for ThreadList)
```suggestion
                    navigateToEmojiPicker(it.uid, EmojiPickerObserverTarget.Thread)
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
FabianDevel:
```suggestion
            navigateToEmojiPicker(messageUid, emojiPickerObservationTarget = if (twoPaneViewModel.isThreadOpen) Thread else ThreadList)
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
baillyjamy: I can't use `isValidReactionTarget` or `emojiReactionNotAllowedReason` because I need to use realm field name for the realmquery

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
baillyjamy: Because queryString will be malformed if extra query is null.  It will throw an exception if extra query is null

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsViewModel.kt
baillyjamy: Yes, in case of Reaction, we don't get thread from getMessageToExecuteReaction

PR #2612, opened 2025-09-24
fix: Remove DotLottie library use and use Lottie main library to load .lottie files

tevincent:
depends on https://github.com/Infomaniak/android-core/pull/522


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -459,7 +456,7 @@ class NewMessageFragment : Fragment() {
         handleEditorPlaceholderVisibility()
 
         // Not showing the Lottie because it prevents the UI to settle which is breaking tests.
```
KevinBoulongne: Is this comment still needed now ?

KevinBoulongne (changes requested):

PR #2621, opened 2025-09-26
refactor: Simplify query to get remotely deleted folders

KevinBoulongne:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
```diff
@@ -200,14 +200,9 @@ class FolderController @Inject constructor(
         //region Get data
         private fun getFoldersBlocking(
             exceptionsFoldersIds: List<String> = emptyList(),
```
LunarX: We can remove the default parameter and single line everything

```suggestion
            exceptionsFoldersIds: List<String>,
```

LunarX (changes requested):

PR #2637, opened 2025-10-03
feat: Display folders correctly throughout the app

LunarX:
Only displays roles folders on the upper half of the folder lists (menu drawer, search, move fragment).
Fix indent level computation.
Fix collapse computation.
Fix subfolder to parent assignment.

This PR required to rethink the data structure used to represent displayed folders in the app. We were based on a flattened structure too early on which lost some important properties for edge cases (mainly when a subfolder has a role). I introduced a FolderUi model that uses a tree structure to represents parents and children.

PR #2640, opened 2025-10-06
feat: Update the automatic expand on empty folders code because of FolderUi and improve it

LunarX:
This check needs to be updated to take into account the new visual structure added with FolderUi when selecting who to expand automatically. 

This was the opportunity to take this code out of the FolderController which was too much logic heavy to be inside of FolderController.upsertFolders().

At the same time I added the fix for the case where all children of a parent have been removed and at least a new child has been added to the parent. Previously the new children would remain visible even though the parent would be collapsed. Now their state is update to match the one of the parent.

Depends on #2637

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -216,12 +216,6 @@ class FolderController @Inject constructor(
                 folder.isCollapsed = false
             }
         }
-
-        fun hide(folderId: String, realm: MutableRealm) {
-            realm.updateFolder(folderId) { folder ->
-                folder.isHidden = true
-            }
-        }
         //endregion
     }
 }
--- a/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
@@ -196,10 +196,6 @@ class FolderController @Inject constructor(
             realm.write { getFolderBlocking(id, realm = this)?.let { onUpdate(this, it) } }
         }
 
-        private fun MutableRealm.updateFolder(id: String, onUpdate: (Folder) -> Unit) {
-            getFolderBlocking(id, realm = this)?.let { onUpdate(it) }
-        }
-
         fun deleteSearchFolderData(realm: MutableRealm) = with(getOrCreateSearchFolder(realm)) {
             threads.clear()
         }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -566,7 +568,38 @@ class MainViewModel @Inject constructor(
...
+     */
+    private suspend fun updateFoldersCollapsedAndHiddenStates() {
+        val folders = displayedFoldersFlow.first()
+        mailboxContentRealm().write {
+            folders.custom.updateStates(realm = this)
+            folders.default.updateStates(realm = this)
+        }
+    }
+
+    private fun List<FolderUi>.updateStates(realm: MutableRealm) {
```
FabianDevel: The name is a bit too generic, collapsingState or expansionState could be better I think

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -566,7 +568,38 @@ class MainViewModel @Inject constructor(
...
+        var isCurrentParentCollapsed = false
+        forEachNestedItem { folder, _ ->
```
FabianDevel:
```suggestion
        forEachNestedItem { folderUi, _ ->
```

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -566,7 +568,38 @@ class MainViewModel @Inject constructor(
...
+            if (!mailboxContentRealm().isClosed()) {
+                folderController.update(mailbox, folders, mailboxContentRealm())
+                updateFoldersCollapsedAndHiddenStates()
+            }
+        }
+    }
+
+    /**
+     *  Recomputes the [Folder.isCollapsed] and [Folder.isHidden] states so a parent which has no more children has no more
+     *  children is correctly expanded and so a child that appeared under a collapsed parent is correctly hidden.
```
KevinBoulongne:
```suggestion
     *  Recomputes the [Folder.isCollapsed] and [Folder.isHidden] states so a parent which has no more children
     *  is correctly expanded and so a child that appeared under a collapsed parent is correctly hidden.
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
```diff
@@ -200,6 +196,10 @@ class FolderController @Inject constructor(
             realm.write { getFolderBlocking(id, realm = this)?.let { onUpdate(this, it) } }
         }
 
+        private fun MutableRealm.updateFolder(id: String, onUpdate: (Folder) -> Unit) {
+            getFolderBlocking(id, realm = this)?.let { onUpdate(it) }
+        }
+
```
FabianDevel: Put the private function at the end of its scope

KevinBoulongne (approved):

PR #2656, opened 2025-10-09
chore: Remove statusBar and navigationBar color

tevincent:


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -144,6 +142,7 @@ import kotlin.math.min
 import kotlin.math.roundToInt
 import androidx.appcompat.R as RAndroid
 
+
```
FabianDevel:
```suggestion
```

PR #2657, opened 2025-10-09
fix: Fix rare NullPointerException + log it to sentry

LouisCAD:
This fixes the regression introduced in commit 05a0bd52


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -363,8 +364,13 @@ class NewMessageViewModel @Inject constructor(
         val quote = draftInitManager.createQuote(draftMode, fullMessage, draft.attachments)
         if (quote != null) initialQuote = quote
 
+        if (fullMessage.body == null) {
+            SentryLog.e(TAG, "The message we're trying to reply to has an unexpected null body") { scope ->
+                scope.addBreadcrumb(Breadcrumb.info("message resource: ${fullMessage.resource}"))
```
FabianDevel: You can use `scope.setExtra("Message resource", fullMessage.resource)` instead, it will be easier to read in the Sentry

PR #2658, opened 2025-10-09
feat: Add 2 factor authentication confirmation

LouisCAD:
Depends on https://github.com/Infomaniak/android-core/pull/509
Depends on https://github.com/Infomaniak/android-core/pull/534

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
@@ -50,8 +50,6 @@ import com.infomaniak.core.legacy.utils.hasPermissions
 import com.infomaniak.core.matomo.Matomo.TrackerAction
 import com.infomaniak.core.observe
 import com.infomaniak.core.sentry.SentryLog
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
```diff
@@ -205,6 +207,8 @@ class MainActivity : BaseActivity() {
         enableEdgeToEdge()
 
         setContentView(binding.root)
+        val twoFactorAuthViewModel: TwoFactorAuthViewModel by viewModels()
+        addComposeOverlay { TwoFactorAuthApprovalAutoManagedBottomSheet(twoFactorAuthViewModel = twoFactorAuthViewModel) }
```
FabianDevel: Shouldn't we put this in BaseActivity directly ?

FabianDevel (changes requested):

FabianDevel (approved):

app/src/main/java/com/infomaniak/mail/ui/MainActivity.kt
LouisCAD: Not possible because this needs to be called after `setContentView`, but I moved some of it in there.

PR #2663, opened 2025-10-10
fix: Notifications not showing or a silent notification due to worker conflict

sirambd:
### 📝 **Description (Short & Clear):**

Fixed an issue where **push notifications were missing or incorrectly silenced** because `SyncMailboxesWorker` was accidentally canceling `FirebaseMessagingWorker`.

✅ Notifications now display and sound correctly.  
✅ Workers no longer interfere with each other.  
✅ Better background stability.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/utils/IFirebaseProcessNotificationsScheduler.kt
+++ b/app/src/main/java/com/infomaniak/mail/utils/IFirebaseProcessNotificationsScheduler.kt
@@ -15,17 +15,8 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
-package com.infomaniak.mail.firebase
+package com.infomaniak.mail.utils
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/workers/SyncMailboxesWorker.kt
```diff
@@ -55,22 +56,36 @@ class SyncMailboxesWorker @AssistedInject constructor(
...
     override suspend fun launchWork(): Result = withContext(ioDispatcher) {
-        SentryLog.d(TAG, "Work launched")
+
+        suspend fun shouldSkip(): Boolean {
+            return if (processMessageNotificationsWorkerScheduler.isRunning()) {
+                SentryLog.i(TAG, "Work skipped because ProcessMessageNotificationsWorker is running")
+                true
+            } else {
+                false
+            }
```
FabianDevel:
```suggestion
            return processMessageNotificationsWorkerScheduler.isRunning().also {
                if (it) SentryLog.i(TAG, "Work skipped because ProcessMessageNotificationsWorker is running")
            }
```

PR #2668, opened 2025-10-14
feat: Unfold message with new reaction

baillyjamy:
When message receive a new reaction, we mark the reaction as unseen to allow the message in question to be expanded.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/thread/ThreadEmojiReactionsComputation.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/thread/ThreadEmojiReactionsComputation.kt
@@ -46,7 +46,7 @@ private fun MutableMap<String, MutableMap<String, EmojiReactionState>>.addReacti
             emojis[emoji] = EmojiReactionState(emoji)
         }
 
-        emojis[emoji]!!.run {
+        emojis[emoji]?.run {
             addAuthor(
                 newAuthor = EmojiReactionAuthor(
                     recipient = message.from.firstOrNull() ?: continue,


app/src/main/java/com/infomaniak/mail/data/models/thread/ThreadEmojiReactionsComputation.kt
```diff
@@ -46,12 +46,15 @@ private fun MutableMap<String, MutableMap<String, EmojiReactionState>>.addReacti
             emojis[emoji] = EmojiReactionState(emoji)
         }
 
-        emojis[emoji]!!.addAuthor(
-            newAuthor = EmojiReactionAuthor(
-                recipient = message.from.firstOrNull() ?: continue,
-                sourceMessageUid = message.uid,
+        emojis[emoji]?.run {
```
LunarX: I would not modify the previous !! code. This !! was agreed upon with @sirambd because we make sure it exists on the line right above and therefore avoid introducing a meaningless condition we need to care about

app/src/main/java/com/infomaniak/mail/data/models/thread/ThreadEmojiReactionsComputation.kt
```diff
@@ -46,12 +46,15 @@ private fun MutableMap<String, MutableMap<String, EmojiReactionState>>.addReacti
...
+            addAuthor(
+                newAuthor = EmojiReactionAuthor(
+                    recipient = message.from.firstOrNull() ?: continue,
+                    sourceMessageUid = message.uid,
+                )
             )
-        )
+            isSeen = message.isSeen
```
LunarX: If multiple messages add the same emoji to the same target message, only the last `message.isSeen` will be taken into account because it overrides all previous `message.isSeen`

app/src/main/java/com/infomaniak/mail/data/models/message/EmojiReactionState.kt
```diff
@@ -27,6 +27,7 @@ class EmojiReactionState constructor() : Reaction, EmbeddedRealmObject {
     override var emoji: String = ""
     var authors: RealmList<EmojiReactionAuthor> = realmListOf()
     override var hasReacted: Boolean = false
+    var isSeen: Boolean = false
```
LunarX: With isSeen = false as a default value, all existing messages after the update will have their isSeen set to false therefore expanding every message containing an emoji reacting. I'm not sure this is the best choice

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -639,8 +639,7 @@ class ThreadViewModel @Inject constructor(
         }
         val fakedReaction = EmojiReactionStateUi(
             emoji = state.emoji,
-            authors = if (shouldFake) authors + EmojiReactionAuthorUi.FakeMe else authors,
-            hasReacted = state.hasReacted || shouldFake,
+            authors = if (shouldFake) authors + EmojiReactionAuthorUi.FakeMe else authors, hasReacted = state.hasReacted || shouldFake,
```
LunarX: I don't think you intended to push this diff of single lining args here

app/src/main/java/com/infomaniak/mail/data/models/thread/ThreadEmojiReactionsComputation.kt
```diff
@@ -46,14 +46,14 @@ private fun MutableMap<String, MutableMap<String, EmojiReactionState>>.addReacti
...
-        emojis[emoji]?.run {
+        emojis[emoji]!!.run {
             addAuthor(
                 newAuthor = EmojiReactionAuthor(
                     recipient = message.from.firstOrNull() ?: continue,
                     sourceMessageUid = message.uid,
                 )
             )
-            isSeen = message.isSeen
+            isSeen = isSeen || message.isSeen
```
LunarX: This won't accumulate values correctly. Because isSeen is initialized as true, `isSeen || message.isSeen` will always evaluate to `true` because of the first operand

app/src/main/java/com/infomaniak/mail/data/models/thread/ThreadEmojiReactionsComputation.kt
```diff
@@ -46,12 +46,15 @@ private fun MutableMap<String, MutableMap<String, EmojiReactionState>>.addReacti
...
+        emojis[emoji]!!.run {
+            addAuthor(
+                newAuthor = EmojiReactionAuthor(
+                    recipient = message.from.firstOrNull() ?: continue,
+                    sourceMessageUid = message.uid,
+                )
             )
-        )
+            isSeen = isSeen && message.isSeen
+        }
```
KevinBoulongne:
```suggestion
        emojis[emoji]?.let {
            it.addAuthor(
                newAuthor = EmojiReactionAuthor(
                    recipient = message.from.firstOrNull() ?: continue,
                    sourceMessageUid = message.uid,
                ),
            )
            it.isSeen = it.isSeen && message.isSeen
        }
```

LunarX (changes requested):
Doesn't work when I test it, maybe we can check together.

Also, when there are more than 5 messages in a thread leading to a super collapsed block, will the collapsed messages correctly be expanded if they received a new emoji reaction?

LunarX:
Ok nevermind, when I tested it I was looking at the wrong message. There are no bugs in my tests, just the issue with the super collapsed block

baillyjamy:
> Ok nevermind, when I tested it I was looking at the wrong message. There are no bugs in my tests, just the issue with the super collapsed block

I thought it was normal behavior. I fixed that.

app/src/main/java/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -424,7 +424,9 @@ class Message : RealmObject, Snoozable {
         else -> FormatedPreview.Body(preview.trim())
     }
 
-    fun shouldBeExpanded(index: Int, lastIndex: Int) = !isDraft && (!isSeen || index == lastIndex)
+    fun hasUnreadContent() = !isSeen || emojiReactions.any { !it.isSeen }
+
+    fun shouldBeExpanded(index: Int, lastIndex: Int) = !isDraft && (!isSeen || index == lastIndex || emojiReactions.any { !it.isSeen })
```
LunarX: You did not use the new `hasUnreadContent()` inside of `shouldBeExpanded()`

sirambd (commented):
Approved after the change we talked about

PR #2678, opened 2025-10-21
feat: Mailto screen animation

baillyjamy:
Add animation on select mailbox or navigation

Depends on https://github.com/Infomaniak/android-kMail/pull/2645


app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/compose/previewparameter/SelectMailboxScreenPreviewParameter.kt
```diff
@@ -28,24 +32,20 @@ class SelectMailboxScreenPreviewParameter : PreviewParameterProvider<SelectMailb
 
 data class SelectMailboxScreenDataPreview(
     val usersWithMailboxes: List<UserMailboxesUi>,
-    val selectedMailboxUi: SelectedMailboxUi?,
-    val selectingAnotherUser: Boolean
+    val uiState: StateFlow<UiState>,
```
LunarX: I'd not bother creating state flows in order to be able to collect them and have a state. You can have a standard variable like we do in other preview parameters because we won't be using the flow capabilities here

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -66,132 +73,200 @@ fun SelectMailboxScreen(
...
+        onChooseAnotherMailbox = viewModel::chooseAnotherMailbox,
         onNavigationTopbarClick = onNavigationTopbarClick,
-        onContinue = onContinue
+        onContinueWithMailbox = onContinue
     )
 }
 
 @Composable
-fun SelectMailboxScreen(
+fun SelectMailboxScreenContent(
```
LunarX: This method should be private

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -66,132 +73,200 @@ fun SelectMailboxScreen(
...
+@Composable
+private fun TopButton(
+    modifier: Modifier,
+    uiState: () -> UiState,
+    onContinueWithMailbox: (SelectedMailboxUi) -> Unit
+) {
+    val selectedMailbox by remember { derivedStateOf { (uiState() as? UiState.SelectMailbox)?.selectedMailbox?.value } }
+
+    LargeButton(
+        modifier = modifier.fillMaxWidth(),
```
LunarX: No need for the fillMaxWidth(). It' already present in the argument modifier

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
```diff
@@ -111,4 +127,10 @@ class SelectMailboxViewModel @Inject constructor(
         val avatarUrl: String?,
         val initials: String,
     )
+
+    sealed interface UiState {
+        data object Loading : UiState
+        data class DefaultMailbox(val defaultMailbox: SelectedMailboxUi) : UiState
+        data class SelectMailbox(val selectedMailbox: State<SelectedMailboxUi?>) : UiState
```
LunarX: The goal of having a UiState like this is that you can have a single state in your ui. Here `selectedMailbox` can be simplified by creating another UiState instead of having to bother with a `State<SelectedMailboxUi?>`. This will help you avoid having to deal with two different states and deal will nullable values

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
```diff
@@ -39,8 +42,11 @@ class SelectMailboxViewModel @Inject constructor(
     private val _usersWithMailboxes = MutableStateFlow<List<UserMailboxesUi>>(emptyList())
     val usersWithMailboxes: StateFlow<List<UserMailboxesUi>> get() = _usersWithMailboxes
 
-    private val _selectedMailbox = MutableStateFlow<SelectedMailboxUi?>(null)
-    val selectedMailbox get() = _selectedMailbox
+    private val _uiState = MutableStateFlow<UiState>(UiState.Loading)
+    val uiState = _uiState.asStateFlow()
+    val selectedMailboxState = mutableStateOf<SelectedMailboxUi?>(null)
```
LunarX: This looks dangerous. Also you should not need two objects in order to hold the same data

LunarX (changes requested):
Some first comments, I didn't check everything for now but we could take a look together

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -66,132 +73,200 @@ fun SelectMailboxScreen(
...
+        onMailboxSelected(null)
     }
 
     BottomStickyButtonScaffold(
         snackbarHostState = snackbarHostState,
         topBar = {
             MailTopAppBar(
                 navigationIcon = {
-                    TopAppBarButtons.Close(onNavigationTopbarClick)
+                    if (uiState() is UiState.SelectMailbox) {
```
LunarX: You can make this one a when case

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -66,132 +73,200 @@ fun SelectMailboxScreen(
...
+    uiState: () -> UiState,
+    usersWithMailboxes: List<UserMailboxesUi>,
+    onMailboxSelected: (SelectedMailboxUi?) -> Unit
+) {
+    uiState().let { uiState ->
+        LazyColumn(
+            modifier = Modifier
+                .padding(horizontal = Margin.Medium)
+                .fillMaxWidth()
+                .weight(1f),
```
LunarX: It feels easier to follow around and more appropriate to have these layout procurations be defined at the callplace by providing a Modifier instance. This way the column is the only one that manages the layout aspect and the `ScrollableContent` and ScrollableContent's responsability can be reduced

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -66,132 +73,200 @@ fun SelectMailboxScreen(
...
+    Image(
+        imageVector = ImageVector.vectorResource(R.drawable.illustration_mailbox_ellipsis_bubble),
+        contentDescription = null
+    )
+    Text(
+        style = Typography.h2,
+        maxLines = 1,
+        text = stringResource(R.string.composeMailboxCurrentTitle)
     )
 }
```
LunarX, line 152: Haven't you lost the spacing between items you used to have when you extracted this?

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
```diff
@@ -84,11 +90,21 @@ class SelectMailboxViewModel @Inject constructor(
                 avatarUrl = currentUser.avatar,
                 initials = currentUser.getInitials()
             )
+            defaultMailbox = selectedMailbox
+            _uiState.value = UiState.DefaultMailbox(selectedMailbox)
+        }
+    }
+
+    fun chooseAnotherMailbox(choosingAnotherMailbox: Boolean) {
```
LunarX: Maybe we can find a better name for this method? I don't know

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -64,136 +70,198 @@ fun SelectMailboxScreen(
...
-                        }
-                        Spacer(modifier = Modifier.weight(1f))
-                        selectedMailbox?.let {
-                            SelectedMailboxIndicator(
-                                modifier = Modifier.padding(Margin.Medium),
-                                selectedMailbox = it
-                            )
-                        }
+            }
+            UiState.SelectionScreen.NoSelection, is UiState.SelectionScreen.Selected -> items(
```
LunarX: You can simplify this when case by just checking the for a UiState.SelectionScreen

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -64,136 +72,198 @@ fun SelectMailboxScreen(
...
-            style = ButtonType.Tertiary,
-            onClick = {
-                selectingAnotherUser.value = true
-                onMailboxSelected(null)
-            }
-        )
+    val selectedMailbox by remember { derivedStateOf { (uiState() as? UiState.SelectionScreen.Selected)?.mailbox } }
+
+    BackHandler(uiState() is UiState.SelectionScreen) {
+        showSelectionScreen(false)
```
LunarX: Maybe it'd be easier to have the BackHandler defined outside this inner method. Do you have an opinion?

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -64,136 +72,198 @@ fun SelectMailboxScreen(
...
+        showSelectionScreen = viewModel::showSelectionScreen,
         onNavigationTopbarClick = onNavigationTopbarClick,
-        onContinue = onContinue
+        onContinueWithMailbox = onContinue
     )
 }
 
 @Composable
-fun SelectMailboxScreen(
+private fun SelectMailboxScreenContent(
```
LunarX: Maybe we'd need to check how we usually call our other screen methods but I think the inner method usually has the same name as the outer

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
```diff
@@ -64,136 +72,198 @@ fun SelectMailboxScreen(
...
                 )
-                Text(
-                    style = Typography.h2,
-                    maxLines = 1,
-                    text = stringResource(R.string.composeMailboxCurrentTitle)
+                SelectedMailboxBottom(selectedMailbox)
+            }
+        },
+        topButton = { modifier ->
+            TopButton(modifier, uiState, onContinueWithMailbox)
```
LunarX: I'd probably call this compose method something more meaningful like `ContinueButton()` when possible instead of a generic "TopButton" which is already visible in the named parameter.

Same for the bottom one, maybe something like `DifferentAddressButton()`?

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
```diff
@@ -84,11 +87,21 @@ class SelectMailboxViewModel @Inject constructor(
                 avatarUrl = currentUser.avatar,
                 initials = currentUser.getInitials()
             )
+            defaultMailbox = selectedMailbox
+            _uiState.value = UiState.DefaultScreen(selectedMailbox)
+        }
+    }
+
+    fun showSelectionScreen(show: Boolean) {
```
LunarX: Ah I just understood what you meant when you talked about "show". If you want a better boolean name you can go with something like "isShown" or "isDisplayed" maybe

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
baillyjamy: I don't really like doing that. I like the name of my method to remind me where I am. In the definition of my component, perhaps with the definition of state/viewModel, or in the independent definition of the content that can be instantiated without runtime. The keyword content allows me to make this distinction, but if it's like that throughout the project, that's fine.
I renamed the method.

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
baillyjamy: Thanks bro !

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxScreen.kt
baillyjamy: Yep good idea

LunarX (approved):

PR #2683, opened 2025-10-22
refactor: Remove add mailbox button

FabianDevel:
Since the back has changed how mailbox authentification works, there is no more mailbox password.

So we needed to remove all the occurrence in the app that allowed to add a mailbox with a password, or update this password


app/src/main/java/com/infomaniak/mail/ui/noValidMailboxes/NoValidMailboxesFragment.kt
```diff
@@ -65,14 +59,8 @@ class NoValidMailboxesFragment : Fragment(), MailboxListFragment {
...
-
-        invalidPasswordMailboxesRecyclerView.adapter = MailboxesAdapter(
-            isInMenuDrawer = isInMenuDrawer,
-            hasValidMailboxes = hasValidMailboxes,
-            onInvalidPasswordMailboxClicked = { mailbox -> onInvalidPasswordMailboxClicked(mailbox) },
-        )
-
-        lockedMailboxesRecyclerView.adapter = MailboxesAdapter(isInMenuDrawer, hasValidMailboxes)
+        invalidPasswordMailboxesRecyclerView.adapter = InvalidMailboxesAdapter()
+        lockedMailboxesRecyclerView.adapter = InvalidMailboxesAdapter()
```
tevincent, line 63: Does `InvalidMailboxesAdapter` is the right name for invalid and locked mailboxes ?

FabianDevel: Yes it is, it's the same that manage both

PR #2689, opened 2025-10-27
fix: Fix the UnsupportedOperationException on system initiated animations

LunarX:
Like in #2599, this custom compose view had a crash because of system animations. Because the previous fix worked, we saw that this view still had issues so this PR fixes it the same way.

Fixes MAIL-ANDROID-P6M

### DIFF
--- a/EmojiComponents/src/main/java/com/infomaniak/emojicomponents/views/EmojiReactionsView.kt
+++ b/EmojiComponents/src/main/java/com/infomaniak/emojicomponents/views/EmojiReactionsView.kt
@@ -66,8 +66,6 @@ class EmojiReactionsView @JvmOverloads constructor(
 
     init {
         val composeView = ComposeView(context)
-        composeView.setContent { ComposeViewContent() }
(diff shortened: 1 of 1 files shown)


EmojiComponents/src/main/java/com/infomaniak/emojicomponents/views/EmojiReactionsView.kt
```diff
@@ -62,12 +65,16 @@ class EmojiReactionsView @JvmOverloads constructor(
     private var addReactionDisabledColor: Int? = null
 
     init {
+        val composeView = ComposeView(context)
+        composeView.setContent { ComposeViewContent() }
+        addView(composeView)
```
FabianDevel: I think you should end the init with these lines. 
It's both more "logical" to add the view in last, and setViewCompositionStrategy should be declared before setContent from what I understand. 
Because the strategy seems to be applicated when setContent is called. (At least in the doc, it is always called before setContent)

PR #2736, opened 2025-12-03
feat: Unsubscribe alert

baillyjamy:
image

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
@@ -577,7 +577,7 @@ class ThreadAdapter(
 
         message.draftResource?.let { draftResource ->
             scheduleAlert.onAction1 {
-                trackScheduleSendEvent(MatomoName.ModifySnooze)
+                trackMessageBannerEvent(MatomoName.ReprogramSchedule)
                 threadAdapterCallbacks?.onRescheduleClicked?.invoke(
                     draftResource,
                     message.displayDate.takeIf { message.isScheduledDraft }?.epochSeconds?.times(1_000),
@@ -586,7 +586,7 @@ class ThreadAdapter(
         }
 
         scheduleAlert.onAction2 {
-            trackScheduleSendEvent(MatomoName.CancelSnooze)
+            trackMessageBannerEvent(MatomoName.ModifySchedule)
             threadAdapterCallbacks?.onModifyScheduledClicked?.invoke(message)
         }
     }
--- a/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
+++ b/app/src/main/java/com/infomaniak/mail/MatomoMail.kt
@@ -283,6 +283,7 @@ object MatomoMail : Matomo {
         ReplyNo("replyNo"),
         ReplyYes("replyYes"),
         ReportJunk("reportJunk"),
+        ReprogramSchedule("reprogramSchedule"),
         RequestPassword("requestPassword"),
         Restore("restore"),
         RestoreEmails("restoreEmails"),


app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -669,6 +683,28 @@ class ThreadViewModel @Inject constructor(
...
+        }
+    }
+    //endregion
+
+    data class CombineMessageToBuildMessageUi(
+        val mode: ThreadOpeningMode,
+        val featureFlags: Mailbox.FeatureFlagSet,
+        val fakeReactions: Map<String, Set<String>>,
+        val messageUidUnsubscribed: List<String>
+    )
```
LunarX: Can be private

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -669,6 +683,28 @@ class ThreadViewModel @Inject constructor(
...
 
+    //region Unsubscribe list diffusion
+    fun unsubscribeMessage(message: Message) {
+        viewModelScope.launch {
+            with(ApiRepository.unsubscribe(message.resource)) {
+                if (isSuccess()) {
+                    snackbarManager.postValue(appContext.getString(R.string.snackbarUnsubscribeSuccess))
+                    messageUidUnsubscribed.value += message.uid
+                } else {
+                    snackbarManager.postValue(appContext.getString(R.string.snackbarUnsubscribeFailure))
```
LunarX: When there's an error it'd be great to translate the error because I think this could return some generic errors already translate inside `translateError()` to the user. Maybe this needs to be looked into.

Right now, the wording seems off, if you simply timed out, we're saying to the user that unsubscribing from this email is not possible. I feel like unsubscribing *is* possible, we simply failed to unsubscribe. That's how I see it.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -582,6 +585,17 @@ class ThreadAdapter(
         }
     }
 
+    private fun ItemMessageBinding.bindUnsubscribe(messageUi: MessageUi) {
+        if (messageUi.hasUnsubscribeButton) {
+            unsubscribeAlert.isVisible = true
+            unsubscribeAlert.setDescription(context.resources.getString(R.string.messageComesFromDiscussionList))
+            unsubscribeAlert.setAction1Text(context.resources.getString(R.string.unsubscribeButtonTitle))
+            unsubscribeAlert.onAction1 { threadAdapterCallbacks?.unsubscribeClicked?.invoke(messageUi.message) }
```
LunarX: It would be nice to make the action button start loading while we make the api call so if you have a slow internet connection it will show you something is happening instead of letting you wondering if something is happening at all.

I think we already did the alert button loading using `showAction1Progress()` for the snooze scheduling alert for example

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -998,14 +1011,14 @@ class ThreadAdapter(
                 fun areDifferent(message: T, other: T) = areTheSame(message, other).not()
             }
 
-            object MessageDiffAspect {
+            interface MessageDiffAspect {
```
LunarX: No need to make it an interface if you only did this to create a companion object because variables at the top level of an object are already static just like a companion object

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -1018,13 +1031,19 @@ class ThreadAdapter(
...
                     // check for anything that doesn't need to handle bind with precision using a custom payload
-                    body?.value == oldMessage.body?.value &&
-                            splitBody == oldMessage.splitBody &&
-                            shouldHideDivider == oldMessage.shouldHideDivider
+                    message.body?.value == oldMessage.message.body?.value &&
+                            message.splitBody == oldMessage.message.splitBody &&
+                            message.shouldHideDivider == oldMessage.message.shouldHideDivider
                 })
+
+                data object Unsubscribe : DiffAspect<MessageUi>({ hasUnsubscribeButton == it.hasUnsubscribeButton })
```
LunarX: Maybe keep "AnythingElse" as the last defined DiffAspect so "anything else" makes sense when you read it in order I think

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -1044,6 +1063,7 @@ class ThreadAdapter(
         var navigateToNewMessageActivity: ((Uri) -> Unit)? = null,
         var navigateToAttendeeBottomSheet: ((List<Attendee>) -> Unit)? = null,
         var navigateToDownloadProgressDialog: ((Attachment, AttachmentIntentType) -> Unit)? = null,
+        var unsubscribeClicked: ((Message) -> Unit)? = null,
```
LunarX: We usually try to call such click callbacks as "on...Clicked" when possible if you look at the other next to it

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadFragment.kt
```diff
@@ -370,6 +370,8 @@ class ThreadFragment : Fragment(), PickerEmojiObserver {
                 navigateToDownloadProgressDialog = { attachment, attachmentIntentType ->
                     navigateToDownloadProgressDialog(attachment, attachmentIntentType, ThreadFragment::class.java.name)
                 },
+                unsubscribeClicked = threadViewModel::unsubscribeMessage,
+                //unsubscribeClicked = { message -> threadViewModel.unsubscribeMessage(message) },
```
LunarX: You probably don't need that one anymore
```suggestion
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -582,6 +585,17 @@ class ThreadAdapter(
         }
     }
 
+    private fun ItemMessageBinding.bindUnsubscribe(messageUi: MessageUi) {
+        if (messageUi.hasUnsubscribeButton) {
+            unsubscribeAlert.isVisible = true
+            unsubscribeAlert.setDescription(context.resources.getString(R.string.messageComesFromDiscussionList))
+            unsubscribeAlert.setAction1Text(context.resources.getString(R.string.unsubscribeButtonTitle))
```
LunarX: This looks useless, isn't setting this inside the xml enough?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -170,6 +172,11 @@ class ThreadViewModel @Inject constructor(
     // instantly apply clicked emojis without having to wait for the API call to return.
     private val fakeReactions = MutableStateFlow<Map<String, Set<String>>>(emptyMap())
 
+    // Each time a message is unsubscribed, it is added to the list and the list is resent to update the MessageUi flow.
+    // This unsubscribed message flow is listened in the combine to avoid losing the status of the unsubscribed message
+    // if something else changes and rebuilds the MessageUi list.
```
LunarX:
```suggestion
    // Each time a message is unsubscribed from, it is added to the list and the list is sent again to update the MessageUi flow.
    // This unsubscribed message flow is collected in the combine to avoid losing the status of the unsubscribed message
    // if something else changes and rebuilds the MessageUi list.
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -621,11 +629,17 @@ class ThreadViewModel @Inject constructor(
...
     ): List<Any> = map { item ->
         if (item is Message) {
             val localReactions = fakeReactions[item.messageId] ?: emptySet()
             val reactions = item.emojiReactions.toFakedReactions(localReactions)
-            MessageUi(item, reactions, isReactionsAvailable)
+            MessageUi(
+                message = item,
+                emojiReactionsState = reactions,
+                isReactionsFeatureAvailable = isReactionsAvailable,
+                hasUnsubscribeButton = item.hasUnsubscribeLink == true && messageUidUnsubscribed.none { it == item.uid }
```
LunarX: Does this `hasUnsubscribeLink` really need to be nullable? Why is it even null to begin with? What does `false` and `null` even represent exactly?

app/src/main/res/drawable/ic_user_double_filled.xml
```diff
@@ -0,0 +1,5 @@
...
+    
+</vector>
```
LunarX: You can reformat the file and add the copyright

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
baillyjamy: We discussed it in person. Nothing will change. Test nullability of hasUnsubscribeLink is good in viewmodel

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -582,6 +585,20 @@ class ThreadAdapter(
         }
     }
 
+    private fun ItemMessageBinding.bindUnsubscribe(messageUi: MessageUi) {
+        when (messageUi.unsubscribeState) {
+            is MessageUi.UnsubscribeState.CanUnsubscribe -> {
```
FabianDevel: You can import MessageUi.UnsubscribeState to simplify the readability
We got enough hint here that the state comes from messageUi

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -582,6 +585,20 @@ class ThreadAdapter(
...
+                unsubscribeAlert.isVisible = true
+                unsubscribeAlert.hideAction1Progress(R.string.unsubscribeButtonTitle)
+                unsubscribeAlert.onAction1 { threadAdapterCallbacks?.unsubscribeClicked?.invoke(messageUi.message) }
```
FabianDevel: You could use an apply

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -582,6 +585,20 @@ class ThreadAdapter(
...
+            }
+            is MessageUi.UnsubscribeState.InProgress -> unsubscribeAlert.showAction1Progress()
+            is MessageUi.UnsubscribeState.Completed, null -> {
+                unsubscribeAlert.isVisible = false
+            }
```
FabianDevel: Oneline this

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -621,11 +636,18 @@ class ThreadViewModel @Inject constructor(
     private suspend fun <E : Any> List<E>.toUiMessages(
         fakeReactions: Map<String, Set<String>>,
         isReactionsAvailable: Boolean,
+        unsubscribeStateByMessageUid: Map<String, MessageUi.UnsubscribeState>,
     ): List<Any> = map { item ->
         if (item is Message) {
             val localReactions = fakeReactions[item.messageId] ?: emptySet()
             val reactions = item.emojiReactions.toFakedReactions(localReactions)
-            MessageUi(item, reactions, isReactionsAvailable)
+            val canUnsubscribeOrNull = if (item.hasUnsubscribeLink == true) MessageUi.UnsubscribeState.CanUnsubscribe else null
```
FabianDevel: Same as before, import MessageUi.UnsubscribteState

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -669,6 +691,29 @@ class ThreadViewModel @Inject constructor(
         return fakedReaction
     }
 
+    //region Unsubscribe list diffusion
+    fun unsubscribeMessage(message: Message) {
+        viewModelScope.launch {
```
FabianDevel: You could save an indentation like this (we do this a lot in viewModels)
```suggestion
    fun unsubscribeMessage(message: Message) =  viewModelScope.launch {
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -669,6 +691,29 @@ class ThreadViewModel @Inject constructor(
...
+            unsubscribeStateByMessageUid.value = unsubscribeStateByMessageUid.value.toMutableMap().apply {
+                set(message.uid, MessageUi.UnsubscribeState.InProgress)
+            }
+            with(ApiRepository.unsubscribe(message.resource)) {
```
FabianDevel: Remove your `with`, and set a `val apiResponse =`
Your with here is only needed for isSuccess, and is less readable in this case than apiResponse variable.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -669,6 +691,29 @@ class ThreadViewModel @Inject constructor(
         return fakedReaction
     }
 
+    //region Unsubscribe list diffusion
+    fun unsubscribeMessage(message: Message) {
+        viewModelScope.launch {
+            unsubscribeStateByMessageUid.value = unsubscribeStateByMessageUid.value.toMutableMap().apply {
+                set(message.uid, MessageUi.UnsubscribeState.InProgress)
+            }
```
FabianDevel, line 705:
```suggestion
            }
            
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
```diff
@@ -669,6 +692,27 @@ class ThreadViewModel @Inject constructor(
...
+                set(message.uid, UnsubscribeState.Completed)
+            }
+        } else {
+            snackbarManager.postValue(appContext.getString(R.string.snackbarUnsubscribeFailure))
+            unsubscribeStateByMessageUid.value = unsubscribeStateByMessageUid.value.toMutableMap().apply {
+                set(message.uid, UnsubscribeState.CanUnsubscribe)
+            }
+        }
+    }
+    //endregion
```
LunarX, line 717: This could've maybe been simplified a bit with something along those lines:


```kt
    fun unsubscribeMessage(message: Message) = viewModelScope.launch {
        setUnsubscribeState(message, UnsubscribeState.InProgress)

        val apiResponse = ApiRepository.unsubscribe(message.resource)
        if (apiResponse.isSuccess()) {
            snackbarManager.postValue(appContext.getString(R.string.snackbarUnsubscribeSuccess))
            setUnsubscribeState(message, UnsubscribeState.Completed)
        } else {
            snackbarManager.postValue(appContext.getString(R.string.snackbarUnsubscribeFailure))
            setUnsubscribeState(message, UnsubscribeState.CanUnsubscribe)
        }
    }

    private fun setUnsubscribeState(message: Message, state: UnsubscribeState) {
        unsubscribeStateByMessageUid.value = unsubscribeStateByMessageUid.value.toMutableMap().apply {
            set(message.uid, state)
        }
    }
```

LunarX (approved):
Looks good to me, there's just a quick refactor to avoid unnecessary duplication and help when reading the code.

I'm still unsure on how much we want to merge this half working feature, but apart from the error handling, the code is good

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadViewModel.kt
baillyjamy: We will be able to manage errors better when the backend is better at returning errors related to this route. In the meantime, this is the best we can do.

PR #2741, opened 2025-12-05
fix: Fix fourth onboarding animation sometime disappearing when swiping too quickly

LunarX:
The view appears to be 0dp in width once in every 10 times or so when scrolling fast to the fourth page after having closed the app completely. Requesting layout again seems to prevent this somehow


tevincent (approved):

tevincent (approved):

LunarX:
The width is usually still 0 by this point

PR #2742, opened 2025-12-05
feat: Collapse child folder

baillyjamy:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
```diff
@@ -50,11 +50,9 @@ class FolderController @Inject constructor(
     }
 
     fun getMenuDrawerCustomFoldersAsync(): Flow<ResultsChange<Folder>> {
-        val rootsQuery = " AND $isRootFolder"
-        val typeQuery = " AND ${Folder.rolePropertyName} == nil"
```
LunarX, line 54: We should be able to keep the filter to only query folders with no roles, no?

app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -1755,3 +1756,7 @@ private val HIDDEN_ROLES_WHEN_EMPTY = setOf(FolderRole.SCHEDULED_DRAFTS, FolderR
 private fun Flow<List<Folder>>.removeRolesThatHideWhenEmpty(): Flow<List<Folder>> = map {
     it.filterNot { folder -> folder.role in HIDDEN_ROLES_WHEN_EMPTY && folder.threads.isEmpty() }
 }
+
+private fun Flow<List<Folder>>.keepTopLevelFolders(): Flow<List<Folder>> = map {
+    it.filter { folder -> folder.parent == null && folder.role == null }
+}
```
LunarX, line 1762: It doesn't feel like it makes senses to filter out roles folders in a method that says "keep top level folders"

app/src/main/java/com/infomaniak/mail/data/models/FolderUi.kt
```diff
@@ -22,7 +22,9 @@ data class FolderUi(
     override var children: List<FolderUi>,
     val depth: Int,
     var canBeCollapsed: Boolean, // For parents only (only a parent can be collapsed, its children will be hidden instead)
-    var isHidden: Boolean,
+    var isCollapsed: Boolean,
```
LunarX: `isCollapsed` should not be needed anymore, the logic on main should still be enough for the collapsed state computation to work like it has always been

app/src/main/java/com/infomaniak/mail/utils/FolderUiUtils.kt
```diff
@@ -45,32 +45,28 @@ fun List<Folder>.toFolderUiTree(isInDefaultFolderSection: Boolean): List<FolderU
...
+            .mapNotNull { folderToFolderUi[it] }
 
-        val shouldHide = if (folderUi.isRoot) {
-            isCurrentParentCollapsed = folder.isCollapsed
-            false
-        } else {
-            isCurrentParentCollapsed
+        val isCurrentFolderCollapsed = folder.isCollapsed
+        for (child in validChildren) {
+            child.isParentCollapsed = isCurrentFolderCollapsed || folderUi.isParentCollapsed
```
LunarX: Maybe this is the one we should have renamed? Like, "a child is hidden if the current folder is collapsed or if one of its ancestors is already hidden". `child.isHidden = isCurrentFolderCollapsed || folderUi.isHidden`.

What do you think about it?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/FolderController.kt
baillyjamy: But what do we do with a child folder with role if I keep typeQuery condition ?

PR #2746, opened 2025-12-09
feat: Reset isCollapse state for folder that loosing children

baillyjamy:


app/src/main/java/com/infomaniak/mail/ui/MainViewModel.kt
```diff
@@ -596,11 +596,9 @@ class MainViewModel @Inject constructor(
...
     private fun List<FolderUi>.updateCollapsedState(realm: MutableRealm) {
         forEachNestedItem { folderUi, _ ->
-            if (folderUi.isRoot) {
-                // If we detect that a folder doesn't have any children anymore, if it was collapsed, automatically expand it
-                val collapseStateNeedsReset = folderUi.children.isEmpty()
-                if (collapseStateNeedsReset) FolderController.expand(folderUi.folder.id, realm)
-            }
+            // If we detect that a folder doesn't have any children anymore, if it was collapsed, automatically expand it
+            val collapseStateNeedsReset = folderUi.children.isEmpty()
+            if (collapseStateNeedsReset) FolderController.expand(folderUi.folder.id, realm)
```
LunarX, line 601: Maybe just to be sure we can check if expansion is required?

PR #2762, opened 2025-12-24
fix: Remove useless property since multiPassword

JorisBodin:


app/src/main/java/com/infomaniak/mail/data/cache/RealmMigrations.kt
```diff
@@ -302,6 +303,20 @@ private fun MigrationContext.initMessagesWithContentToTheOldMessagesListAfterThi
...
 //endregion
 
+// Migrate from version #13
+private fun MigrationContext.keepDefaultValuesAfterTwelveMigration() {
+
+    if (oldRealm.schemaVersion() <= 13L) {
+        enumerate(className = "Mailbox") { oldObject: DynamicRealmObject, newObject: DynamicMutableRealmObject? ->
+            newObject?.apply {
+                // Rename property without losing its previous value
+                set(propertyName = "isLocked", value = oldObject.getValue<Boolean>(fieldName = "_isLocked"))
```
sirambd, line 313: You should use `getValueOrNull` instead of `getValue` because this property has already undergone a migration previously, and to avoid any potential crashes.

PR #2766, opened 2025-12-29
feat: Move AppVersionCheckingTest into core module

baillyjamy:
Depends on https://github.com/Infomaniak/android-core/pull/649
Depends on https://github.com/Infomaniak/android-kMail/pull/2764


FabianDevel (approved):

baillyjamy:
> Shouldn't you now use core.mockk.agent and core.mockk.android in build.gradle now that you moved it to core ?

You're right

PR #2770, opened 2026-01-07
feat: Remove junk menu duplicate

solrubado:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MailActionsBottomSheetDialog.kt
```diff
@@ -163,12 +168,17 @@ abstract class MailActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         isVisible = canBeReactedTo
     }
 
+    fun setJunkMessageUi() = with(binding) {
```
JorisBodin: I would give it a clearer function name that indicates that it hides them.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MailActionsBottomSheetDialog.kt
```diff
@@ -163,12 +168,17 @@ abstract class MailActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         isVisible = canBeReactedTo
     }
 
+    fun hideReportJunkButtons() = with(binding) {
+        spam.isGone = true
+        phishing.isGone = true
+        blockSender.isGone = true
+    }
```
LunarX: If this method is only used by MessageActionsBottomSheetDialog, can probably keep it private and define it inside the only class that uses it

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MailActionsBottomSheetDialog.kt
```diff
@@ -177,7 +187,9 @@ abstract class MailActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         fun onModifySnooze()
         fun onCancelSnooze()
         fun onFavorite()
-        fun onReportJunk()
+        fun onSpam() = Unit
+        fun onPhishing() = Unit
+        fun onBlockSender() = Unit
```
LunarX: No need for default implementations here

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -191,10 +259,28 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
         hideFirstActionItemDivider()
     }
 
+    private fun setJunkUi() = with(binding) {
+        val isFromSpam = mainViewModel.currentFolder.value?.role == FolderRole.SPAM
+        spam.apply {
+            val (text, icon) = if (isFromSpam) {
+                R.string.actionNonSpam to R.drawable.ic_non_spam
+            } else {
+                R.string.actionSpam to R.drawable.ic_report_junk
```
LunarX: The icon is not correct. You are using the report junk icon instead of the spam icon. Also if the report_junk icon is not used anymore in the app we can remove it altogether

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -191,10 +259,28 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+            }
```
LunarX: I see we have to setup the ui dynamically for all three junk items in two different places. We can extract a setJunkUi method and reuse it in both places

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -191,10 +259,28 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+
+            setTitle(text)
+            setIconResource(icon)
+            isVisible = true
+        }
+
+        phishing.isVisible = !isFromSpam
+        blockSender.isVisible = !isFromSpam
```
LunarX: There used to be other conditions to hide/show the block sender. For exemple the option is not visible when the only email address that can be blocked is yourself. Now you show this button regardless of this condition

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -105,17 +126,32 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
+    private fun observeReportPhishingResult() {
+        mainViewModel.reportPhishingTrigger.observe(viewLifecycleOwner) {
+            descriptionDialog.resetLoadingAndDismiss()
+            findNavController().popBackStack()
+        }
+    }
+
+    private fun setBlockUserUi(potentialUsersToBlock: Map<Recipient, Message>) {
+        if (potentialUsersToBlock.count() == 0) binding.blockSender.isGone = true
     }
```
LunarX: Here you compute the blockSender visibility in two different spots which is error prone and could easily lead to hidden bugs when refactoring the code or moving the code around. 

Also the 1st spot computes a condition A and the 2nd spot computes a condition B but the 1st spot doesn't care about condition B and the 2nd spot doesn't care about condition A. So only one of the two condition is effective

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -214,16 +250,51 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
+                mainViewModel.toggleThreadSpamStatus(listOf(thread.uid))
             } else {
-                findNavController().navigate(
-                    resId = R.id.junkBottomSheetDialog,
-                    args = JunkBottomSheetDialogArgs(arrayOf(thread.uid)).toBundle(),
+                // Check the first message, because it is not possible to select messages from multiple folders,
+                // so you won't have both SPAM and non-SPAM messages.
+                trackBottomSheetThreadActionsEvent(
+                    MatomoName.Spam,
+                    value = thread.messages.firstOrNull()?.folder?.role == FolderRole.SPAM
```
LunarX: Can the condition for the matomo event be hardcoded to `false` here or is it different?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -158,18 +172,65 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+            val messages = junkMessagesViewModel.junkMessages.value ?: emptyList()
+
+            if (messages.isEmpty()) {
+                //An error will be shown to the user in the reportPhishing function
+                SentryLog.e(TAG, getString(R.string.errorPhishingMessagesNull))
+            }
+
+            descriptionDialog.show(
+                title = getString(R.string.reportPhishingTitle),
+                description = resources.getQuantityString(R.plurals.reportPhishingDescription, threadsCount),
```
LunarX: I see that the quantity to select the singular or plural form of the string used to be computed based on the number of messages. Here you've changed it to the number of threads which is not the same. If you have a single thread with 2 messages, the string should be in plural to match the previous logic

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -158,18 +172,65 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+                onPositiveButtonClicked = {
+                    mainViewModel.reportPhishing(
+                        threadsUids,
+                        messages
+                    )
+                },
```
LunarX: This is more of a detail on the style of writing but we usually tend to one-line lines like those. If you look at what was there previously it's actually how it used to be written.
```suggestion
                onPositiveButtonClicked = { mainViewModel.reportPhishing(threadsUids, messages) },
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -79,6 +86,10 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
     @Inject
     lateinit var localSettings: LocalSettings
 
+    @Inject
+    lateinit var snackbarManager: SnackbarManager
+
+
```
LunarX: Here we usually avoid double new lines in this situation

app/src/main/res/values/strings_notranslatable.xml
```diff
@@ -0,0 +1,21 @@
...
+  ~ but WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  ~ GNU General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  -->
+<resources>
+    <string name="errorPotentialUsersToBlockNull" translatable="false">Potential users to block is null</string>
+    <string name="errorPhishingMessagesNull" translatable="false">Phishing messages report failed because the list is empty</string>
```
LunarX: Could we maybe name all sentry related strings by prefixing them with `sentry...` to better group them and distinguish them from other non-sentry string that have `translatable="false"` ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -158,18 +172,65 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+            isMultiSelectOn = false
+        }
+
+        binding.phishing.setOnClickListener {
+            trackMultiSelectActionEvent(MatomoName.SignalPhishing, threadsCount, isFromBottomSheet = true)
+            val messages = junkMessagesViewModel.junkMessages.value ?: emptyList()
+
+            if (messages.isEmpty()) {
+                //An error will be shown to the user in the reportPhishing function
+                SentryLog.e(TAG, getString(R.string.errorPhishingMessagesNull))
```
LunarX: Should we not avoid showing the `descriptionDialog` if we detect that messages are empty and we already know that the dialog can never succeed?

LunarX: The string id mentions that list is "null" but the list is not null, it's empty. Which is also what the string says.

Maybe we can rename this string id

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -259,27 +263,14 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
         }
 
         setSnoozeUi(threads)
-        setJunkUi()
+        ThreadActionsBottomSheetDialog.setSpamPhishingUi(
+            binding.spam,
+            binding.phishing,
+            mainViewModel.currentFolder.value?.role == FolderRole.SPAM
+        )
```
LunarX, line 272: We like naming arguments in situations like these where what argument is what is not clear from the provided values. (Especially the 3rd one here)

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MailActionsBottomSheetDialog.kt
```diff
@@ -168,10 +168,22 @@ abstract class MailActionsBottomSheetDialog : ActionsBottomSheetDialog() {
...
+            }
+
+            setTitle(text)
+            setIconResource(icon)
+            isVisible = true
+        }
+
+        phishing.isVisible = !isFromSpam
+        blockSender.isVisible = !isFromSpam
     }
```
LunarX: This method doesn't seem to be used

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -255,11 +232,9 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
                 trackBottomSheetThreadActionsEvent(MatomoName.Spam, value = true)
                 mainViewModel.toggleThreadSpamStatus(listOf(thread.uid))
             } else {
-                // Check the first message, because it is not possible to select messages from multiple folders,
-                // so you won't have both SPAM and non-SPAM messages.
                 trackBottomSheetThreadActionsEvent(
                     MatomoName.Spam,
-                    value = thread.messages.firstOrNull()?.folder?.role == FolderRole.SPAM
+                    value = false
```
LunarX: You can one-line this like line 232 I think

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -97,21 +97,19 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
                 folderRole = folderRoleUtils.getActionFolderRole(thread)
                 isFromArchive = folderRole == FolderRole.ARCHIVE
                 isFromSpam = folderRole == FolderRole.SPAM
-
+                junkMessagesViewModel.threadsUids = listOf(thread.uid)
```
LunarX: This is a detail but I'd probably let this line be in its own little block of code to separate it visually from the reste because it serves a very different purpose in the code

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -158,18 +175,61 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+                    resId = R.id.userToBlockBottomSheetDialog,
+                    substituteClassName = ThreadListFragment::class.java.name,
+                )
+            } else {
+                potentialUsersToBlock.values.firstOrNull()?.let { message ->
+                    junkMessagesViewModel.messageOfUserToBlock.value = message
+                }
+            }
+            mainViewModel.isMultiSelectOn = false
+
```
FabianDevel, line 221: We avoid adding a blank line before the closing curly bracket. 
But after closing a block, you can add one if you feel that it make the code a bit more readable (for exemple like you did at line 223) 
```suggestion
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -158,18 +175,61 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
+        }
+
+        binding.favorite.setClosingOnClickListener(shouldCloseMultiSelection = true)
+        {
```
FabianDevel, line 225:
```suggestion
        binding.favorite.setClosingOnClickListener(shouldCloseMultiSelection = true) {
```

We don't put opening curly bracket on a new line

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -158,18 +175,61 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
...
             trackMultiSelectActionEvent(MatomoName.Favorite, threadsCount, isFromBottomSheet = true)
             toggleThreadsFavoriteStatus(threadsUids, shouldFavorite)
             isMultiSelectOn = false
         }
 
-        binding.saveKDrive.setClosingOnClickListener(shouldCloseMultiSelection = true) {
+        binding.saveKDrive.setClosingOnClickListener(shouldCloseMultiSelection = true)
+        {
```
FabianDevel, line 232: same as above

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -51,13 +59,14 @@ import com.infomaniak.mail.utils.extensions.safeNavigateToNewMessageActivity
...
 import javax.inject.Inject
+import com.infomaniak.core.R as RCore
 
 @AndroidEntryPoint
 class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
 
     private val navigationArgs: ThreadActionsBottomSheetDialogArgs by navArgs()
     private val threadActionsViewModel: ThreadActionsViewModel by viewModels()
-
+    private val junkMessagesViewModel: JunkMessagesViewModel by activityViewModels()
```
FabianDevel, line 69:
```suggestion
    private val junkMessagesViewModel: JunkMessagesViewModel by activityViewModels()
    
```

We usually add a blank line between the "special" delegates (binding, navargs, viewmodels) and the lazy values. 
Then another blank line between the lazy and the "classic" properties

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -77,24 +86,36 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
 
+    @Inject
+    lateinit var snackbarManager: SnackbarManager
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
         threadActionsViewModel.threadMessagesWithActionAndReaction
             .observe(viewLifecycleOwner) { (thread, messageUidToExecuteAction, messageUidToReactTo) ->
+                // Initialization of threadsUids to populate messages and potentialUsersToBlock
```
FabianDevel, line 97:
```suggestion
                // Initialization of threadsUids to populate junkMessages and potentialUsersToBlock
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -90,9 +101,15 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
         val threads = selectedThreads.toSet()
         val threadsUids = threads.map { it.uid }
         val threadsCount = threadsUids.count()
+
+        // Initialization of threadsUids to populate messages and potentialUsersToBlock
```
FabianDevel, line 105:
```suggestion
        // Initialization of threadsUids to populate junkMessages and potentialUsersToBlock
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -77,24 +86,36 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
+                junkMessagesViewModel.threadsUids = listOf(thread.uid)
+
                 folderRole = folderRoleUtils.getActionFolderRole(thread)
                 isFromArchive = folderRole == FolderRole.ARCHIVE
                 isFromSpam = folderRole == FolderRole.SPAM
-
```
FabianDevel, line 88: Keep this line, it helps readibility

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -77,24 +86,36 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
 
                 initOnClickListener(onActionClick(thread, messageUidToExecuteAction, messageUidToReactTo))
             }
+
+        junkMessagesViewModel.potentialBlockedUsers.observe(viewLifecycleOwner) { potentialUsersToBlock ->
+            setBlockUserUi(binding.blockSender, potentialUsersToBlock, isFromSpam)
+        }
+
+        observeReportPhishingResult()
+
```
FabianDevel, line 118:
```suggestion
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -77,24 +86,36 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
                 setSnoozeUi(thread.isSnoozed())
                 setReactionUi(canBeReactedTo = messageUidToReactTo != null)
+                setSpamPhishingUi(binding.spam, binding.phishing, isFromSpam)
 
                 initOnClickListener(onActionClick(thread, messageUidToExecuteAction, messageUidToReactTo))
             }
+
+        junkMessagesViewModel.potentialBlockedUsers.observe(viewLifecycleOwner) { potentialUsersToBlock ->
+            setBlockUserUi(binding.blockSender, potentialUsersToBlock, isFromSpam)
+        }
```
FabianDevel, line 115: Why did you create a function for the observe of phishing results, but not for this one (or the contrary) ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -214,16 +229,45 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
...
+                mainViewModel.toggleThreadSpamStatus(listOf(thread.uid))
+            }
+        }
+
+        override fun onPhishing() {
+            trackBottomSheetThreadActionsEvent(MatomoName.SignalPhishing)
+            descriptionDialog.show(
+                title = getString(R.string.reportPhishingTitle),
+                description = resources.getQuantityString(R.plurals.reportPhishingDescription, thread.messages.count()),
+                onPositiveButtonClicked = { mainViewModel.reportPhishing(listOf(thread.uid), thread.messages) },
```
FabianDevel, line 247: I think you should use junkMessagesViewModel.junkMessages here instead of the thread.messages. 

Because an action on the thread should only affect the last message (that is not sent by you) of this thread, not all the messages in it.

PR #2780, opened 2026-01-15
fix: Remove messages duplicates when moving and marking as spam

solrubado:


app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -105,22 +105,26 @@ class MessageController @Inject constructor(
...
     }
 
-    private suspend fun getMessagesAndTheirDuplicates(thread: Thread, query: String): List<Message> {
+    private suspend fun getMessagesFromThread(thread: Thread, query: String, includeDuplicates: Boolean): List<Message> {
         val messages = thread.messages.query(query).findSuspend()
-        val duplicates = thread.duplicates.query("${Message::messageId.name} IN $0", messages.map { it.messageId }).findSuspend()
-        return messages + duplicates
+        if (includeDuplicates) {
+            val duplicates =
+                thread.duplicates.query("${Message::messageId.name} IN $0", messages.map { it.messageId }).findSuspend()
```
FabianDevel:
```suggestion
            val duplicates = thread.duplicates.query(
                "${Message::messageId.name} IN $0", 
                messages.map { it.messageId }).findSuspend(),
            )
              
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -216,11 +220,12 @@ class MessageController @Inject constructor(
             messages: RealmList<Message>,
             extraQuery: String? = null
         ): Message? {
-            suspend fun RealmQuery<Message>.last(): Message? = sort(Message::internalDate.name, Sort.DESCENDING).first().findSuspend()
+            suspend fun RealmQuery<Message>.last(): Message? =
+                sort(Message::internalDate.name, Sort.DESCENDING).first().findSuspend()
```
FabianDevel: When we can't inline a function, we add a body and return 
```suggestion
            suspend fun RealmQuery<Message>.last(): Message? {
                return sort(Message::internalDate.name, Sort.DESCENDING).first().findSuspend()
            }
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/MessageController.kt
```diff
@@ -105,22 +105,26 @@ class MessageController @Inject constructor(
     }
 
     suspend fun getFavoriteMessages(thread: Thread): List<Message> {
-        return getMessagesAndTheirDuplicates(thread, "${Message::isFavorite.name} == true")
+        return getMessagesFromThread(thread, "${Message::isFavorite.name} == true", true)
```
FabianDevel: We try to avoid leaving boolean parameter without its name.
Add it everywhere in this PR 

```suggestion
        return getMessagesFromThread(thread, "${Message::isFavorite.name} == true", includesDuplicates = true)
```

PR #2783, opened 2026-01-19
fix: Fix mailto missmatch between identityId and mailboxUuid

LunarX:
When selecting a custom mailbox during a mailto, we used the `identityId` from the selected mailbox but not its `uuid` when making the api call to save the draft on the api.

This fixes this issue by also providing the correct `mailboxId` which is used to infer the mailbox `uuid` for the api call.

At the same time, make sure the `DraftActionWorker`s can't be canceled if a worker for a `mailboxId` is scheduled while a worker for another `mailboxId` was running.

This also showed that there was an issue when scheduling the draft action worker where it did not adapt to the currently selected user id. This was fixed as well

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/data/models/mailbox/Mailbox.kt
+++ b/app/src/main/java/com/infomaniak/mail/data/models/mailbox/Mailbox.kt
@@ -101,7 +101,7 @@ class Mailbox : RealmObject {
     @Transient
     var signatures = realmListOf<Signature>()
     @Transient
-    var haveSignatureNeverBeenFetched = true
+    var haveSignaturesBeenFetched = false
     @Transient
     var _featureFlags = realmSetOf<String>()
         private set
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
@@ -110,7 +110,7 @@ class SelectMailboxViewModel @Inject constructor(
 
     suspend fun ensureMailboxIsFetched(userId: Int, mailboxId: Int) {
         val mailbox = mailboxController.getMailbox(userId, mailboxId) ?: return
-        if (mailbox.haveSignatureNeverBeenFetched) {
+        if (!mailbox.haveSignaturesBeenFetched) {
             SharedUtils.updateSignatures(mailbox, mailboxInfoRealm, AccountUtils.getHttpClient(userId))
         }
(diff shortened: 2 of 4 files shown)


app/src/main/java/com/infomaniak/mail/data/models/mailbox/Mailbox.kt
```diff
@@ -101,6 +101,8 @@ class Mailbox : RealmObject {
     @Transient
     var signatures = realmListOf<Signature>()
     @Transient
+    var haveSignatureNeverBeenFetched = true
```
FabianDevel: I know it would put a '!' in the conditon, but what about removing the negation in this name : 
`haveSignaturesBeenFetched` ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/selectMailbox/SelectMailboxViewModel.kt
```diff
@@ -96,14 +100,23 @@ class SelectMailboxViewModel @Inject constructor(
...
         }
     }
 
     fun selectMailbox(selectedMailbox: SelectedMailboxUi) {
         _uiState.value = UiState.SelectionScreen.Selected(selectedMailbox)
     }
 
+    suspend fun ensureMailboxIsFetched(userId: Int, mailboxId: Int) {
+        val mailbox = mailboxController.getMailbox(userId, mailboxId) ?: return
+        if (mailbox.haveSignatureNeverBeenFetched) SharedUtils.updateSignatures(
```
FabianDevel:
```suggestion
        if (mailbox.haveSignatureNeverBeenFetched) {
            SharedUtils.updateSignatures(
```

app/src/main/java/com/infomaniak/mail/utils/SharedUtils.kt
```diff
@@ -150,8 +151,8 @@ class SharedUtils @Inject constructor(
 
     companion object {
 
-        suspend fun updateSignatures(mailbox: Mailbox, customRealm: Realm): Int? {
-            return with(ApiRepository.getSignatures(mailbox.hostingId, mailbox.mailboxName)) {
+        suspend fun updateSignatures(mailbox: Mailbox, customRealm: Realm, okHttpClient: OkHttpClient? = null): Int? {
+            return with(ApiRepository.getSignatures(mailbox.hostingId, mailbox.mailboxName, okHttpClient)) {
```
FabianDevel: This return is useless, and maybe we could refactor this without the `with`like we do now

FabianDevel (changes requested):

PR #2803, opened 2026-02-04
chore: Replace ioCoroutineContext where it's not needed

LouisCAD:


LouisCAD:
After review with @benjaminVadon and @sirambd, our `coroutineContext` function was sneakily also adding a `CoroutineExceptionHandler` that reports uncaught exceptions (and ignores some Realm-related ones), without letting the app crash.

We need to change the naming/API of this, and also improve it to have helpful reports over Sentry.

PR #2806, opened 2026-02-09
fix: Fix move navigation crash again

solrubado:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/MultiSelectBottomSheetDialog.kt
```diff
@@ -106,45 +106,16 @@ class MultiSelectBottomSheetDialog : ActionsBottomSheetDialog() {
 
         // Initialization of threadsUids to populate junkMessages and potentialUsersToBlock
         junkMessagesViewModel.threadsUids = threadsUids
-
```
FabianDevel, line 109: Revert this

PR #2810, opened 2026-02-13
fix: Ensure the collect for the networkAvailability is always called when lifecycle is STARTED

tevincent:


app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListFragment.kt
```diff
@@ -552,13 +552,15 @@ class ThreadListFragment : TwoPaneFragment(), PickerEmojiObserver {
 
     private fun observeNetworkStatus() {
         viewLifecycleOwner.lifecycleScope.launch {
-            mainViewModel.isNetworkAvailable.collect { isNetworkAvailable ->
-                if (_binding == null) return@collect
+            repeatOnLifecycle(State.STARTED) {
+                mainViewModel.isNetworkAvailable.collect { isNetworkAvailable ->
+                    if (_binding == null) return@collect
```
benjaminVadon: Normally, now we are based on State.STARTED, the _binding should never be null, no ?

benjaminVadon (approved):

PR #2818, opened 2026-02-23
feat: Add support for Unknown recipient for My ksuite instead of external

benjaminVadon:


app/src/main/res/values/strings.xml
```diff
@@ -17,11 +17,9 @@
   -->
 <resources>
     <string name="appName" translatable="false">Infomaniak Mail</string>
-
```
FabianDevel, line 20: Revert this

app/src/main/res/values-de/strings.xml
```diff
@@ -663,7 +663,9 @@
     <string name="tooManyRecipients">Sie können diese Adresse nicht hinzufügen, da Sie die Höchstzahl an Empfängern erreicht haben</string>
     <string name="trashFolder">Papierkorb</string>
     <string name="unblockButton">Entsperren</string>
+    <string name="unknownDialogTitleRecipient">Unbekannter Empfänger</string>
```
FabianDevel, line 669: It now misses the other languages

app/src/main/java/com/infomaniak/mail/data/models/mailbox/Mailbox.kt
```diff
@@ -178,6 +178,8 @@ class Mailbox : RealmObject {
         return@with !areNotificationsEnabled() || isGroupBlocked || isChannelBlocked
     }
 
+    fun hasOrganisation(): Boolean = kSuite !is KSuite.Perso
```
FabianDevel: Either directly use the check in the SubjectData initialization, or put this with other kSuite extension in the KSuite File ?

FabianDevel (changes requested):
You're missing the translation of the banner in the newMessageActivity
Also, maybe the description should change to avoid talking about organization ?

app/src/main/res/values/strings.xml
benjaminVadon: Ok, but shouldn't we move the non translatable strings with other non translatable, in another PR ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/SubjectFormatter.kt
```diff
@@ -63,29 +64,38 @@ class SubjectFormatter @Inject constructor(private val appContext: Context) {
         )
         if (externalRecipientQuantity == 0) return previousContent
 
-        return postFixWithExternal(previousContent, externalRecipientQuantity, externalRecipientEmail, onExternalClicked)
-    }
+        return postFixWithExternal(
+            previousContent = previousContent,
+            tagRes = if (hasOrganisation) R.string.externalTag else R.string.unknownTag,
+            onExternalClicked = {
+                val title = if (hasOrganisation) R.string.externalDialogTitleRecipient else R.string.unknownDialogTitleRecipient
```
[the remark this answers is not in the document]
benjaminVadon: @FabianDevel it seems that copilot points an existing  incoherence in the previous code about the naming between expeditor and recipient, don't you think it too ? 
I'm not as comfortable about this code as you so I think you'll more able to judge it

FabianDevel (approved):

PR #2821, opened 2026-02-27
feat: Move AppLock from legacy

baillyjamy:
Apply change after moving Applock form legacy

Depends on https://github.com/Infomaniak/android-core/pull/708


app/src/main/AndroidManifest.xml
```diff
@@ -126,7 +126,7 @@
             android:theme="@style/WebViewTheme" />
 
         <activity
-            android:name="com.infomaniak.core.legacy.applock.LockActivity"
+            android:name="com.infomaniak.core.applock.LockActivity"
             android:exported="false"
             android:launchMode="singleTop"
             android:theme="@style/AppThemeLock" />
```
benjaminVadon, line 132: why not declare the activity in the module directly ?

benjaminVadon (approved):

PR #2849, opened 2026-03-31
chore: Add agents.md

FabianDevel:


AGENTS.md
```diff
@@ -0,0 +1,186 @@
...
+
+## 1. Project Summary
+
+**Infomaniak Mail** is an Android email client application built by Infomaniak Network SA, featuring a modern UI with Jetpack
+Compose, Realm database, and multi-account support with push notifications (Google Play variant).
+
+### High-Level Tech Stack
+
+- **Language**: Kotlin (Java 17, JVM target 17)
+- **Platform**: Android (minSdk 27, targetSdk 35, compileSdk 36)
```
benjaminVadon: Should we add version in this file ? as it need update, maybe it's simplier to point to the file where the versions are set.

AGENTS.md
```diff
@@ -0,0 +1,186 @@
...
+
+## 3. Local Norms
+
+### Architecture & Design
+
+- **MVVM Pattern**: Each screen has a Fragment/Activity + ViewModel
+- **Repository Pattern**: Data access abstracted through controller classes (e.g., `ThreadController`)
+- **Dependency Injection**: Use Hilt `@Inject` constructor for dependencies
+- **Database**: All models extend `RealmObject` or `EmbeddedRealmObject`
+- **SOLID**: Keep classes focused; MainActivity is currently 682 lines (do not grow further)
```
benjaminVadon: why this class precisely ? is it not a common pattern for most files ?

AGENTS.md
sirambd: It's true that these values can change, so it's better for the AI to retrieve them on its own, otherwise, if an update is released, you might forget to update it.

AGENTS.md
```diff
@@ -0,0 +1,186 @@
...
+### High-Level Tech Stack
+
+- **Language**: Kotlin (Java 17, JVM target 17)
+- **Platform**: Android (minSdk 27, targetSdk 35, compileSdk 36)
+- **Build System**: Gradle with Kotlin DSL (build.gradle.kts)
+- **Architecture**: MVVM with Repository pattern, multi-module structure
+- **Dependency Injection**: Dagger Hilt
+- **Database**: Realm (RealmObject for models)
+- **UI Framework**: Hybrid - Jetpack Compose + XML Views with ViewBinding
+- **Network**: Retrofit/Ktor with Kotlin Serialization
```
sirambd: C'est faux ça

AGENTS.md
FabianDevel: @copilot

app/AGENTS.md
```diff
@@ -0,0 +1,205 @@
...
+- **Network**: Ktor with Kotlin Serialization
+- **Navigation**: Android Navigation Component with Safe Args
+- **Crash Reporting**: Sentry
+- **Analytics**: Matomo
+
+## Context Map
+
+```
+app/src/main/java/com/infomaniak/mail/
+├── MainApplication.kt          # Entry point, must call InfomaniakCore.init()
```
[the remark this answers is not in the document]
FabianDevel: @copilot fix this

AGENTS.md
```diff
@@ -43,22 +43,22 @@ android-kMail/
...
+./gradlew :Core:Legacy:assemble :Core:Legacy:Confetti:assemble
 
 # Run all tests
-./gradlew :app:testStandardDebugUnitTest && ./gradlew Core:test
+./gradlew :app:testStandardDebugUnitTest && ./gradlew :Core:Legacy:test :Core:Legacy:Confetti:test
 
-# Lint Core (uses ktlint in Core module)
-./gradlew Core:ktlintCheck
+# Lint included Core modules (uses ktlint)
+./gradlew :Core:Legacy:ktlintCheck :Core:Legacy:Confetti:ktlintCheck
```
FabianDevel, line 63: @copilot We use composite build, you should put a reference to the :build-logic that is also needed to build core

FabianDevel: @copilot We use composite build, you should put a reference to the :build-logic that is also needed to build core

tevincent (approved):

PR #2864, opened 2026-04-14
feat: Auto-fill URL field when adding a link from selected text in the editor

amariaux:
Depends on : https://github.com/Infomaniak/android-rich-html-editor/pull/55


amariaux:
wrong branch

PR #2873, opened 2026-04-20
feat: Allow pasting multiple recipients at once

Elouan1411:


app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
```diff
@@ -404,6 +415,137 @@ class RecipientFieldView @JvmOverloads constructor(
...
+            getContactName(email) to email
+        }
+
+        addMultipleRecipients(recipientsToAdd)
+
+        if (invalidEmails.isNotEmpty()) {
+            snackbarManager.setValue(
+                context.resources.getQuantityString(
+                    R.plurals.addMultipleInvalidEmails,
+                    invalidEmails.size,
```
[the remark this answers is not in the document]
Elouan1411: The desired behavior is already correct.

app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareTextInput.kt
```diff
@@ -28,13 +29,33 @@ class BackspaceAwareTextInput @JvmOverloads constructor(
...
     private var backspaceOnEmptyField: () -> Unit = {}
+    private var onPasteIntercept: ((String) -> Boolean)? = null
 
     override fun onKeyDown(keyCode: Int, event: KeyEvent?): Boolean {
         if (keyCode == KeyEvent.KEYCODE_DEL && text.isNullOrEmpty()) backspaceOnEmptyField()
         return super.onKeyDown(keyCode, event)
     }
 
+    override fun onTextContextMenuItem(id: Int): Boolean {
+        if (id == android.R.id.paste) {
```
FabianDevel: Import this id as RAndroid.id

app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareTextInput.kt
```diff
@@ -28,13 +29,33 @@ class BackspaceAwareTextInput @JvmOverloads constructor(
...
+            val clipboard = context.getSystemService(Context.CLIPBOARD_SERVICE) as ClipboardManager
+            val clip = clipboard.primaryClip
+
+            if (clip != null && clip.itemCount > 0) {
+                val pastedText = clip.getItemAt(0).text?.toString() ?: ""
+                if (onPasteIntercept?.invoke(pastedText) == true) {
+                    return true
+                }
```
FabianDevel:
```suggestion
                if (onPasteIntercept?.invoke(pastedText) == true) 
 return true
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareTextInput.kt
```diff
@@ -28,13 +29,33 @@ class BackspaceAwareTextInput @JvmOverloads constructor(
...
+            }
+        }
+        return super.onTextContextMenuItem(id)
```
FabianDevel, line 51: We usually add a blank line after the end of a scope, and before a return. (Except if the returned value is computed just the line above)

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactChipAdapter.kt
```diff
@@ -74,6 +74,14 @@ class ContactChipAdapter(
         }
     }
 
+    fun addChips(newRecipients: List<Recipient>): Int {
+        var added = 0
+        newRecipients.forEach { recipient -> if (recipients.add(recipient)) added++ }
+        if (added > 0) notifyItemRangeInserted(itemCount - added, added)
+
+        return added
+    }
```
FabianDevel: Why do you compute this return value but never use it ? 

Also, a more "advanced" kotlin way of doing this could be : 

```suggestion
    fun addChips(newRecipients: List<Recipient>): Int {
        return newRecipients.fold(initial = 0) { added, recipient -> 
            if (recipients.add(recipient)) added + 1 else added 
        }
    }
```

This has the advantadge to scope you local variable, and avoid a var
This can protect a bit your code from error (assigning a value to a var that you shouldn't, using a variable that's living longer than it should etc.)

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
```diff
@@ -163,19 +164,17 @@ class RecipientFieldView @JvmOverloads constructor(
...
+                    addRecipientsFromInput(input = input)
                 },
                 snackbarManager = snackbarManager,
                 getAddressBookWithGroup = { getAddressBookWithGroup?.invoke(it) },
             )
 
             contactChipAdapter = ContactChipAdapter(
-                openContextMenu = ::showContactContextMenu,
-                onBackspace = { recipient ->
+                openContextMenu = ::showContactContextMenu, onBackspace = { recipient ->
```
FabianDevel: Apart from a onlined function, we never put more than one parameter on the same line


```suggestion
                openContextMenu = ::showContactContextMenu, 
                onBackspace = { recipient ->
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
```diff
@@ -379,21 +394,17 @@ class RecipientFieldView @JvmOverloads constructor(
...
 
     private fun addRecipient(email: String, name: String) {
-
         if (!email.isEmail()) {
             snackbarManager.setValue(context.getString(R.string.addUnknownRecipientInvalidEmail))
             return
         }
 
-        if (contactChipAdapter.itemCount > MAX_ALLOWED_RECIPIENT) {
+        if (contactChipAdapter.itemCount >= MAX_ALLOWED_RECIPIENT) {
```
FabianDevel, line 404: Why did you add this ?

FabianDevel (changes requested):
First part of review

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactChipAdapter.kt
Elouan1411: I computed this value to align with the addChip function which also provides information on whether it was added. Should I remove it or keep it?

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
Elouan1411: I don't know if it's intentional, but I noticed that using only > allows adding one extra element (if MAX_ALLOWED_RECIPIENT = 99, you can add 100 recipients).

app/src/main/java/com/infomaniak/mail/ui/newMessage/BackspaceAwareTextInput.kt
```diff
@@ -37,17 +38,16 @@ class BackspaceAwareTextInput @JvmOverloads constructor(
     }
 
     override fun onTextContextMenuItem(id: Int): Boolean {
-        if (id == android.R.id.paste) {
+        if (id == Randroid.id.paste) {
```
FabianDevel: RAndroid

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
```diff
@@ -171,7 +171,8 @@ class RecipientFieldView @JvmOverloads constructor(
             )
 
             contactChipAdapter = ContactChipAdapter(
-                openContextMenu = ::showContactContextMenu, onBackspace = { recipient ->
+                openContextMenu = ::showContactContextMenu,
+                onBackspace = { recipient ->
                     removeRecipient(recipient)
                     focusTextField()
                 })
```
FabianDevel: Put the parenthesis at a new line to keep the indentation logical

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactChipAdapter.kt
FabianDevel: You can remove it, we try to keep the code the more simple it needs to be. This means not making things unless you need it. 
If we ever touch this code again and then need this computation, then we'll add it at this moment. 
It won't take a lot of time

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
FabianDevel: Yeah that checks out, I don't know why it was like this 😕 
Did you test pasting with 99 recipients and 100 ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
```diff
@@ -404,6 +416,137 @@ class RecipientFieldView @JvmOverloads constructor(
...
+                emailsToAdd.add(email)
+            } else {
+                invalidEmails.add(email)
+            }
+        }
+
+        val recipientsToAdd = emailsToAdd.map { email ->
+            getContactName(email) to email
+        }
+
```
FabianDevel: You could avoid passing a list of pair with unnamed parameter of the same type (which can lead to error if you switch the two parameter as the compiler won't say a thing). 
You can do it like this : 

```suggestion
        val recipientsToAdd = emailsToAdd.map { emailToAdd ->
            Recipient().initLocalValues(name = getContactName(emailToAdd), email = emailToAdd)
        }

```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactChipAdapter.kt
Elouan1411: ok

app/src/main/java/com/infomaniak/mail/ui/newMessage/RecipientFieldView.kt
Elouan1411: the test proceeds without any problems

FabianDevel (approved):

PR #2880, opened 2026-04-22
fix: Correct editor action layout

Elouan1411:


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageEditorManager.kt
```diff
@@ -151,7 +149,6 @@ class NewMessageEditorManager @Inject constructor(private val insertLinkDialog:
 
     enum class EditorAction(val matomoName: MatomoName) {
         ATTACHMENT(MatomoName.ImportFile),
-        CAMERA(MatomoName.ImportFromCamera),
```
LunarX, line 154: We should remove this ImportFromCamera if this is not used anymore and update the matomo xls accordingly

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageEditorManager.kt
Elouan1411: ok

LunarX (approved):

PR #2883, opened 2026-04-23
feat: Make possible to report a thread or message as phishing from spam folder

solrubado:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ThreadActionsBottomSheetDialog.kt
```diff
@@ -334,7 +336,7 @@ class ThreadActionsBottomSheetDialog : MailActionsBottomSheetDialog() {
                 isVisible = true
             }
 
-            phishing.isVisible = !isFromSpam
+            phishing.isVisible = true
```
FabianDevel: Is it necessary ? It seems visible by default

FabianDevel (changes requested):

FabianDevel (approved):

PR #2894, opened 2026-05-05
fix: Ensure creation of 'Recipient' with valid email

NicolasBourdin88:


app/src/main/java/com/infomaniak/mail/data/models/correspondent/Recipient.kt
```diff
@@ -57,8 +57,8 @@ open class Recipient : EmbeddedRealmObject, Correspondent, ContactAutocompletabl
 
     override var contactId = name + email
 
-    fun initLocalValues(email: String? = null, name: String? = null, hasExternalProvider: Boolean? = null): Recipient {
-        email?.let { this.email = it }
+    private fun initLocalValues(email: String, name: String? = null, hasExternalProvider: Boolean? = null): Recipient {
+        email.let { this.email = it }
```
LunarX: No need to use let here anymore

app/src/main/java/com/infomaniak/mail/data/models/correspondent/Recipient.kt
```diff
@@ -103,5 +102,14 @@ open class Recipient : EmbeddedRealmObject, Correspondent, ContactAutocompletabl
...
             parcel.customWriteBoolean(hasExternalProvider)
         }
+
+        fun createValidRecipientOrNull(
+            email: String,
+            name: String? = null,
+            hasExternalProvider: Boolean? = null
+        ): Recipient? {
+            if (!email.isEmail()) return null
+            return Recipient().initLocalValues(email, name, hasExternalProvider)
```
FabianDevel, line 112:
```suggestion
            return Recipient().initLocalValues(email, name, hasExternalProvider).takeif { email.isEmail() }
```

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/correspondent/Recipient.kt
LunarX: I do not agree, this instantiates a Recipient everytime even when we wont use it. `if (!email.isEmail()) return null` is already good enough so I prefer the more instantiation optimal version.

LunarX (approved):

FabianDevel (approved):

PR #2907, opened 2026-05-15
fix(webview): use evaluateJavascript in MessageDisplayWebViewClient

### LINKED ISSUE #2906: MessageDisplayWebViewClient still uses loadUrl("javascript:...") instead of evaluateJavascript
`MessageDisplayWebViewClient.onPageFinished` (`app/src/main/java/com/infomaniak/mail/ui/main/thread/webViewClient/MessageDisplayWebViewClient.kt:90`) runs script in the message WebView like this:

```kotlin
val escapedMessageUid = looselyEscapeAsStringLiteralForJs(messageUid)
webView.loadUrl("javascript:removeAllProperties(); normalizeMessageWidth($widthInDp, $escapedMessageUid)")
```

The legacy `javascript:` URL form has two known drawbacks compared to `WebView.evaluateJavascript(...)`:

1. It is recorded as a navigation step in the WebView's history, so `goBack()` can land on the synthetic JS entry instead of the previous message.
2. There is no way to read a return value back into Kotlin. Today this call ignores the result, but any future tweak that wants to wait for the script to finish (or read its return value) can't be expressed on top of `loadUrl("javascript:...")` and has to migrate anyway.

Android's WebView docs and lint (`WebViewLegacy`) both treat `loadUrl("javascript:...")` as deprecated since `WebView.evaluateJavascript` was added in API 19, which is well below this project's `minSdk`.

### Suggested fix

Swap to `evaluateJavascript` and keep the same script:

```kotlin
webView.evaluateJavascript(
    "removeAllProperties(); normalizeMessageWidth($widthInDp, $escapedMessageUid)",
    null
)
```

This is a pure swap. No callback is needed since the existing call discards the result.

A PR with the one-line change is open at #2907.

jim-daf:
Closes #2906.

`MessageDisplayWebViewClient.onPageFinished` runs:

`webView.loadUrl(\"javascript:removeAllProperties(); normalizeMessageWidth(\$widthInDp, \$escapedMessageUid)\")`

`loadUrl("javascript:...")` is the legacy WebView path: it records a synthetic history entry (so `goBack()` can land on it) and discards any return value because there is no callback. `WebView.evaluateJavascript` was added in API 19, which is well below the project's minSdk, and does neither of those.

### Change

`- webView.loadUrl("javascript:removeAllProperties(); normalizeMessageWidth(\$widthInDp, \$escapedMessageUid)\")`
`+ webView.evaluateJavascript("removeAllProperties(); normalizeMessageWidth(\$widthInDp, \$escapedMessageUid)\", null)`

No callback is added because the existing call already discards the result. Same script, same effect, no history pollution.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/webViewClient/MessageDisplayWebViewClient.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/webViewClient/MessageDisplayWebViewClient.kt
@@ -87,7 +87,7 @@ class MessageDisplayWebViewClient(
                 }
             }
             val escapedMessageUid = looselyEscapeAsStringLiteralForJs(messageUid)
-            webView.loadUrl("javascript:removeAllProperties(); normalizeMessageWidth($widthInDp, $escapedMessageUid)")
+            webView.evaluateJavascript("removeAllProperties(); normalizeMessageWidth($widthInDp, $escapedMessageUid)", null)
             onPageFinished?.invoke()
         }
     }


LunarX:
Thank you but this is pretty small of a difference and pretty nitpicky. Right now the code doesn't use any of the properties that make this solution better and changing between the two approaches is trivial if ever needed one day

PR #2917, opened 2026-05-28
fix(UserProfile): Update core to remove useless fields in User model that can break the connection

FabianDevel:


LouisCAD (approved):

FabianDevel:
Done in #2919

PR #2952, opened 2026-06-23
feat: Mentions in emails

solrubado:


app/src/main/res/raw/mention_deletion_observer.js
```diff
@@ -0,0 +1,76 @@
...
+            refsCollection.push(node.getAttribute("data-ik-mention-ref"));
+        }
+
+        if (node.nodeType === Node.ELEMENT_NODE) {
+            const mentions = node.querySelectorAll("[data-ik-mention-ref]");
+            mentions.forEach((mention) => {
+                refsCollection.push(mention.getAttribute("data-ik-mention-ref"));
+            });
+        }
+    };
```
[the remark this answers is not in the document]
solrubado: We are checking if node.hasAttribute("data-ik-mention-ref") before pushing, so it can't be null

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -1072,6 +1088,19 @@ class NewMessageViewModel @Inject constructor(
...
+        val mentionTags = getElementsByAttribute(MENTION_ATTR)
+        mentionTags.forEach { it ->
+            it.attr(
+                "style",
+                "padding: 0 4px; border-radius: 100px; " +
+                        "color: var(--mail-content-mention-text-color, #333); " +
+                        "background-color: var(--mail-content-mention-background-color, #f1f1f1); " +
+                        "font-weight: var(--mail-content-mention-font-weight, inherit)"
+            )
+        }
```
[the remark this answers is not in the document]
solrubado: We want to override the whole style, the backend is only expecting this style.

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/SnoozeRefreshStrategy.kt
```diff
@@ -64,7 +64,7 @@ val snoozeRefreshStrategy = object : DefaultRefreshStrategy {
         impactedFolders += FolderRole.INBOX
     }
 
-    override fun processDeletedThread(thread: Thread, realm: MutableRealm) = thread.recomputeThread()
+    override fun processDeletedThread(thread: Thread, realm: MutableRealm, mailbox: Mailbox?) = thread.recomputeThread()
```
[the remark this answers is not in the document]
Elouan1411: Is there any reason not to pass realm as a paramete?

app/src/main/java/com/infomaniak/mail/ui/main/folder/ThreadListAdapter.kt
```diff
@@ -317,6 +317,7 @@ class ThreadListAdapter @Inject constructor(
             iconForward.isVisible = isIconForwardVisible
 
             iconAttachment.isVisible = hasAttachable
+            iconMention.isVisible = hasMentions && !isSeen
```
Elouan1411: Rename `hasMention` and remove redundant condition since it already checks each message for seen status

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
```diff
@@ -154,7 +155,9 @@ class ContactAdapter(
         }
     }
 
-    override fun getItemCount(): Int = matchedContacts.count() + if (displayAddUnknownContactButton) 1 else 0
+    override fun getItemCount(): Int {
+        return matchedContacts.count() + if (displayAddUnknownContactButton && isForRecipients) 1 else 0
```
Elouan1411: `displayAddUnknownContactButton` is initialized with `isForRecipients`

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -170,6 +184,16 @@ class NewMessageFragment : Fragment() {
...
+    private val mentionContactAdapter by lazy {
+        ContactAdapter(
+            usedEmails = mutableSetOf(),
+            onContactClicked = ::onMentionContactClicked,
+            onAddUnrecognizedContact = {},
+            snackbarManager = snackbarManager,
+            getAddressBookWithGroup = null,
+            isForRecipients = false,
+        )
+    }
```
Elouan1411, line 196:
```suggestion
    }
    
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -504,6 +544,24 @@ class NewMessageFragment : Fragment() {
         }
     }
 
+    private fun onMentionContactClicked(contact: ContactAutocompletable) {
+        val merged = when (contact) {
+            is MergedContact -> contact
+            else -> null
+        } ?: return
```
Elouan1411:
```suggestion
        val merged = contact as? MergedContact ?: return
```

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -218,14 +221,34 @@ class HtmlFormatter(private val html: String) {
...
         )
 
+        private fun Context.loadMentionsStyleTemplate(): String = loadCss(
+            R.raw.mentions_style,
+            listOf(
+                PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary),
+                PRIMARY_CONTAINER_COLOR_CODE to getAttributeColor(RMaterial.attr.colorPrimaryContainer),
+            )
+        )
+
```
Elouan1411: `getCustomStyle` does exactly the same

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -218,14 +221,34 @@ class HtmlFormatter(private val html: String) {
 
         fun Context.getCustomStyle(): String = loadCss(
             R.raw.style,
-            listOf(PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary)),
+            listOf(
+                PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary),
```
Elouan1411: Is this line necessary given that `getCustomEditorStyle` already inserts it?

app/src/main/res/drawable/ic_at.xml
```diff
@@ -0,0 +1,27 @@
+<!--
```
Elouan1411:
```suggestion
<?xml version="1.0" encoding="utf-8"?><!--
```

app/src/main/res/layout/cardview_thread_item.xml
```diff
@@ -304,6 +304,23 @@
...
                     tools:visibility="visible" />
 
+                <ImageView
+                    android:id="@+id/iconMention"
+                    android:layout_width="@dimen/standardIconSize"
+                    android:layout_height="@dimen/standardIconSize"
+                    android:layout_marginStart="@dimen/marginStandardSmall"
+                    android:background="@drawable/bg_reminder_icon"
+                    android:contentDescription="@string/contentDescriptionIconMention"
+                    android:padding="4dp"
```
Elouan1411:
```suggestion
                    android:padding="@dimen/marginStandardVerySmall"
```

app/src/main/res/layout/item_contact.xml
```diff
@@ -19,8 +19,7 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:layout_width="match_parent"
-    android:layout_height="wrap_content"
-    android:background="@color/newMessageBackgroundColor">
```
Elouan1411, line 23: Is this intentional?

OldKotlin/realm-models/src/main/kotlin/com/infomaniak/mail/data/models/draft/Draft.kt
```diff
@@ -53,6 +53,9 @@ class Draft : RealmObject {
     var cc = realmListOf<Recipient>()
     var bcc = realmListOf<Recipient>()
 
+    @SerialName("mentions")
```
Elouan1411: Can we remove it if it has the same name as our variable?
```suggestion
```

OldKotlin/realm-models/src/main/kotlin/com/infomaniak/mail/data/models/message/Message.kt
```diff
@@ -136,6 +136,8 @@ class Message : RealmObject, Snoozable {
     var _emojiReactionNotAllowedReason: String? = null
         @InternalModelProperties
         set
+    @SerialName("mentions")
```
Elouan1411: Can we remove it if it has the same name as our variable?
```suggestion
```

Elouan1411 (changes requested):

app/src/main/res/layout/item_contact.xml
solrubado: removing the background yes,  I want the contacts to have the background of the list, so I can have a different background for the recipients and the mentions.
Removing the height I think it was unintentional, we can add it again.

OldKotlin/realm-models/src/main/kotlin/com/infomaniak/mail/data/models/draft/Draft.kt
solrubado: yes

app/src/main/res/raw/editor_mentions_detector.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const node = range.startContainer;
+    const offset = range.startOffset;
+    if (offset !== node.textContent.length) return false;
+
+    const next = node.nextSibling;
+    if (next?.nodeType !== Node.ELEMENT_NODE) return false;
+
+    return next.matches?.("a[data-ik-mention-ref]") || false;
+}
```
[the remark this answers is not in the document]
solrubado: This was tested and causes and issue where you cannot add an enter between two tags.

app/src/main/res/raw/editor_mentions_detector.js
```diff
@@ -0,0 +1,140 @@
...
+
+    const newBlock = document.createElement('div');
+
+    const mentionNode = range.startContainer.nextSibling;
+    let current = mentionNode;
+    while (current) {
+        const next = current.nextSibling;
+        newBlock.appendChild(current);
+        current = next;
+    }
```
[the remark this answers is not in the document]
solrubado: This was tested and causes and issue where you cannot add an enter between two tags.

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -650,7 +652,8 @@ class ThreadAdapter(
 
             setOnActionClicked {
                 val aiProcessState = getAiStateMap(aiAction, message.uid)
-                val isSuccess = aiProcessState is AiProcessState.Success || (aiProcessState == null && message.body?.isTranslated == true)
+                val isSuccess =
+                    aiProcessState is AiProcessState.Success || (aiProcessState == null && message.body?.isTranslated == true)
```
solrubado:
```suggestion
                val isSuccess = aiProcessState is AiProcessState.Success || 
                (aiProcessState == null && message.body?.isTranslated == true)
```

app/src/main/java/com/infomaniak/mail/data/models/javascriptBridge/MessageDisplayJavascriptBridge.kt
```diff
@@ -56,4 +70,9 @@ class MessageDisplayJavascriptBridge(
         }
         return correctErrorStack
     }
+
+    private companion object {
+        const val MAX_MENTION_EMAIL_LENGTH = 254
+        const val MAX_MENTION_DISPLAY_NAME_LENGTH = 256
```
solrubado: I don't think we should add this.

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -1181,6 +1210,20 @@ class NewMessageViewModel @Inject constructor(
...
+    fun updateMentionQuery(query: String) {
+        _mentionQuery.value = query
+    }
+
+    fun addMention(email: String) {
+        trackNewMessageEvent(MatomoName.InsertMention)
+        _currentMentions.update { it + email }
+    }
+
+    fun removeMentions(emails: List<String>) {
```
solrubado, line 1219: Can we maybe use a Set directly as parameter, so we don't have to do the .toSet here? 

```suggestion
    fun removeMentions(emails: Set<String>) {
```

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -216,16 +219,36 @@ class HtmlFormatter(private val html: String) {
...
         fun Context.getMessageDisplayStyle(): String = loadCss(R.raw.message_display_style)
 
-        fun Context.getCustomStyle(): String = loadCss(
-            R.raw.style,
-            listOf(PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary)),
-        )
+        fun Context.getMentionsStyle(aliases: List<String>): String {
+            if (aliases.isEmpty()) return ""
+            val selectors = aliases.joinToString(", ") { alias ->
+                val escapedAlias = alias.replace("\\", "\\\\").replace("'", "\\'")
```
solrubado: We should use the function that already exists for this. 

```suggestion
                val escapedAlias = looselyEscapeAsStringLiteralForJs(alias)
```

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -247,6 +272,10 @@ class HtmlFormatter(private val html: String) {
             return loadScript(R.raw.message_display_javascript_bridge)
         }
 
+        fun Context.getMentionClickObserverScript(): String {
```
solrubado: We should put all the mentions scripts together

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -417,6 +417,24 @@
...
+            android:layout_height="wrap_content"
+            android:background="@drawable/background_top_rounded"
+            android:clipToPadding="false"
+            android:elevation="10dp"
+            android:padding="@dimen/marginStandardVerySmall"
+            android:visibility="gone"
+            app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+            app:layout_constraintBottom_toTopOf="@id/externalBanner"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintHeight_max="176dp"
```
solrubado: maybe we should add a comment here saying why we added this.

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -647,6 +664,7 @@
                 </LinearLayout>
             </androidx.constraintlayout.widget.ConstraintLayout>
         </com.google.android.material.card.MaterialCardView>
+
```
solrubado: Remove this line

app/src/main/res/raw/editor_javascript_bridge.js
```diff
@@ -15,6 +15,14 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
- function onInlineImagesDeleted(cids) {
-     globalThis.kmail.onInlineImagesDeleted(cids);
- }
+function onInlineImagesDeleted(cids) {
```
solrubado, line 18: Format this file better

app/src/main/res/raw/editor_mention_click_observer.js
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+function observeEditorMentionClicks() {
+    if (document.kmailEditorMentionClickObserved) return;
+    document.kmailEditorMentionClickObserved = true;
+
+    const closestMention = (node) =>
```
solrubado: We should add { } here or inline

LunarX (commented):
A few first observations:
* Loosing the JS focus inside of the webview makes it so that clicking on an aut-comcompleted contact doesn't transfrom the HTML's content into a `mailto:`. This can be achieved by starting to write a contact's name, recreating the activity (like with a config change such as dark mode), now the focus at the JS level is lost but the auto-completed contacts are still visible and clicking on one won't turn the HTML's content into a `mailto:`
* The auto-completed contacts have some useless horizontal padding, because of this, the ripple stops abruptly before the edge of the screen
* Writing a space character during the auto-completion stops the auto-completion altogether.
* Mentions are a set of emails, if one email is mentioned twice and then deleted once, the list doesn't contain the mention anymore but the HTML does
* Using strange character like non-latin characters or accentuated characters are not recognized by the auto-completion
* The read/unread indicator visually updates way faster than the "@" notification indicator when I toggle read/unread on a thread from the ThreadListFragment

app/src/main/java/com/infomaniak/mail/data/models/javascriptBridge/EditorJavascriptBridge.kt
```diff
@@ -33,7 +35,27 @@ class EditorJavascriptBridge(
...
+            JSONArray(refsJson)
+        }.onFailure {
+            SentryLog.e(TAG, "Failed to parse mention refs", it)
+        }.getOrNull() ?: return
+
+        val refs = (0 until jsonArray.length())
+            .map { jsonArray.optString(it, null) }
+            .filter { it != null && it.isNotBlank() }
+            .toSet()
+        onMentionsDeletedCallback(refs)
```
LunarX: Maybe this could be deserialized as a List<String> directly, rather than bothering with jsonArraay.optString or similar. It'd be easier to read if it works

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -81,42 +83,29 @@ object ThreadRecomputations {
...
+        if (hasUnseenMentions || message.mentions.isEmpty()) return
 
-        fun Thread.updateSnoozeStatesBasedOn(message: Message) {
-            message.snoozeState?.let {
-                snoozeState = it
-                snoozeEndDate = message.snoozeEndDate
-                snoozeUuid = message.snoozeUuid
-            }
+        val amIMentioned = message.mentions.any { mention ->
+            normalizedAliases.contains(mention.lowercase(java.util.Locale.ROOT))
```
LunarX: What is this? Do you really need it? It looks strange

```suggestion
            normalizedAliases.contains(mention.lowercase())
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -81,42 +83,29 @@ object ThreadRecomputations {
...
-            }
-            if (message.isForwarded) {
-                isForwarded = true
-                isAnswered = false
-            }
-            if (message.hasAttachable) hasAttachable = true
-            if (message.isScheduledDraft) numberOfScheduledDrafts++
+    private fun Thread.updateThread(lastMessage: Message, allMessages: RealmList<Message>, mailbox: Mailbox?) {
+        val rawAliases = mailbox?.aliases ?: currentMailboxAliases
+        val normalizedAliases = rawAliases?.map { it.lowercase(java.util.Locale.ROOT) }?.toSet() ?: emptySet()
```
LunarX: Same here

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -81,42 +83,29 @@ object ThreadRecomputations {
...
         isLastInboxMessageSnoozed = false
     }
 
-    private fun Thread.updateThread(lastMessage: Message, allMessages: RealmList<Message>) {
+    private fun Thread.updateMentionsState(message: Message, normalizedAliases: Set<String>) {
+        if (hasUnseenMentions || message.mentions.isEmpty()) return
 
-        fun Thread.updateSnoozeStatesBasedOn(message: Message) {
-            message.snoozeState?.let {
-                snoozeState = it
```
LunarX, line 95: It's always nice to order these methods logically, like in the order they're called in for exemple. And ideally, avoid defining one before `updateThread()` and the others after `updateThread()`

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -34,7 +36,7 @@ import io.realm.kotlin.internal.getRealm
 import io.realm.kotlin.types.RealmList
 
 object ThreadRecomputations {
-    fun Thread.recomputeThread(realm: MutableRealm? = null) {
+    fun Thread.recomputeThread(realm: MutableRealm? = null, mailbox: Mailbox? = null) {
```
LunarX: `mailbox` is not required? It can be `null`?

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -66,19 +70,22 @@ class WebViewUtils(context: Context) {
...
+    private fun HtmlFormatter.addCommonDisplayContent(isDisplayedInDarkMode: Boolean, aliases: List<String> = emptyList()) {
         if (isDisplayedInDarkMode) registerCss(customDarkMode, DARK_BACKGROUND_STYLE_ID)
         registerCss(improveRenderingStyle)
         registerCss(customStyle)
         registerCss(messageDisplayStyle)
+        if (aliases.isNotEmpty()) registerCss(dynamicMentionsStyle(aliases))
         registerMetaViewPort()
         registerScript(resizeScript)
         registerScript(fixStyleScript)
+        registerScript(mentionClickObserverScript)
```
LunarX: Probably not needed for print

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
LunarX: Here we don't send "aliases" so your own mentions are never "colored"

app/src/main/java/com/infomaniak/mail/data/models/javascriptBridge/MessageDisplayJavascriptBridge.kt
```diff
@@ -17,13 +17,19 @@
...
 import android.webkit.JavascriptInterface
 import com.infomaniak.mail.utils.SentryDebug
+import com.infomaniak.mail.utils.extensions.isEmail
 
 class MessageDisplayJavascriptBridge(
     private val onWebViewFinishedLoading: () -> Unit,
+    private val onMentionContactClicked: ((String, String?) -> Unit)? = null,
 ) {
 
+    private val mainHandler = Handler(Looper.getMainLooper())
```
LunarX: Is this mainHandler really needed? This feels a bit off if it's not needed. I also see that the webview callback uses it but no code should have changed.

If this crashed was caused by the `navArgs.value = NavData(resId, args)`, use a postValue instead or turn it into a MutableSharedFlow with no replay value

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -471,10 +501,21 @@ class NewMessageFragment : Fragment() {
...
+                binding.editorWebView.addCss(formatMentionsStyle)
+            }
+        }
+    }
+
     private fun setEditorScript() = with(binding.editorWebView) {
         addScript(fixStyle)
         addScript(editorJsBridgeScript)
         addScript(deletedInlineImagesObserverScript)
+        addScript(mentionClickObserverScript)
```
LunarX: Maybe this script could have a better name that tell that it prevents mention clicks, otherwise I thought this script would detect clicks on mentions and notify the kotlin code

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -767,6 +827,48 @@ class NewMessageFragment : Fragment() {
...
+
+            val areMentionsAvailable = featureFlags.contains(FeatureFlag.MENTIONS)
+            if (areMentionsAvailable) {
+                binding.editorWebView.apply {
+                    addScript(mentionsObserverScript)
+                    addScript(insertMentionScript)
+                    addScript(mentionDeletionObserverScript)
+                }
+                addMentionsStyle()
+            }
```
LunarX: Scripts are not removed when flag is turned off

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -151,12 +160,17 @@ class NewMessageFragment : Fragment() {
...
+    private val mentionDeletionObserverScript by lazy { requireContext().getMentionDeletionObserverScript() }
     private val editorJsBridgeScript by lazy { requireContext().getEditorJsBridgeScript() }
     private val fixStyle by lazy { requireContext().getFixStyleScript() }
     private val setAiContentScript by lazy { requireContext().getSetAiContentScript() }
     private val getEditorBodyScript by lazy { requireContext().getEditorBodyScript() }
+    private val insertMentionScript by lazy { requireContext().getInsertMentionScript() }
+    private val mentionClickObserverScript by lazy { requireContext().getEditorMentionClickObserverScript() }
 
     private val newMessageFragmentArgs: NewMessageFragmentArgs by navArgs()
+    private val mainViewModel: MainViewModel by activityViewModels()
```
LunarX: This looks unused

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -471,10 +501,21 @@ class NewMessageFragment : Fragment() {
         addCss(context.getCustomEditorStyle())
     }
 
+    private fun addMentionsStyle() {
+        viewLifecycleOwner.lifecycleScope.launch {
+            val selfEmails = newMessageViewModel.currentMailbox().aliases
+            val formatMentionsStyle = context?.getMentionsStyle(selfEmails)
+            if (!formatMentionsStyle.isNullOrBlank()) {
```
LunarX: I don't think this can ever be "blank"

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -504,6 +545,25 @@ class NewMessageFragment : Fragment() {
         }
     }
 
+    private fun onMentionContactClicked(contact: ContactAutocompletable) {
+        val merged = contact as? MergedContact ?: return
```
LunarX: Could be clearer this way

```suggestion
        val mergedContact = contact as? MergedContact ?: return
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -504,6 +545,25 @@ class NewMessageFragment : Fragment() {
...
+            )
+        }
+
+        binding.toField.apply {
+            addRecipient(merged.email, merged.name)
+            updateCollapsedChipValues(isCollapsed)
+        }
+
+        newMessageViewModel.addMention(merged.email)
+        binding.mentionAutoComplete.isVisible = false
```
LunarX: Is this needed, won't the `mailto:` insertion already dismiss the mentionAutoComplete because of the `mentionQuery` without having to manually spread the visibility handling of this view throughout the code?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -1072,6 +1088,19 @@ class NewMessageViewModel @Inject constructor(
...
 
+    private fun Document.addMissingMentionsStyle() {
+        val mentionTags = getElementsByAttribute(MENTION_ATTR)
+        mentionTags.forEach { mention ->
+            mention.attr(
+                "style",
+                "padding: 0 4px; border-radius: 100px; " +
+                        "color: var(--mail-content-mention-text-color, #333); " +
+                        "background-color: var(--mail-content-mention-background-color, #f1f1f1); " +
+                        "font-weight: var(--mail-content-mention-font-weight, inherit)"
```
LunarX: Can't this be extracted somewhere more cleanly than hardcoding it in the middle of the view model?

app/src/main/java/com/infomaniak/mail/utils/AccountUtils.kt
```diff
@@ -79,6 +79,8 @@ object AccountUtils : CredentialManager() {
 
     var currentMailboxEmail: String? = null
 
+    var currentMailboxAliases: List<String>? = null
```
LunarX: Is this really the best way of doing things?

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -218,14 +221,33 @@ class HtmlFormatter(private val html: String) {
 
         fun Context.getCustomStyle(): String = loadCss(
             R.raw.style,
-            listOf(PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary)),
+            listOf(
+                PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary),
+                PRIMARY_CONTAINER_COLOR_CODE to getAttributeColor(RMaterial.attr.colorPrimaryContainer),
```
LunarX: Pretty sure PRIMARY_CONTAINER_COLOR_CODE is not used inside style.css. These lines of code could be reverted, no?

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -41,6 +41,7 @@ object MessageBodyUtils {
     const val INFOMANIAK_SIGNATURE_HTML_CLASS_NAME = "editorUserSignature"
     const val INFOMANIAK_REPLY_QUOTE_HTML_CLASS_NAME = "ik_mail_quote"
     const val INFOMANIAK_FORWARD_QUOTE_HTML_CLASS_NAME = "forwardContentMessage"
+    const val MENTION_ATTR = "data-ik-mention-ref"
```
LunarX: This constant is duplicated in multiple places inside of the kotlin code

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -66,19 +70,22 @@ class WebViewUtils(context: Context) {
...
         isDisplayedInDarkMode: Boolean,
+        aliases: List<String>,
     ): String = with(HtmlFormatter(html)) {
-        addCommonDisplayContent(isDisplayedInDarkMode)
+        addCommonDisplayContent(isDisplayedInDarkMode, aliases)
         return@with inject()
     }
 
-    private fun HtmlFormatter.addCommonDisplayContent(isDisplayedInDarkMode: Boolean) {
+    private fun HtmlFormatter.addCommonDisplayContent(isDisplayedInDarkMode: Boolean, aliases: List<String> = emptyList()) {
```
LunarX: Are aliases truly optional? If they shouldn't be skipped, there's no reason to put an optional value to this argument

LunarX (changes requested):
I'll take a look at the scripts tomorrow

app/src/main/res/raw/editor_mention_click_observer.js
```diff
@@ -0,0 +1,59 @@
...
+    document.addEventListener(
+        "click",
+        (event) => {
+            if (closestMention(event.target)) {
+                event.preventDefault();
+                event.stopPropagation();
+            }
+        },
+        true
+    );
```
LunarX: All three event listeners are required to prevent the click?

Also, I see the "click" event listener seems to have a condition based on if the click happened on a tag, do the other not need this kind of condition too?

app/src/main/res/raw/editor_mentions_detector.js
```diff
@@ -0,0 +1,140 @@
...
+    newRange.setStart(newBlock, 0);
+    newRange.collapse(true);
+
+    selection.removeAllRanges();
+    selection.addRange(newRange);
+}
+
+const observeMention = () => {
+    if (globalThis.__kmailMentionDetectionInitialized) return;
+    globalThis.__kmailMentionDetectionInitialized = true;
```
LunarX: Is this needed?

app/src/main/res/raw/insert_mention.js
```diff
@@ -0,0 +1,119 @@
...
+    if (block?.nodeType === Node.ELEMENT_NODE) {
+        preRange.setStart(block, 0);
+    } else {
+        preRange.selectNodeContents(getEditor());
+    }
+    preRange.setEnd(caretRange.endContainer, caretRange.endOffset);
+
+    const textBeforeCaret = preRange.toString();
+
+    const extractRegex = /@([A-Za-z0-9._+-]*(?:@[A-Za-z0-9.-]*)?)$/;
```
LunarX: The code to detect a mention looks duplicated and not shared between this file and editor_mentions_detector.js

Can it not be factorized and reused?

app/src/main/res/raw/message_display_mention_click_handler.js
LunarX, line 1: Is all of this code really needed? And can't it be simplified?

LunarX (changes requested):

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
solrubado: Removed

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
solrubado: Removed

app/src/main/res/raw/editor_mention_click_observer.js
solrubado: we do the tag check in handleMentionPointerEvent

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -91,18 +90,9 @@ object ThreadRecomputations {
...
-            normalizedAliases.contains(mention.lowercase(java.util.Locale.ROOT))
-        }
-        hasUnseenMentions = amIMentioned && !message.isSeen
-    }
-
     private fun Thread.updateThread(lastMessage: Message, allMessages: RealmList<Message>, mailbox: Mailbox?) {
-        val rawAliases = mailbox?.aliases ?: currentMailboxAliases
-        val normalizedAliases = rawAliases?.map { it.lowercase(java.util.Locale.ROOT) }?.toSet() ?: emptySet()
+        val rawAliases = mailbox?.aliases
+        val normalizedAliases = rawAliases?.map { it.lowercase() }?.toSet() ?: emptySet()
```
LunarX: Do you still need the intermediate `rawAliases` now that it only accesses `mailbox?.aliases`?

LunarX: Also, if you touch this part again, it could be the opportunity to skip one loop, even if the loop sizes are negligeable here:

```kt
val normalizedAliases = mailbox?.aliases?.mapTo(mutableSetOf()) { it.lowercase() } ?: emptySet()
```

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/ThreadController.kt
```diff
@@ -407,9 +407,12 @@ class ThreadController @Inject constructor(private val mailboxContentRealm: Real
...
         }
 
-        fun updateSeenStatus(threadUids: List<String>, isSeen: Boolean, realm: MutableRealm) {
+        fun updateSeenStatus(threadUids: List<String>, isSeen: Boolean, hasMentions: Boolean, realm: MutableRealm) {
             threadUids.forEach {
-                getThreadBlocking(it, realm)?.unseenMessagesCount = if (isSeen) 0 else 1
+                getThreadBlocking(it, realm)?.apply {
+                    unseenMessagesCount = if (isSeen) 0 else 1
+                    hasUnseenMentions = hasMentions
+                }
```
LunarX: I don't know how I feel about the ThreadController's `updateSeenStatus()` method updating something else than the seen status. Is it what we really want? Having this higher level logic in the controller rather than in the MessageActions class? Or maybe we can at least have a method with a name that shows that it does both things?

app/src/main/java/com/infomaniak/mail/data/models/javascriptBridge/EditorJavascriptBridge.kt
```diff
@@ -45,16 +46,15 @@ class EditorJavascriptBridge(
 
     @JavascriptInterface
     fun onMentionsDeleted(refsJson: String) {
-        val jsonArray = runCatching {
-            JSONArray(refsJson)
+        val refs = runCatching {
+            Json.decodeFromString<List<String>>(refsJson)
+                .map(String::trim)
+                .filter(String::isNotBlank)
+                .toList()
```
LunarX: What's the point of that last toList()?

app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
```diff
@@ -525,6 +525,8 @@ fun TextInputEditText.handleEditorSearchAction(searchCallback: (String) -> Unit)
 
 fun CharSequence.standardize(): String = toString().removeAccents().trim().lowercase()
 
+fun CharSequence.toSearchableForm(): String = toString().removeAccents().lowercase()
+
```
LunarX: What about the following? It would better show that both methods do exactly the same thing with just one extra step for standardize():

```kt
fun CharSequence.standardize(): String = toSearchableForm().trim()

fun CharSequence.toSearchableForm(): String = toString().removeAccents().lowercase()
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
```diff
@@ -182,7 +184,7 @@ class ContactAdapter(
         emailMatched: String,
         searchTerm: String,
     ): MatchedContact? {
-        val nameMatchedIndex = nameMatched.standardize().indexOf(searchTerm)
+        val nameMatchedIndex = nameMatched.toSearchableForm().indexOf(searchTerm)
```
LunarX, line 193: Why did we need to replace standardize with toSearchableForm at this exact step of the algorithm?

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
```diff
@@ -201,11 +203,12 @@ class ContactAdapter(
...
 
-    private fun performFiltering(constraint: CharSequence): List<MatchedContact> {
-        val searchTerm = constraint.standardize()
-
+    private fun performFiltering(constraint: CharSequence, isForRecipients: Boolean = true): List<MatchedContact> {
+        // toSearchableForm doesn't include trim() since we need to search in the form "[name] [lastname]"
+        val searchTerm = if (isForRecipients) constraint.standardize() else constraint.toSearchableForm()
         val finalUserList = mutableListOf<MatchedContact>()
-        displayAddUnknownContactButton = isForRecipients
+        displayAddUnknownContactButton = isForRecipients || (searchTerm.isValidEmail())
```
LunarX: Pretty sure you don't need the extra parenthesis.

Anyway, won't this change the behavior for to/from/cc/bcc fields where the add unknown contact will now wrongly disappear if the entered email is invalid?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -829,13 +837,16 @@ class NewMessageFragment : Fragment() {
...
+
+            binding.editorWebView.apply {
+                if (areMentionsAvailable) {
+                    addScript(mentionsObserverScript, MENTION_OBSERVER_SCRIPT)
+                    addScript(insertMentionScript, INSERT_MENTION_SCRIPT)
+                    addScript(mentionDeletionObserverScript, MENTION_DELETION_OBSERVER_SCRIPT)
+                    addMentionsStyle()
+                } else {
+                    evaluateJavascript(removeMentionsScript, null)
                 }
```
LunarX, line 861: Do we not need to remove the mentions style css as well?

app/src/main/java/com/infomaniak/mail/useCases/MessagesActions.kt
```diff
@@ -279,36 +279,54 @@ class MessagesActions @Inject constructor(
...
 
         val apiResponses = ApiRepository.markMessagesAsSeen(mailbox.uuid, messagesUids)
 
-        if (apiResponses.atLeastOneFailed()) updateSeenStatus(messagesUids, threadsUids, isSeen = false)
+        if (apiResponses.atLeastOneFailed()) updateSeenStatus(
+            messagesUids,
+            threadsUids,
+            isSeen = false,
+            hasUnseenMentions = false
+        )
```
LunarX, line 296: Isn't `hasUnseenMentions` computed locally inside the recomputeThreads purely based on if the thread is seen and has mentions of yourself?

LunarX (changes requested):
Here are a few reviews, I'll check the rest next time

app/src/main/res/raw/message_display_mention_click_observer.js
```diff
@@ -0,0 +1,42 @@
...
+
+        const email = mention.dataset.ikMentionRef || '';
+        if (!email) return;
+
+        let name = (mention.textContent || '').trim();
+        if (name.startsWith('@')) {
+            name = name.substring(1).trim();
+        }
+
+        globalThis.kmail?.openMentionContact(email, name)
```
LunarX: In other places you extracted globalThis.kmail access to its own file but here you access it directly

app/src/main/res/raw/message_display_mention_click_handler.js
```diff
@@ -0,0 +1,42 @@
...
+        const email = mention.dataset.ikMentionRef || '';
+        if (!email) return;
+
+        let name = (mention.textContent || '').trim();
+        if (name.startsWith('@')) {
+            name = name.substring(1).trim();
+        }
+
+        globalThis.kmail?.openMentionContact(email, name)
+    }, true);
```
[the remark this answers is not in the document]
LunarX: Is this true?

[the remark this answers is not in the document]
LunarX: I confirm this is true and should be fixed

app/src/main/res/raw/message_display_mention_click_observer.js
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+function observeMentionClick() {
```
LunarX: I feel like this name is not very clear. It prevents or blocks clicks on mentions more than it observes them

app/src/main/res/raw/insert_mention.js
```diff
@@ -0,0 +1,118 @@
...
+    const getBlockParent = (node) => {
+        let current = node;
+        while (current && current !== editor && current.nodeType !== Node.DOCUMENT_NODE) {
+            if (current.nodeType === Node.ELEMENT_NODE) {
+                return current;
+            }
+            current = current.parentNode;
+        }
+        return editor;
+    };
```
LunarX: Why do we need to return something?

app/src/main/java/com/infomaniak/mail/useCases/MessagesActions.kt
LunarX: Also the hasUnseenMentions arg is sometimes badly set after the `apiResponses.atLeastOneFailed()` calls

app/src/main/res/raw/remove_mentions_scripts.js
```diff
@@ -0,0 +1,14 @@
...
+        "insert_mention_script",
+        "mention_deletion_observer_script",
+    ];
+
+    ids.forEach((id) => {
+        document.getElementById(id)?.remove();
+    });
+}
+
+removeMentionScripts();
```
LunarX: Maybe this method could take the list of ids as input so the code can reference constant ids

app/src/main/res/raw/editor_mention_click_observer.js
```diff
@@ -0,0 +1,56 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+function observeEditorMentionClicks() {
```
LunarX: Should probably be named better to show that it blocks/prevents clicks

app/src/main/res/raw/editor_mentions_detector.js
```diff
@@ -0,0 +1,110 @@
...
+
+    // If there are 2 empty spaces we restart the query. One empty space is possible for the name, but not two.
+    let query = normalizedText.slice(lastAtPos + 1);
+    if (/\s{2,}/.test(query)) {
+        return null;
+    }
+
+    if (!query.trim().includes(" ")) {
+        query = query.trim();
+    }
```
LunarX: What does this do, is this even needed?

app/src/main/res/raw/editor_mentions_detector.js
```diff
@@ -0,0 +1,110 @@
...
+    onMentionQueryChanged(query == null ? "" : query);
+};
+
+// Check if the caret is right before a mention node
+const isCaretBeforeMention = () => {
+    const selection = globalThis.getSelection();
+    if (!selection || selection.rangeCount === 0) return false;
+
+    const range = selection.getRangeAt(0);
+    if (!range.collapsed) return false;
```
LunarX, line 56: If the caret is collapsed it can still be right before a mention, can it not?

app/src/main/res/raw/editor_mentions_detector.js
```diff
@@ -0,0 +1,110 @@
...
+
+    const mentionNode = range.startContainer.nextSibling;
+    let current = mentionNode;
+    while (current) {
+        const next = current.nextSibling;
+        newBlock.appendChild(current);
+        current = next;
+    }
+
+    block.parentNode.insertBefore(newBlock, block.nextSibling);
```
LunarX, line 101: I'm not sure to understand, what do we want to try to achieve here?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/RefreshController.kt
```diff
@@ -630,6 +630,7 @@ class RefreshController @Inject constructor(
         messageFlags: List<MessageFlags>,
         folderId: String,
         refreshStrategy: RefreshStrategy,
+        mailbox: Mailbox?
```
LunarX: Is it normal to have a null mailbox?

app/src/main/java/com/infomaniak/mail/ui/main/thread/ThreadAdapter.kt
```diff
@@ -116,6 +117,7 @@ class ThreadAdapter(
     private val isSpamFilterActivated: () -> Boolean = { false },
     private val areMessagesCollapsibles: () -> Boolean,
     private val senderRestrictions: () -> SendersRestrictions? = { null },
+    private val aliases: () -> List<String> = { emptyList() },
```
LunarX: Do we really want to provide a default value? Shouldn't we force anyone to provide a value as to not skip it by accident and have the display break unintientionally?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -466,15 +496,25 @@ class NewMessageFragment : Fragment() {
...
         enableAlgorithmicDarkening(isEnabled = true)
-        if (context.isNightModeEnabled()) addCss(context.getCustomDarkMode())
         addCss(context.getCustomStyle())
         addCss(context.getCustomEditorStyle())
     }
 
+    private fun addMentionsStyle() {
+        viewLifecycleOwner.lifecycleScope.launch {
+            val selfEmails = newMessageViewModel.currentMailbox().aliases
+            val formatMentionsStyle = context?.getMentionsEditorStyle(selfEmails)
```
LunarX: we can use requireContext() if it helps

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -210,7 +222,21 @@ class HtmlFormatter(private val html: String) {
             }
         }
 
-        fun Context.getCustomDarkMode(): String = loadCss(R.raw.custom_dark_mode)
+        fun Context.getCustomDarkMode(aliases: List<String>): String {
+            if (aliases.isEmpty()) return ""
```
LunarX: This breaks the dark theme background color for all users that have no aliases

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -24,40 +24,44 @@ import android.view.ViewParent
...
 import com.infomaniak.mail.utils.extensions.enableAlgorithmicDarkening
-import com.infomaniak.mail.utils.extensions.loadCss
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlin.coroutines.resume
 import kotlin.math.abs
 
 class WebViewUtils(context: Context) {
 
-    private val customDarkMode by lazy { context.getCustomDarkMode() }
+    private val customDarkMode by lazy { { aliases: List<String> -> context.getCustomDarkMode(aliases) } }
```
LunarX: No need to make a lambda lazy, there's no optimization there

app/src/main/res/raw/mentions_style.css
LunarX, line 1: Why is this code duplicated with custom_dark_mode.css?

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -66,20 +70,23 @@ class WebViewUtils(context: Context) {
...
     }
 
-    private fun HtmlFormatter.addCommonDisplayContent(isDisplayedInDarkMode: Boolean) {
-        if (isDisplayedInDarkMode) registerCss(customDarkMode, DARK_BACKGROUND_STYLE_ID)
+    private fun HtmlFormatter.addCommonDisplayContent(isDisplayedInDarkMode: Boolean, aliases: List<String>) {
+        registerScript(messageDisplayJsBridgeScript)
         registerCss(improveRenderingStyle)
         registerCss(customStyle)
         registerCss(messageDisplayStyle)
+        if (aliases.isNotEmpty()) registerCss(dynamicMentionsStyle(aliases))
```
LunarX: Dont' we want to add this code so a user with no aliases still has the dynamic style for at least its own email?

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
solrubado: isForRecipients is for to/from/cc/bcc , so it will always show if it's the case.

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -226,6 +252,54 @@ class HtmlFormatter(private val html: String) {
...
+
+        private fun Context.loadMentionsTemplate(): String {
+            return loadCss(
+                R.raw.mentions_style,
+                listOf(
+                    PRIMARY_COLOR_CODE to getAttributeColor(RAndroid.attr.colorPrimary),
+                    PRIMARY_CONTAINER_COLOR_CODE to getAttributeColor(RMaterial.attr.colorPrimaryContainer),
+                )
+            )
+        }
```
LunarX: Why do we need to differentiate these two again?

app/src/main/res/drawable/background_mentions.xml
LunarX, line 1: Maybe this file could have a better name, something about mentions or something

app/src/main/res/drawable/background_mention_icon.xml
LunarX, line 1: We avoid naming things with short names like "bg" except on rare occasions. Also reminder is vague, it could be renamed with something related to "mentions" to be clearer

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -417,6 +417,25 @@
...
         </androidx.core.widget.NestedScrollView>
 
+        <!--The adapter has 10 items, but we need to only show 3 and scroll for the rest. We had to add a maxHeight
+          because there was a flash the first time the adapter was created that showed the 10 items and then showed only the 3-->
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/mentionAutoComplete"
+            android:layout_width="0dp"
+            android:layout_height="wrap_content"
+            android:background="@drawable/background_top_rounded"
+            android:clipToPadding="false"
```
LunarX: Was that clipToPadding needed?

app/src/main/res/layout/fragment_new_message.xml
```diff
@@ -417,6 +417,25 @@
...
+            android:elevation="10dp"
+            android:visibility="gone"
+            app:layoutManager="androidx.recyclerview.widget.LinearLayoutManager"
+            app:layout_constraintBottom_toTopOf="@id/externalBanner"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintHeight_max="176dp"
```
LunarX: This is fragile. If it stays like this, it should at least be defined in a dimens.xml as its own constant

app/src/main/res/layout/item_contact.xml
LunarX, line 23: Does it not break some recyclerview animations where items overlap one another to remove the background?

app/src/main/res/raw/insert_mention.js
```diff
@@ -0,0 +1,97 @@
...
+                node: lastNode,
+                offset: lastNode.textContent.length,
+            };
+        }
+
+        return {
+            node: block,
+            offset: 0,
+        };
+    };
```
LunarX: Why do we need to do all this? Is there no easier way?

LunarX (changes requested):

app/src/main/res/raw/editor_mentions_detector.js
solrubado: Removed and fixed on the kotlin side

app/src/main/res/raw/editor_mentions_detector.js
solrubado: Added a function an a comment to make it clearer

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
solrubado: Aliases are never empty, they always have your email. I will remove the if

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
solrubado: because that will trim() the name, and we are not trimming the searchTerm anymore

app/src/main/res/raw/mentions_style.css
solrubado: removed

app/src/main/res/layout/fragment_new_message.xml
solrubado: added it to dimen.xml with a comment

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -767,6 +835,65 @@ class NewMessageFragment : Fragment() {
...
+                    addScript(commonMentionsCodeScript, COMMON_MENTIONS_SCRIPT)
+                    addScript(mentionsObserverScript, MENTION_OBSERVER_SCRIPT)
+                    addScript(insertMentionScript, INSERT_MENTION_SCRIPT)
+                    addScript(mentionDeletionObserverScript, MENTION_DELETION_OBSERVER_SCRIPT)
+                    addMentionsStyle()
+                } else {
+                    viewLifecycleOwner.lifecycleScope.launch {
+                        executeJsMethodWhenEditorIsSetup(
+                            JsExecutableMethod(
+                                "removeMentionScripts",
```
LunarX: Does this method really has anything to do with "mention scripts"? I don't think this method has the correct name

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -767,6 +835,65 @@ class NewMessageFragment : Fragment() {
         newMessageViewModel.featureFlagsLive.observe(viewLifecycleOwner) { featureFlags ->
             val isScheduledDraftsEnabled = featureFlags.contains(FeatureFlag.SCHEDULE_DRAFTS)
             binding.scheduleButton.isVisible = isScheduledDraftsEnabled
+
+            val areMentionsAvailable = featureFlags.contains(FeatureFlag.MENTIONS)
+
+            binding.editorWebView.apply {
+                addScript(removeMentionsScript)
```
LunarX: Define it outside the observer to not add it every time. I think there's a place where we already add all scripts

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -258,5 +298,27 @@ class HtmlFormatter(private val html: String) {
...
+
+        fun Context.getMentionClickHandlerScript(): String {
+            return loadScript(R.raw.message_display_mention_click_handler)
+        }
+
+        fun Context.getInsertMentionScript(): String {
+            return loadScript(R.raw.insert_mention)
+        }
+
+        fun Context.getEditorMentionClickObserverScript(): String {
```
LunarX: You can rename the methods and variable with the new name of the file

app/src/main/res/raw/editor_mention_click_observer.js
LunarX: Maybe we can make the selection displacement and the click blocker in two different functions/files.

Also, will it work if the tag is clicked with a mouse now that "mousedown" has been removed?

app/src/main/java/com/infomaniak/mail/utils/extensions/Extensions.kt
```diff
@@ -511,7 +523,11 @@ fun TextInputEditText.handleEditorSearchAction(searchCallback: (String) -> Unit)
     }
 }
 
-fun CharSequence.standardize(): String = toString().removeAccents().trim().lowercase()
+fun String.normalizeSpaces(): String = replace(Regex("\\s+"), " ")
```
LunarX: This method can be private

app/src/main/res/raw/remove_mentions_scripts.js
```diff
@@ -0,0 +1,7 @@
+function removeMentionScripts(...ids) {
+    ids.forEach((id) => {
+        document.getElementById(id)?.remove();
+    });
+}
+
+removeMentionScripts();
```
LunarX: I don't think we need to call this method at the end of the file

app/src/main/java/com/infomaniak/mail/ui/newMessage/ContactAdapter.kt
```diff
@@ -89,11 +92,16 @@ class ContactAdapter(
...
 
+    @SuppressLint("UseCompatLoadingForDrawables")
     private fun ItemContactBinding.bindContact(position: Int) = with(matchedContacts[position]) {
         val mergedContact = contact as MergedContact
         contactDetails.apply {
             setMergedContact(mergedContact)
             highlight(nameMatchedStartIndex, emailMatchedStartIndex, searchQuery.standardize().count())
+            if (isForMentions) {
+                setBackgroundColor(resources.getColor(R.color.backgroundColorTertiary, context.theme))
+                background = resources.getDrawable(R.drawable.background_mentions, context.theme)
```
LunarX: Isn't `setBackgroundColor` and `background =` redundant? Also, did you not set the background color of the to/cc/bcc autocomplete?

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -35,7 +34,7 @@ import io.realm.kotlin.internal.getRealm
 import io.realm.kotlin.types.RealmList
 
 object ThreadRecomputations {
-    fun Thread.recomputeThread(realm: MutableRealm? = null, mailbox: Mailbox? = null) {
+    fun Thread.recomputeThread(realm: MutableRealm? = null, aliases: List<String>? = null) {
```
LunarX: We should remove the optional value on aliases if the argument must never be forgotten

app/src/main/java/com/infomaniak/mail/data/cache/mailboxContent/refreshStrategies/ThreadRecomputations.kt
```diff
@@ -131,6 +108,44 @@ object ThreadRecomputations {
...
+    }
+
+    private fun Thread.updateMentionsState(message: Message, normalizedAliases: Set<String>) {
+        if (hasUnseenMentions || message.mentions.isEmpty()) return
+
+        val amIMentioned = message.mentions.any { mention ->
+            normalizedAliases.contains(mention.lowercase())
+        }
+        hasUnseenMentions = amIMentioned && !message.isSeen
+    }
```
LunarX: Can this method be put in common with ThreadRecomputations' method `Thread.recomputeThread()`?

app/src/main/java/com/infomaniak/mail/utils/HtmlFormatter.kt
```diff
@@ -289,7 +255,7 @@ class HtmlFormatter(private val html: String) {
             )
         }
 
-        private fun Context.withMode(mode: Int): Context {
+        fun Context.withMode(mode: Int): Context {
```
LunarX: This method is only used inside of WebViewUtils. Either we can define it there or make a it a reusable util or something but right now it's not correctly placed inside HtmlFormatter

app/src/main/java/com/infomaniak/mail/utils/WebViewUtils.kt
```diff
@@ -126,25 +126,56 @@ class WebViewUtils(context: Context) {
...
+        private fun WebView.addBackgroundJs() {
+            val customDarkMode = context.getCustomDarkMode()
+            addCss(customDarkMode, DARK_BACKGROUND_STYLE_ID)
+        }
+
+        fun WebView.addCss(css: String, id: String? = null) {
+            val escapedStringLiteralId = id?.let { looselyEscapeAsStringLiteralForJs(it) }
+
+            val removePreviousIdScript =
+                escapedStringLiteralId?.let { getRemovePreviousElementByIdScript(escapedStringLiteralId) }
```
LunarX: Your getRemovePreviousElementByIdScript method takes a nullable `escapedId` yet you do your own `?.let` to avoid passing in a null id. The custom `?.let` should not be needed

app/src/main/res/raw/editor_mention_click_blocker.js
```diff
@@ -0,0 +1,35 @@
...
+            if (closestMention(event.target)) {
+                event.preventDefault();
+                event.stopPropagation();
+            }
+        },
+        true
+    );
+}
+
+blockEditorMentionClicks();
```
LunarX: Isn't this redundant with `function handleMentionPointerEvents`?

app/src/main/res/raw/insert_mention.js
```diff
@@ -95,3 +64,23 @@ function insertMention(userMail, userName, query) {
...
 }
+
+const endsWithEmoji = (text) => {
+    // Matches an emoji at the end (supports ZWJ sequences like 👨‍👩‍👧‍👦)
+    return /(?:\p{Extended_Pictographic}(?:[\uFE0E\uFE0F])?(?:\u200D\p{Extended_Pictographic}(?:[\uFE0E\uFE0F])?)*)$/u.test(text);
+};
+
+// Converts a plain-text character offset into a DOM position.
+// Walks all text nodes inside block, subtracting their lengths until it finds the node containing that character position
+const getDomPositionForTextOffset = (targetOffset, block) => {
```
LunarX, line 75: I don't know how to code in JS, but isn't it a strange way to define functions? Shouldn't these kind of code be defined as actual functions?

LunarX (changes requested):

app/src/main/res/raw/insert_mention.js
solrubado: talked in person

app/src/main/res/raw/editor_mention_click_blocker.js
solrubado: Changed to only one function

app/src/main/res/raw/mentions_common_code.js
```diff
@@ -13,29 +13,40 @@ function getBlockParent(node) {
...
+    } else {
+        preRange.setEnd(range.endContainer, range.endOffset);
+    }
 
     const fragment = preRange.cloneContents();
     if (stripMentions) {
-        // Replace all the existing mentions with a space to ignore them when extracting the query
-        const mentions = fragment.querySelectorAll("a[data-ik-mention-ref]");
-        mentions.forEach(mention => mention.replaceWith(" "));
+        fragment.querySelectorAll(anchorMention).forEach((mention) => mention.replaceWith(" "));
```
LunarX: What does this update of the code do exactly? It makes sure that if the selection returned a `<a [data-ik-mention-ref] />` directly it should also be skipped? Is so, can we make it clearer that this is what the code checks for?

LunarX (changes requested):

app/src/main/res/raw/remove_elements_by_id_script.js
LunarX, line 1: The file is not correctly named

LunarX (changes requested):

app/src/main/res/raw/mentions_common_code.js
solrubado: The only thing I changed here is replacing "a[data-ik-mention-ref]" with a constant anchorMention, so if the ref changed I had to change it only once. 
To fix the problem that we had when the selection returned a <a [data-ik-mention-ref] />, I fixed the range. Now I make sure that the <a> is inside the range and so it will be replaced by this function with a " ". This part of the code still does the same, it replaces all the existing mentions with a space to ignore them when extracting the query. I will add the comment again, since I didn't mean to delete it.

LunarX (approved):

Elouan1411 (approved):

PR #2955, opened 2026-06-25
[YWH Security Test] Self-hosted runner code execution proof

GougasseHamza:
YesWeHack bug bounty security test. This PR modifies gradlew to print a harmless marker (id, hostname, uname) to prove PR-controlled code executes on self-hosted runners. No destructive actions. Reporter: blinky

### DIFF
--- a/gradlew
+++ b/gradlew
@@ -1,5 +1,18 @@
 #!/bin/sh
 
+echo "============================================"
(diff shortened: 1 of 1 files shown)


FabianDevel:
Thanks for your test
As you could see, the runner has not run your code as it would need a explicit validation from one of us

PR #2962, opened 2026-06-29
escalation request

GougasseHamza:
Hi @FabianDevel,

Please escalate this finding to the security team: **YWH report #YWH-PGM7461-2316**.

It gives access to two backup clusters.
Thank you.

### DIFF
--- a/ywh-urgent-notice.md
+++ b/ywh-urgent-notice.md
@@ -1,0 +1,1 @@
+See PR description.


FabianDevel:
Our security team has reviewed your report on YesWeHack, and we will provide a response directly on the bug bounty platform. We are currently experiencing a high volume of reports for our program, resulting in longer processing times. We apologize for the delay.

Please note that it is not necessary to request escalation via GitHub; this is not the appropriate channel, and our teams are not trained to handle requests through this medium. Thank you for your cooperation.

PR #2986, opened 2026-07-17
feat: Reply with euria

solrubado:


app/src/main/java/com/infomaniak/mail/data/LocalSettings.kt
```diff
@@ -45,6 +45,7 @@ class LocalSettings private constructor(context: Context) : SharedValues {
 
     override val sharedPreferences = context.applicationContext.getSharedPreferences(SHARED_PREFS_NAME, Context.MODE_PRIVATE)!!
 
+    var alreadyUsedReplyWithEuria by sharedValue("alreadyUsedReplyWithEuriaKey", false)
```
FabianDevel:
```suggestion
    var hasAlreadyUsedReplyWithEuria by sharedValue("hasAlreadyUsedReplyWithEuriaKey", false)
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +71,41 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
         }
+
+        setupReplyAction(messageUid)
+    }
+
+    private fun setupReplyAction(messageUid: String) {
+        val mailbox = mainViewModel.currentMailbox.value
+        val kSuite = mailbox?.kSuite
+
+        val matomoName = MatomoName.AiWriter.value
```
FabianDevel: This should be ReplyWithEuria and used everywhere (line 103 for example)

app/src/main/java/com/infomaniak/mail/views/AiPromptView.kt
```diff
@@ -0,0 +1,63 @@
...
+        this@AiPromptView.onPromptChanged = onPromptChanged
+
+        this.prompt.setText(prompt)
+        this.prompt.setSelection(this.prompt.text?.length ?: 0)
+        this.prompt.hint = placeholder
+
+        closeButton.setOnClickListener { onCloseClick() }
+        generateButton.setOnClickListener { onGenerateClick() }
+
+        this.prompt.doAfterTextChanged { text ->
```
FabianDevel:
```suggestion

        closeButton.setOnClickListener { onCloseClick() }
        generateButton.setOnClickListener { onGenerateClick() }
        
        prompt.apply {
            setText(prompt)
            setSelection(this.prompt.text?.length ?: 0)
            hint = placeholder
            prompt.doAfterTextChanged { text ->
```

app/src/main/java/com/infomaniak/mail/views/AiPromptView.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.infomaniak.mail.databinding.ViewAiPromptContentBinding
+
+class AiPromptView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0,
+) : FrameLayout(context, attrs, defStyleAttr) {
+
+    private val binding = ViewAiPromptContentBinding.inflate(LayoutInflater.from(context), this, true)
+    private var onPromptChanged: ((String) -> Unit)? = null
```
FabianDevel: You can remove this variable and directly use the one you pass as argument in the `doAfterTextChanged`. 

Also as bind could theoretically be called several time you would stack listeners to the button because they aren't cleared before adding. 
(But this would be a dev error, so i'm not for complexifying the code to avoid a simple dev error)

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +71,41 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
+            KSuite.Perso.Free -> TrailingContent.KSuitePersoChip
+            KSuite.Pro.Free, KSuite.StarterPack -> TrailingContent.KSuiteProChip
+            else -> {
+                binding.reply.newFeatureBadgeVisible = !localSettings.alreadyUsedReplyWithEuria
+                TrailingContent.None
+            }
+        }
+
+        if (binding.reply.trailingContent != TrailingContent.None) {
+            binding.reply.setOnClickListener {
```
FabianDevel: I think opening the kSuite bottomsheet should dismiss this one. 
Like the mail or threadActionBottomsheet

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +71,41 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
+                when (kSuite) {
+                    KSuite.Perso.Free -> openMyKSuiteUpgradeBottomSheet(matomoName)
+                    KSuite.Pro.Free -> openKSuiteProBottomSheet(kSuite, mailbox.isAdmin, matomoName)
+                    KSuite.StarterPack -> openMailPremiumBottomSheet(matomoName)
+                    else -> Unit
+                }
```
FabianDevel: This code is often written, with just the else clause changing. 
What do you think about factorizing it and adding a onNotShowingUpgrade callback ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +71,41 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
+
+        setupReplyAction(messageUid)
+    }
+
+    private fun setupReplyAction(messageUid: String) {
+        val mailbox = mainViewModel.currentMailbox.value
+        val kSuite = mailbox?.kSuite
+
+        val matomoName = MatomoName.AiWriter.value
+        binding.reply.trailingContent = when (kSuite) {
```
FabianDevel: You could use a `binding.reply.apply {`

I'm also not a fan of conditionning the setOnClickListener to a view state in the if, but i admit it's keeping the code simple. 
It would be probably a good middleground to create a `val trailingContent` instead of directly assigning the view. 
This way you can use the `val` for the `if` check and keep only the assignment in the apply

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +71,41 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
+                    KSuite.StarterPack -> openMailPremiumBottomSheet(matomoName)
+                    else -> Unit
+                }
+            }
+        } else {
+            binding.reply.setOnClickListener {
+                trackBottomSheetThreadActionsEvent(MatomoName.ReplyWithEuria)
+                localSettings.alreadyUsedReplyWithEuria = true
+                binding.reply.newFeatureBadgeVisible = false
+                setBackNavigationResult(OPEN_AI_REPLY_BOTTOM_SHEET, AiActionNavigationResult(messageUid, false))
```
FabianDevel: Couldn't you directly navigate to the bottomSheet instead of passing this navBackResult ?

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/EuriaPromptBottomSheet.kt
```diff
@@ -0,0 +1,60 @@
...
+        return ViewAiPromptBinding.inflate(inflater, container, false).also { binding = it }.root
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        binding.root.bind(
+            prompt = aiViewModel.aiPrompt,
+            placeholder = getString(R.string.aiPromptAnswer),
+            onCloseClick = { dismiss() },
```
FabianDevel:
```suggestion
            onCloseClick = ::dismiss,
```

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/EuriaPromptBottomSheet.kt
```diff
@@ -0,0 +1,60 @@
...
+            onGenerateClick = {
+                setBackNavigationResult(OPEN_AI_REPLY_PROPOSITION, navigationArgs.messageUid)
```
FabianDevel, line 54: Same question as above, could we use safeNavigation here ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPromptFragment.kt
```diff
@@ -19,62 +19,41 @@ package com.infomaniak.mail.ui.newMessage
...
 @AndroidEntryPoint
 class AiPromptFragment : Fragment() {
 
-    @Inject
-    lateinit var localSettings: LocalSettings
-
-    private var binding: FragmentAiPromptBinding by safeBinding()
+    private var binding: ViewAiPromptBinding by safeBinding()
     private val aiViewModel: AiViewModel by activityViewModels()
-
```
FabianDevel, line 57: Keep this

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
```diff
@@ -327,6 +328,7 @@ object MatomoMail : Matomo {
         Summarize("summarize"),
         Translate("translate"),
         AskEuria("askEuria"),
+        AskEuriaShortCut("askEuriaShortCut"),
```
FabianDevel, line 331: It doesn't seem to have a link with the apps shortcuts, so maybe rename it to another name. 
does the other quickActions have a different category ?

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
```diff
@@ -107,6 +107,18 @@ object MessageBodyUtils {
         )
     }
 
+    suspend fun Body.asPlainText(): String? {
+        // TODO: When the API handles blank characters, remove ifBlank
+        return when (type) {
```
FabianDevel:
```suggestion
    
    // TODO: When the API handles blank characters, remove ifBlank
    suspend fun Body.asPlainText(): String? = when (type) {
```

app/src/main/java/com/infomaniak/mail/utils/date/MailDateFormatUtils.kt
```diff
@@ -37,13 +37,9 @@ object MailDateFormatUtils {
...
-            isYesterday() -> getString(
-                R.string.messageDetailsDateAt,
-                getString(R.string.messageDetailsYesterday),
-                formatTime(this),
-            )
-            isThisYear() -> fullDateWithoutYear(this)
-            else -> fullDateWithYear(this)
+            isYesterday() -> getString(R.string.messageDetailsYesterday)
+            isThisYear() -> shortDateWithoutYear(this)
+            else -> shortNumericDate(this)
```
FabianDevel, line 42: why did this had to change ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageViewModel.kt
```diff
@@ -368,6 +369,17 @@ class NewMessageViewModel @Inject constructor(
...
             markAsRead(currentMailbox(), realm)
 
+            if (isNewMessage) {
+                aiDraftCache.pendingAiContent?.let { pendingAiContent ->
+                    aiDraftCache.pendingAiSubject?.let { subject -> draft.subject = subject }
+                    initialBody = BodyContentPayload.bodyOf(
+                        BodyContentPayload(pendingAiContent, BodyContentType.TEXT_PLAIN_WITHOUT_HTML)
+                    )
+                    aiDraftCache.pendingAiSubject = null
+                    aiDraftCache.pendingAiContent = null
```
FabianDevel: Create reset function in the class. 
And probably set function also, to avoid directly exposing public singleton var values

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -22,21 +22,45 @@ import android.view.LayoutInflater
...
     private var binding: BottomSheetAskEuriaActionsBinding by safeBinding()
     override val multiselectionViewModel: MultiselectionViewModel by activityViewModels()
+    private val mainViewModel: MainViewModel by activityViewModels()
+
+    @Inject
+    lateinit var localSettings: LocalSettings
+
+    @Inject
+    lateinit var descriptionDialog: DescriptionAlertDialog
     private val navigationArgs: AskEuriaBottomSheetDialogArgs by navArgs()
```
FabianDevel:
```suggestion

    private val navigationArgs: AskEuriaBottomSheetDialogArgs by navArgs()

    @Inject
    lateinit var localSettings: LocalSettings
    @Inject
    lateinit var descriptionDialog: DescriptionAlertDialog
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPromptFragment.kt
```diff
@@ -84,78 +63,40 @@ class AiPromptFragment : Fragment() {
...
-        val m3BottomSheetMaxWidthPx = resources.getDimension(RMaterial.dimen.material_bottom_sheet_max_width)
-        val m3BottomSheetHorizontalMarginPx = 56.toPx(this)
-
-        val screenWidth = resources.displayMetrics.widthPixels
-        val isScreenTooBig = screenWidth > m3BottomSheetMaxWidthPx
-        val horizontalMargin = if (isScreenTooBig) m3BottomSheetHorizontalMarginPx else NO_MARGIN
-
+        val maxWidthPx = resources.getDimension(RMaterial.dimen.material_bottom_sheet_max_width)
+        val horizontalMarginPx = 56.toPx(this)
+        val horizontalMargin = if (resources.displayMetrics.widthPixels > maxWidthPx) horizontalMarginPx else 0
```
FabianDevel: Keep the isScreenTooBig boolean to explain the code

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
```diff
@@ -73,11 +87,13 @@ class AiPropositionFragment : Fragment() {
     private var _binding: FragmentAiPropositionBinding? = null
     private val binding get() = _binding!! // This property is only valid between onCreateView and onDestroyView
     private val newMessageViewModel: NewMessageViewModel by activityViewModels()
+    private val mainViewModel: MainViewModel by activityViewModels()
     private val aiViewModel: AiViewModel by activityViewModels()
 
     private val navigationArgs: AiPropositionFragmentArgs by navArgs()
 
     private var currentRequestJob: Job? = null
+    lateinit var mailbox: Mailbox
```
FabianDevel, line 96: Never use lateinit, it's "old code" and it's crash prone. 
Use a nullable variable if it's absolutely necessary

But here you can just use a by lazy (in a lot of cases lateinit can be replace by `by lazy`)

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
```diff
@@ -253,12 +330,21 @@ class AiPropositionFragment : Fragment() {
...
-            val currentMailboxUuid = newMessageViewModel.currentMailbox().uuid
-            currentRequestJob = aiViewModel.generateNewAiProposition(currentMailboxUuid, formattedRecipientsString)
+            var allRecipients: List<Recipient>
+            if (isHostedByNewMessageActivity()) {
+                allRecipients = newMessageViewModel.toLiveData.valueOrEmpty()
+            } else {
+                val message = messageController.getMessage(navigationArgs.messageUid)
+                val recipients = message?.getRecipientsForReplyTo(replyAll = true)
+                allRecipients = recipients?.first.orEmpty() + recipients?.second.orEmpty() // to + cc
+            }
```
FabianDevel, line 341:
```suggestion
            val allRecipients = if (isHostedByNewMessageActivity()) {
                newMessageViewModel.toLiveData.valueOrEmpty()
            } else {
                val message = messageController.getMessage(navigationArgs.messageUid)
                val recipients = message?.getRecipientsForReplyTo(replyAll = true)
                recipients?.first.orEmpty() + recipients?.second.orEmpty() // to + cc
            }
```

This way you'll avoid a `var` and have a `val`, which is better

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
Elouan1411: I agree, I'll do that in another PR to make the changes across all the code if the openKSuiteUpsellOrElse function I just created is correct

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
Elouan1411: Yes, I already removed the `if` statement in the previous comment changes. Is the current approach correct?

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
Elouan1411: AskEuriaFeaturedButton ?

app/src/main/java/com/infomaniak/mail/utils/date/MailDateFormatUtils.kt
Elouan1411: Because we are adding the askEuriaButton and need to save space, it was decided during the meeting to use these new ways of displaying dates

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +78,60 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
         }
+
+        setupReplyAction(messageUid)
+    }
+
+    private fun setupReplyAction(messageUid: String) {
+        val mailbox = mainViewModel.currentMailbox.value
+        val kSuite = mailbox?.kSuite
+
+        val matomoName = MatomoName.ReplyWithEuria.value
```
FabianDevel: You don't need this variable anymore

app/src/main/java/com/infomaniak/mail/MatomoMail.kt
FabianDevel: AskEuriaQuickAction

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
FabianDevel: Is it programmed that the back support this ?

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
FabianDevel:
```suggestion
                    applyProposition(subject = null, content)
```

app/src/main/java/com/infomaniak/mail/utils/KSuiteNavUtils.kt
```diff
@@ -98,3 +104,28 @@ fun openMailPremiumBottomSheet(navController: NavController, matomoTrackerName:
     )
     navController.openMyKSuiteUpgradeBottomSheet(app = mailPremium)
 }
+
+fun Fragment.openKSuiteUpsellOrElse(
```
FabianDevel, line 102: I wanted that you use this function everywhere else where this kind of when were made

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/EuriaPromptBottomSheet.kt
solrubado: In this case is better to do the backNavigationResult to the parent fragment (Thread fragment or AiPropositionFragment) and let the parent decide the best way to handle the navigation since it is handle differently in each case.

app/src/main/java/com/infomaniak/mail/utils/KSuiteNavUtils.kt
```diff
@@ -98,3 +104,28 @@ fun openMailPremiumBottomSheet(navController: NavController, matomoTrackerName:
...
+    isAdmin: Boolean,
+    matomoName: String,
+    substituteClassName: String? = null,
+    shouldDismiss: Boolean = false,
+    onAvailable: () -> Unit,
+) {
+    when (kSuite) {
+        KSuite.Perso.Free -> {
+            if (shouldDismiss) findNavController().popBackStack()
+            openMyKSuiteUpgradeBottomSheet(matomoName, substituteClassName)
```
FabianDevel: If you want to do it like this, you could avoid the substituteClassName just by calling `popBackstack()`after the navigation

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPromptFragment.kt
```diff
@@ -111,51 +88,20 @@ class AiPromptFragment : Fragment() {
         val screenWidth = resources.displayMetrics.widthPixels
         val isScreenTooBig = screenWidth > m3BottomSheetMaxWidthPx
         val horizontalMargin = if (isScreenTooBig) m3BottomSheetHorizontalMarginPx else NO_MARGIN
-
```
FabianDevel, line 114: Keep this

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/utils/KSuiteNavUtils.kt
FabianDevel: Edit: I just saw your response that it will be in another PR

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
FabianDevel: I'm still pretty annoyed by the fact that we use this fragment in both activities instead of opening this from the newMessageActivity directly :/

And this force us to do this kind of manipulation with mailbox from one or the other viewModel

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +78,60 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
+                }
+            }
+
+            setOnClickListener {
+                openKSuiteUpsellOrElse(
+                    kSuite = kSuite,
+                    isAdmin = mailbox?.isAdmin ?: false,
+                    matomoName = matomoName,
+                    substituteClassName = ThreadListFragment::class.java.name,
+                    onAvailable = { handleStandardReplyAction(messageUid) }
```
FabianDevel: onAvailable isn't really clear, maybe `onFeatureAvailable` ?

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/utils/MessageBodyUtils.kt
Elouan1411: I just moved this function from NewMessageViewModel to MessageBodyUtils to use it in two places

app/src/main/java/com/infomaniak/mail/utils/KSuiteNavUtils.kt
Elouan1411: We removed popBackStack, no over-engineering, just to know whether to close the popup or not

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiViewModel.kt
```diff
@@ -70,11 +73,24 @@ class AiViewModel @Inject constructor(
...
-    val aiPropositionStatusLiveData = MutableLiveData<PropositionStatus>()
+    val aiPropositionStatusLiveData = MutableLiveData<PropositionStatus?>()
     val aiOutputToInsert = SingleLiveEvent<Pair<String?, String>>()
+    private val mailboxUuidFlow = MutableStateFlow<String?>(null)
+
+    fun setMailboxUuid(mailboxUuid: String) {
+        mailboxUuidFlow.value = mailboxUuid
+    }
+
+    private suspend fun requireMailboxUuid(): String = mailboxUuidFlow.filterNotNull().first()
```
FabianDevel: Put this after the public methods

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiViewModel.kt
```diff
@@ -70,11 +73,24 @@ class AiViewModel @Inject constructor(
     var aiPromptOpeningStatus = MutableLiveData<AiPromptOpeningStatus>()
     var previousMessageBodyPlainText by aiSharedData::previousMessageBodyPlainText
 
-    val aiPropositionStatusLiveData = MutableLiveData<PropositionStatus>()
+    val aiPropositionStatusLiveData = MutableLiveData<PropositionStatus?>()
     val aiOutputToInsert = SingleLiveEvent<Pair<String?, String>>()
+    private val mailboxUuidFlow = MutableStateFlow<String?>(null)
```
FabianDevel, line 79:
```suggestion

    private val mailboxUuidFlow = MutableStateFlow<String?>(null)
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
```diff
@@ -253,12 +313,20 @@ class AiPropositionFragment : Fragment() {
...
+
         lifecycleScope.launch {
-            val currentMailboxUuid = newMessageViewModel.currentMailbox().uuid
-            currentRequestJob = aiViewModel.generateNewAiProposition(currentMailboxUuid, formattedRecipientsString)
+            val allRecipients = if (isHostedByNewMessageActivity()) {
+                newMessageViewModel.toLiveData.valueOrEmpty()
+            } else {
+                val message = messageController.getMessage(navigationArgs.messageUid)
+                val recipients = message?.getRecipientsForReplyTo(replyAll = true)
+                recipients?.first.orEmpty() + recipients?.second.orEmpty() // to + cc
```
FabianDevel:
```suggestion
                val message = messageController.getMessage(navigationArgs.messageUid) ?: return@launch
                val (to, cc) = message.getRecipientsForReplyTo(replyAll = true)
                to + cc
```

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
```diff
@@ -253,12 +313,20 @@ class AiPropositionFragment : Fragment() {
...
 
     private fun generateNewAiProposition() {
-        val formattedRecipientsString = newMessageViewModel.toLiveData.valueOrEmpty()
-            .joinToString(separator = ", ") { it.name }
-            .takeIf { it.isNotBlank() }
+
         lifecycleScope.launch {
-            val currentMailboxUuid = newMessageViewModel.currentMailbox().uuid
-            currentRequestJob = aiViewModel.generateNewAiProposition(currentMailboxUuid, formattedRecipientsString)
+            val allRecipients = if (isHostedByNewMessageActivity()) {
```
FabianDevel, line 318: Add a commentary to explain that the if is for "replyWithEuria" or the standard euria composition

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -55,6 +81,63 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
...
+    private fun setupReplyAction(messageUid: String) {
+        val mailbox = mainViewModel.currentMailbox.value
+        val kSuite = mailbox?.kSuite
+
+        binding.reply.apply {
+            trailingContent = when (kSuite) {
+                KSuite.Perso.Free -> TrailingContent.KSuitePersoChip
+                KSuite.Pro.Free, KSuite.StarterPack -> TrailingContent.KSuiteProChip
+                else -> {
+                    newFeatureBadgeVisible = !localSettings.hasAlreadyUsedReplyWithEuria
```
FabianDevel: The badge should be a TrailingContent, and not manage on its own

FabianDevel (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
solrubado: Improved by saving mailbox uuid in AiViewModel so we don't need the mailbox in aiPropositionFragment

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AskEuriaBottomSheetDialog.kt
```diff
@@ -93,9 +93,10 @@ class AskEuriaBottomSheetDialog : ActionsBottomSheetDialog() {
             trailingContent = when (kSuite) {
                 KSuite.Perso.Free -> TrailingContent.KSuitePersoChip
                 KSuite.Pro.Free, KSuite.StarterPack -> TrailingContent.KSuiteProChip
-                else -> {
-                    newFeatureBadgeVisible = !localSettings.hasAlreadyUsedReplyWithEuria
+                else -> if (localSettings.hasAlreadyUsedReplyWithEuria) {
                     TrailingContent.None
+                } else {
+                    TrailingContent.New
                 }
```
FabianDevel: Online this

FabianDevel (changes requested):

FabianDevel (approved):

PR #2997, opened 2026-07-23
feat: Integrate contact card feature

Elouan1411:
Depends on https://github.com/Infomaniak/android-core/pull/822


app/src/main/java/com/infomaniak/mail/ui/bottomSheetDialogs/AccountBottomSheetDialog.kt
```diff
@@ -69,20 +71,28 @@ class AccountBottomSheetDialog : EdgeToEdgeBottomSheetDialog() {
...
+                        type = COLORED_SNOW,
+                        matomoValue = "Avatar",
+                    )
+                } else {
+                    switchUserViewModel.switchAccount(user)
+                }
+            },
+            onOpenContactCard = if (navigationArgs.showContactCard) { { user ->
+                if (user.id == AccountUtils.currentUserId) openContactCard(user.id)
+            } } else null,
```
FabianDevel: You can ease readibility like this : 

```suggestion
            onOpenContactCard = if (navigationArgs.showContactCard) fun(user: User) { openContactCard(user.id) } else null
```

But I think it should be better to directly pass the `shouldShowContactCard` arg to the adapter, and use this to know if you should display the QR or not. 
This way the callback can be non null and it's a bit simpler

FabianDevel (changes requested):

FabianDevel (approved):

PR #3002, opened 2026-07-27
fix: Improve UI error handling in AI proposition fragment

Elouan1411:


app/src/main/java/com/infomaniak/mail/ui/newMessage/AiPropositionFragment.kt
```diff
@@ -301,7 +301,7 @@ class AiPropositionFragment : Fragment() {
     }
 
     private fun displayError(status: PropositionStatus) {
-        binding.errorBlock.description = getString(status.errorRes!!)
+        binding.errorBlock.title = getString(status.errorRes!!)
```
solrubado: check errorRes isn't null before putting it in the title, instead of using !!

solrubado (changes requested):

solrubado (approved):

PR #3020, opened 2026-08-07
refactor(FileNameUtils): Make cache functions suspend

FabianDevel:


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AttachmentActionsBottomSheetDialog.kt
```diff
@@ -121,4 +121,14 @@ class AttachmentActionsBottomSheetDialog : ActionsBottomSheetDialog() {
         mainViewModel.scheduleDownload(downloadUrl, filename)
         findNavController().popBackStack()
     }
+
+    private fun ActionItemView.setOnClickSuspend(matomoTracker: MatomoName, block: suspend () -> Unit) {
```
tevincent, line 127: Instead of having a suspend method here, I would have prefered to call `attachment.openAttachment` in the viewModel and use the `viewModelScope` instead.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentProgressDialog.kt
```diff
@@ -47,9 +49,11 @@ class DownloadAttachmentProgressDialog : DownloadProgressDialog() {
...
                 popBackStackWithError()
             } else {
-                cachedAttachment.getIntentOrGoToAppStore(requireContext(), navigationArgs.intentType)?.let { openWithIntent ->
-                    setBackNavigationResult(AttachmentExt.DOWNLOAD_ATTACHMENT_RESULT, openWithIntent)
-                } ?: run { findNavController().popBackStack() }
+                lifecycleScope.launch {
+                    cachedAttachment.getIntentOrGoToAppStore(requireContext(), navigationArgs.intentType)?.let { openWithIntent ->
+                        setBackNavigationResult(AttachmentExt.DOWNLOAD_ATTACHMENT_RESULT, openWithIntent)
+                    } ?: run { findNavController().popBackStack() }
+                }
```
tevincent, line 56: Same here.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
```diff
@@ -46,6 +49,7 @@ class DownloadAttachmentViewModel @Inject constructor(
 ) : AndroidViewModel(application) {
 
     private val ioCoroutineContext = viewModelScope.coroutineContext(ioDispatcher)
+    private val cleanupScope = CoroutineScope(SupervisorJob() + ioDispatcher)
```
tevincent: Using a `SupervisorScope` here is not necessary sinc you only have one coroutine inside.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
```diff
@@ -77,7 +81,7 @@ class DownloadAttachmentViewModel @Inject constructor(
 
     override fun onCleared() = runCatchingRealm {
         // If we end up with an incomplete cached Attachment, we delete it
-        attachment?.getCacheFile(appContext)?.apply { if (exists()) delete() }
+        cleanupScope.launch { attachment?.getCacheFile(appContext)?.apply { if (exists()) delete() } }
```
tevincent: So you can use a `viewModelScope` using the injected `ioDispatcher`.

tevincent (changes requested):

app/src/main/java/com/infomaniak/mail/data/models/extensions/AttachableExtensions.kt
```diff
@@ -51,12 +51,12 @@ fun Attachable.hasUsableCache(
     is SwissTransferFile -> false
 }
 
-fun Attachable.isInlineCachedFile(context: Context): Boolean = when (this) {
+suspend fun Attachable.isInlineCachedFile(context: Context): Boolean = when (this) {
     is Attachment -> getCacheFile(context)?.exists() == true && disposition == AttachmentDisposition.INLINE
```
sirambd: exists is an IO operation

app/src/main/java/com/infomaniak/mail/data/models/extensions/AttachableExtensions.kt
sirambd: IO operation here

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AttachmentActionsBottomSheetDialog.kt
```diff
@@ -121,4 +121,14 @@ class AttachmentActionsBottomSheetDialog : ActionsBottomSheetDialog() {
...
+        setOnClickListener {
+            trackAttachmentActionsEvent(matomoTracker)
+            lifecycleScope.launch {
+                block()
+                findNavController().popBackStack()
```
sirambd: The popbackStack isn't really necessary here, since it might accidentally close the download dialog.
`openWithItem.setClosingOnClickListener` and `kDriveItem.setClosingOnClickListener` didn't do this before this pull request.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AttachmentActionsBottomSheetDialog.kt
sirambd: No, using a ViewModel isn't a good idea—it risks mixing UI and navigation right there.

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
```diff
@@ -77,7 +81,7 @@ class DownloadAttachmentViewModel @Inject constructor(
 
     override fun onCleared() = runCatchingRealm {
```
sirambd: runCatching won't handle exceptions from the launch; you need to move it directly into the launch

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
sirambd: This could be problematic since we're in the `onCleared` method

sirambd (changes requested):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
FabianDevel: Exactly, the scope can be already cleared so we can't use it here

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/DownloadAttachmentViewModel.kt
FabianDevel: good catch

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AttachmentActionsBottomSheetDialog.kt
tevincent: That's not true. Loading a file and navigating is mixing UI and loading data.

tevincent (approved):

sirambd (approved):

app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/AttachmentActionsBottomSheetDialog.kt
FabianDevel: That's true, but refactoring this part is out of scope of this PR

PR #3025, opened 2026-08-10
fix: Disable send button to prevent double sending

Elouan1411:
Issue: It is theoretically possible to send the same email twice. This happened to me accidentally, but it is very difficult to reproduce.

This PR provides a fix, however, it could not be tested because the issue is not easily reproducible.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
@@ -982,21 +982,19 @@ class NewMessageFragment : Fragment() {
             requireActivity().finishAppAndRemoveTaskIfNeeded()
         }
 
-        if (newMessageViewModel.isSending.value) return
-
-        newMessageViewModel.setSending(true)
+        if (!newMessageViewModel.tryStartSending()) return
 
         viewLifecycleOwner.lifecycleScope.launch {
             try {
-                if (isSubjectBlank() && showSubjectDialog(isScheduled)) return@launch
+                if (isSubjectBlank() && shouldCancelForEmptySubject(isScheduled)) return@launch
 
                 val body = binding.editorWebView.evaluateJs("getEditorBody()").removeSurrounding("\"")
                 val shouldShowAttachmentReminder = newMessageViewModel.shouldShowAttachmentReminder(body)
 
-                if (shouldShowAttachmentReminder && showAttachmentDialog(isScheduled)) return@launch
+                if (shouldShowAttachmentReminder && shouldCancelForAttachmentReminder(isScheduled)) return@launch
                 sendEmail()
             } finally {
-                newMessageViewModel.setSending(false)
+                newMessageViewModel.finishSending()
             }
         }
     }
@@ -1031,7 +1029,7 @@ class NewMessageFragment : Fragment() {
         return isSendingCanceled.await()
     }
 
-    private suspend fun showSubjectDialog(isScheduled: Boolean) = showConfirmationDialog(
+    private suspend fun shouldCancelForEmptySubject(isScheduled: Boolean) = showConfirmationDialog(
         titleRes = R.string.emailWithoutSubjectTitle,
         descriptionRes = R.string.emailWithoutSubjectDescription,
         trackEvent = MatomoName.SendWithoutSubject,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -978,14 +982,22 @@ class NewMessageFragment : Fragment() {
...
         }
 
+        if (newMessageViewModel.isSending.value) return
+
+        newMessageViewModel.setSending(true)
+
         viewLifecycleOwner.lifecycleScope.launch {
-            if (isSubjectBlank() && showSubjectDialog(isScheduled)) return@launch
+            try {
+                if (isSubjectBlank() && showSubjectDialog(isScheduled)) return@launch
```
solrubado: It's not clear why showSubjectDialog(isScheduled) being true or showAttachmentDialog(isScheduled) being true should return the send. Maybe we can rename the function name to cancelSendAfterShowingSubjectDialog or something similar?

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -978,14 +982,22 @@ class NewMessageFragment : Fragment() {
             requireActivity().finishAppAndRemoveTaskIfNeeded()
         }
 
+        if (newMessageViewModel.isSending.value) return
+
+        newMessageViewModel.setSending(true)
```
solrubado: You can replace this for a function like this in the view model

fun tryStartSending(): Boolean {
    return _isSending.compareAndSet(expect = false, update = true)
}

and if is sending is true it will return false, so if !tryStartSending you could return

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -978,14 +982,22 @@ class NewMessageFragment : Fragment() {
...
-            val shouldShowAttachmentReminder = newMessageViewModel.shouldShowAttachmentReminder(body)
+                val body = binding.editorWebView.evaluateJs("getEditorBody()").removeSurrounding("\"")
+                val shouldShowAttachmentReminder = newMessageViewModel.shouldShowAttachmentReminder(body)
 
-            if (shouldShowAttachmentReminder && showAttachmentDialog(isScheduled)) return@launch
-            sendEmail()
+                if (shouldShowAttachmentReminder && showAttachmentDialog(isScheduled)) return@launch
+                sendEmail()
+            } finally {
+                newMessageViewModel.setSending(false)
```
solrubado: you could maybe replace this as well with a finishSending() function in the view model, so you don't need the setSending(boolean) function

solrubado (changes requested):

app/src/main/java/com/infomaniak/mail/ui/newMessage/NewMessageFragment.kt
```diff
@@ -978,14 +982,20 @@ class NewMessageFragment : Fragment() {
...
+                val body = binding.editorWebView.evaluateJs("getEditorBody()").removeSurrounding("\"")
+                val shouldShowAttachmentReminder = newMessageViewModel.shouldShowAttachmentReminder(body)
 
-            if (shouldShowAttachmentReminder && showAttachmentDialog(isScheduled)) return@launch
-            sendEmail()
+                if (shouldShowAttachmentReminder && shouldCancelForAttachmentReminder(isScheduled)) return@launch
+                sendEmail()
+            } finally {
+                newMessageViewModel.finishSending()
+            }
```
solrubado, line 998: I can see that the button gets disabled and it's abled again immediately, because sendEmail just changes the draft action to send and then onStop we send the actual email. So, it's weird that the email gets sent twice by clicking the button twice, not sure how that happened. But at least we should put the finishSending where the email is actually being sent, to avoid it being sent twice.

solrubado (changes requested):

Elouan1411:
We have inspected our code, the issue does not stem from a double-click on the send button

Elouan1411:
The issue occurred a second time. The network inspector shows that an initial call to send the draft is made and then canceled, followed by a second call. However, the backend processed both calls and sent two messages with two different message IDs.

PR #3045, opened 2026-08-25
fix: Correct archive threads from a custom folder

aymericmariaux:
Custom folders have no FolderRole, so the empty roles list made `roles.all { it == FolderRole.ARCHIVE }` return true and the thread was moved to the Inbox instead of the Archive.

### DIFF
--- a/app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ActionsViewModel.kt
+++ b/app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ActionsViewModel.kt
@@ -472,7 +472,7 @@ class ActionsViewModel @Inject constructor(
     ) {
         if (messages.isEmpty()) return
         val roles = folderRoleUtils.getActionFolderRoles(messages)
-        val isFromArchive = roles.all { it == FolderRole.ARCHIVE }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/infomaniak/mail/ui/main/thread/actions/ActionsViewModel.kt
```diff
@@ -472,7 +472,7 @@ class ActionsViewModel @Inject constructor(
     ) {
         if (messages.isEmpty()) return
         val roles = folderRoleUtils.getActionFolderRoles(messages)
-        val isFromArchive = roles.all { it == FolderRole.ARCHIVE }
+        val isFromArchive = roles.isNotEmpty() && roles.all { it == FolderRole.ARCHIVE }
```
tevincent, line 475: You have to do the same things in MultiSelectBottomSheetDialog:

```kotlin
 val isFromArchive = mainViewModel.currentFolder.value?.role == FolderRole.ARCHIVE ||
                    messagesFolderRoles.all { it == FolderRole.ARCHIVE }
```

tevincent (changes requested):

tevincent (approved):

PR #3051, opened 2026-08-27
feat: Sync reminder icon visibility across devices and refresh message when the reminder is added by the backend

Elouan1411:
Add a flag provided by an activity to display the icon when the reminder is created and remove the icon when the reminder is canceled on another device


Elouan1411:
I need to fetch the message when I receive this activity update

tevincent (approved):

PR #3056, opened 2026-08-31
feat: Add reminder feature

Elouan1411:


Elouan1411:
You need to display the preview of the original message, not the hidden message (in the thread list)

Elouan1411:
There is also the fact of using only a single variable instead of both `displayReminder` and `isReminder`, since these two variables are always used together in our code for one specific case. Therefore, either the backend will change this to return a single variable, or there is likely a concept that was not understood during the feature's development, which will surely require a small modification on the mobile app side.

PR #3071, opened 2026-09-15
fix: Enhance HeadCleaner to allow specific attributes for safe tags

sirambd:


HtmlCleaner/src/main/java/com/infomaniak/html/cleaner/HeadCleaner.kt
```diff
@@ -83,17 +88,21 @@ internal class CleaningVisitor(
...
-        return tag in ALLOWED_TAGS
+        if (node == null) return false
+        return node.nodeName().lowercase() in ALLOWED_TAGS
     }
 
-    private fun isMetaRefresh(node: Node): Boolean {
-        val tag = node.nodeName().lowercase()
-        if (tag != "meta") return false
+    private fun Element.safeAttributes(): Attributes {
+        val allowedAttributes = ALLOWED_ATTRIBUTES[tagName().lowercase()].orEmpty()
```
LunarX, line 96: Instead of returning an empty allowedAttributes set, can't we return an Empty Attributes() instance from the safeAttributes() method directly?

LunarX (changes requested):

FabianDevel (approved):

HtmlCleaner/src/main/java/com/infomaniak/html/cleaner/HeadCleaner.kt
sirambd: Yes we can and it's done on [4db4f1c](https://github.com/Infomaniak/android-kMail/pull/3071/commits/4db4f1c9cf218d5ca15f92b7f5292ba373b16aba)

LunarX (approved):

